PR #13, opened 2019-08-28
added solution to task 1 "ArrayList"

SkaterWeather:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int init_capacity = 10;
```
Sofasmile:
```suggestion
    private static final int INIT_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
...
+    private Object[] data;
+    private int size;
+
+    public <T> ArrayList() {
+        data = new Object[init_capacity];
+        size = 0;
```
Sofasmile: it is redundant, by default size = 0

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
...
+    }
+
+    private void resizeIfFull() {
+        if (size == data.length) {
+            Object[] newData = new Object[size + 5];
```
Sofasmile: how many times we need to increase the size of the ArrayList?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
...
     public void add(T value) {
-
+        resizeIfFull();
+        data[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        resizeIfFull();
```
Sofasmile: move calling this method after `if` statement

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
...
+        }
+        System.arraycopy(data, index, data, index + 1, size - index);
+        data[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
Sofasmile: optional:
see how this method is implemented in the source code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
...
     @Override
     public void set(T value, int index) {
-
+        if (index >= size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        data[index] = value;
     }
 
+    //TODO HERE
```
Sofasmile: do we need todo here?

src/main/java/core/basesyntax/ArrayList.java
SkaterWeather: yes, at first i also wanted to name constants like that, but StyleChecker didn't allow me this

src/main/java/core/basesyntax/ArrayList.java
SkaterWeather: oops, forgot to delete

src/main/java/core/basesyntax/ArrayList.java
SkaterWeather: Yes, that's not optimal realisation, but if we want to implement it like in source code, we need also to implement method toArray() only for one call, because native method System.arrayscopy() recives only arrays type.

src/main/java/core/basesyntax/ArrayList.java
boroda4436: @SkaterWeather I have cloned your solution, renamed to INIT_CAPACITY: run the mvn package and everything works well. So comment from @Sofasmile is suitable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,49 +4,93 @@
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int INIT_CAPACITY = 8;
```
Sofasmile: why 8? by default  - 10

PR #14, opened 2019-08-28
added hw-7-1

kenu21:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List</p>
  */
+
+import java.util.Arrays;
+
 public class ArrayList<T> implements List<T> {
+    private Object[] array = new Object[16];
```
Sofasmile: why 16?

better rename to `elements` or `dataElements`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
...
     public void add(T value, int index) {
-
+        if (value == null) {
+            return;
+        }
+        if (index < 0 || index > indexTopArray) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        if (indexTopArray >= array.length) {
+            array = resizeAndCopyArray(array);
```
Sofasmile: can we make method resizeAndCopyArray() void?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
...
+        }
+        if (index == indexTopArray) {
```
Sofasmile: do we need this checking?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
...
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null) {
+            return;
+        }
+        for (int i = 0; i < list.size(); i++) {
```
Sofasmile: optional:
see how this method is implemented in the source code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= indexTopArray) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        T t = (T) array[index];
+        for (int i = index + 1; i < indexTopArray; i++) {
+            array[i - 1] = array[i];
```
Sofasmile: we can use method System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -3,50 +3,112 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return indexTopArray < 1;
+    }
+
+    private Object[] resizeAndCopyArray(Object[] input) {
+        return Arrays.copyOf(input, input.length * 2);
```
Sofasmile: why we increment size in twice?

kenu21: We increment size in 1.5, for example, to don't waste time ever time when we have full array.

src/main/java/core/basesyntax/ArrayList.java
kenu21: No, we don't need it. I will fix.

src/main/java/core/basesyntax/ArrayList.java
kenu21: Yes, I will fix.

src/main/java/core/basesyntax/ArrayList.java
kenu21: You said me at retrospective that 8 and 16 are good number for arrays :-) 
I don't have any "elements". What did you mean?

PR #24, opened 2019-08-29
added solution with my implementation of ArrayList

kozhukhovsky:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,52 +1,112 @@
...
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    private static final int DEFAULT_CAPACITY = 10;
+
+    private Object[] data;
+    private int currentCapacity;
+    private int pointer;
```
boroda4436: @kozhukhovsky it was better to name this variable size

kozhukhovsky: ok, thx

PR #34, opened 2019-12-11
7.1 is done

dmytro-hryhoruk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY];
+    }
+
+    public void ensureCapacity() {
+        Object[] newElementData = Arrays.copyOf(elementData, size + 1);
```
Tarabarnik: new size should be 1,5 times bigger as previous. you should resize your array only when it's full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
+        elementData = newElementData;
+    }
+
+    public void rangeCheckForAdd(int index) {
```
Tarabarnik: suggested name: checkIndex

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        rangeCheckForAdd(index);
+        ensureCapacity();
+        Object[] tempData = new Object[elementData.length + 1];
+        System.arraycopy(elementData, 0, tempData, 0, index);
```
Tarabarnik: try to solve with single arraycopy call and no temporary arrays

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
-
+        rangeCheckForAdd(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        rangeCheckForAdd(index);
+        Object[] tempData = new Object[elementData.length];
```
Tarabarnik: try to solve with single arraycopy call and no temporary arrays

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
+        for (int i = 0; i < elementData.length; i++) {
+            if (t != null && t.equals(elementData[i]) || t == elementData[i]) {
+                index = i;
+                elementExist = true;
+            }
+        }
+        if (!elementExist) {
+            throw new NoSuchElementException("There is no such element");
+        }
+        return remove(index);
```
Tarabarnik: you can use this return in a loop to simplify

PR #44, opened 2019-12-12
hw7.1

Mykhaylo12:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+    private int amountOfElements = 0;
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_CAPACITY];
+    }
 
     @Override
     public void add(T value) {
-
+        if (amountOfElements == array.length - 1) {
```
Tarabarnik:
```suggestion
        if (amountOfElements == array.length) {
```
this might work

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+            array[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > amountOfElements || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
+        } else {
```
Tarabarnik: do wee need else after throw?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        if (amountOfElements == 0) {
+            return true;
+        } else {
+            return false;
+        }
```
Tarabarnik: `return amountOfElements == 0;`

src/main/java/core/basesyntax/ArrayList.java
Mykhaylo12: got it

src/main/java/core/basesyntax/ArrayList.java
Mykhaylo12: ok

PR #58, opened 2020-04-07
added solution

Nick97-git:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        T[] oldArr = elementArr;
+        elementArr = (T[]) new Object[newCapacity];
+        System.arraycopy(oldArr,0,elementArr,0,size);
+    }
+
+    private boolean isInArray(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        return true;
```
RomanDubovskyi, line 99: Optional: 
actually you don't need to return boolean here, you can just make this method void and in case something wrong it will throw exception anyway

PR #65, opened 2020-04-07
implemented ArrayList

Handsome0711:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+            size++;
+        } else {
+            throw new ArrayIndexOutOfBoundsException();
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int resultLength = size + list.size();
```
RomanDubovskyi, line 44: you can solve it with just one for loop and call you add() method for each element of input list, it will be much shorter

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        }
+        throw new ArrayIndexOutOfBoundsException();
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < size && index >= 0) {
+            elements[index] = value;
+        } else {
```
RomanDubovskyi, line 68: you can do it without else block like in add method

PR #67, opened 2020-04-07
Added solution.

Merkald:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_CAPASITY = 10;
```
RomanDubovskyi, line 10:
```suggestion
    private static final int DEFAULT_ARRAY_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+    private Object[] elementData;
+    private int arrayCapasity;
```
RomanDubovskyi, line 12: you don't need separate variable for capacity, cause your capacity is always elementData.length

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+                ammountOfElements++;
+            } else {
+                Object[] oldData = elementData;
+                newArrayCapasity(arrayCapasity);
+                System.arraycopy(oldData, index, elementData,
+                        index + 1, ammountOfElements + 1 - index);
+                elementData[index] = value;
+                ammountOfElements++;
+            }
+        } else {
```
RomanDubovskyi, line 49: this else block is redundant, try to make 
if (check index if it fits) { 
     throw new Array....
}
and you won't need else then

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+                    + ammountOfElements + "<" + "index: " + index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (ensureCapacity(ammountOfElements + list.size())) {
+            for (int i = 0; i < list.size(); i++) {
+                elementData[ammountOfElements++] = list.get(i);
```
RomanDubovskyi, line 59: you can make it shorter: try to execute for loop for list and call your add() method for each of element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+            }
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < ammountOfElements) {
+            return (T) elementData[index];
+        } else {
```
RomanDubovskyi, line 75: same here, this else block is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+                    + ammountOfElements + "<" + "index: " + index);
+        }
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < ammountOfElements || index < 0) {
+            elementData[index] = value;
+        } else {
```
RomanDubovskyi, line 85: and here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,138 @@
...
+        }
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < ammountOfElements; i++) {
+            if ((t == null && elementData[i] == null)
+                    || (elementData[i] != null && elementData[i].equals(t))) {
+                int numMoved = ammountOfElements - i - 1;
```
RomanDubovskyi, line 110: when you found your element just call your remove(int index) method, cause same code repeats

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,135 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        if (ensureCapacity(ammountOfElements + list.size())) {
+            for (int i = 0; i < list.size(); i++) {
+                add(list.get(i));
+            }
+        } else {
+            Object[] oldData = elementData;
+            newArrayCapasity(ammountOfElements + list.size());
```
RomanDubovskyi, line 60: you don't need to do ensureCapacity here, you're doing it inside add() method anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,135 @@
...
     public boolean isEmpty() {
-        return false;
+        return ammountOfElements == 0;
+    }
+
+    private boolean ensureCapacity(int size) {
+        return this.elementData.length > size;
+    }
+
+    private void newArrayCapasity(int oldCapasity) {
```
RomanDubovskyi, line 124:
```suggestion
    private void newArrayCapacity(int oldCapacity) {
```
fix it everywhere

PR #68, opened 2020-04-07
7.1

ana-dav:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
- * реалізації ArrayList (default capacity, newCapacity...)</p>
- */
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE;
+    transient Object[] elementData;
```
RomanDubovskyi, line 10: do we really need transient here ? 
https://www.geeksforgeeks.org/transient-keyword-java/
and what about access modifier for elementData, should it be default?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+    private int size;
```
RomanDubovskyi, line 11: it's better to initialize this field inside of constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+        } else {
+            elementData = new Object[initialCapacity];
+        }
+    }
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
+    private void resize() {
```
RomanDubovskyi, line 27: usually we locate private method after public in the end

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+        if (size >= MAX_ARRAY_SIZE) {
+            throw new OutOfMemoryError("out of memory");
+        }
+        if (size == elementData.length) {
+            int newCapacity = 0;
+            if (size == elementData.length) {
```
RomanDubovskyi, line 33: this second check is just duplicate of 31 line, i don't think we need it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+        if (index > elementData.length) {
+            throw new IndexOutOfBoundsException("Put correct index");
+        }
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        //petezatiraet
```
RomanDubovskyi, line 83: let's remove these comments before sending solution )

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        rangeCheck(index);
+        if (index > elementData.length) {
```
RomanDubovskyi, line 75: i think this check after rangeCheck() is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,113 @@
...
     public void add(T value) {
-
+        resize();
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        //add elem on the last place - ne perezatiraet
```
RomanDubovskyi, line 43: this comment as well

PR #70, opened 2020-04-07
first commit, added solution

maks45:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
-
+        validateIndex(index);
+        objects[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        Object[] decreasedObjects = new Object[objects.length - 1];
```
RomanDubovskyi, line 63: my bad, i confused you on QnA. Actual java.util.Arraylist doesn't trim itself after you remove object from it, its expensive for performance. That's just for you info.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+        return (T) removableObject;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if ((objects[i] != null && ((T) objects[i]).equals(t)) || objects[i] == t) {
+                index = i;
```
RomanDubovskyi, line 80:
```suggestion
            return remove(i);
```
this  should be enough, so you don't need index variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        validateIndex(index);
+        Object[] increasedObjects = new Object[objects.length + 1];
+        if (index > 0) {
+            System.arraycopy(objects, 0, increasedObjects, 0, index);
```
RomanDubovskyi, line 33: it should be enough to use instead of 31-34 line 
```
if (objects.length == size) {
            increaseObjectsLength();
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+        }
+        System.arraycopy(objects, index, increasedObjects, index + 1, objects.length - index);
```
RomanDubovskyi, line 35:
```suggestion
        System.arraycopy(objects, index, objects, index + 1, size - index);
```

src/main/java/core/basesyntax/ArrayList.java
maks45: Ok, thank you. I will keep it in mind.

src/main/java/core/basesyntax/ArrayList.java
maks45: done

src/main/java/core/basesyntax/ArrayList.java
maks45: done

src/main/java/core/basesyntax/ArrayList.java
maks45: done

PR #72, opened 2020-04-07
implementation class ArrayList

KrohIgor:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    private static final int CAPACITY = 10;
+
+    private T[] arrayList;
+    private int size = 0;
```
RomanDubovskyi, line 14: it's better to initialize it inside of constructor

KrohIgor: ok, thank you.

PR #75, opened 2020-04-07
added solution

AlexeyBavika:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] values;
+    private int size;
```
RomanDubovskyi, line 12: it's better to initialize this variable inside of constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+
+    public ArrayList() {
+        this.values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int initialCapacity) {
```
RomanDubovskyi, line 18: you don't really need second constructor in this task

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        checkIfIndexIsNotOutOfBound(index);
+        T oldValue = get(index);
+        decrease(index);
+        return oldValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        if (t == null) {
```
RomanDubovskyi, line 73: you can optimize it with just one loop 
```
            if (values[i] == t || values[i] !=null && values[i].equals(t)) {
                return remove(i);
            }
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -7,22 +7,13 @@
...
-
-    public ArrayList(int initialCapacity) {
-        if (initialCapacity > 0) {
-            this.values = (T[]) new Object[initialCapacity];
-        } else if (initialCapacity == 0) {
-            this.values = (T[]) new Object[0];
-        } else {
-            throw new IllegalArgumentException("Wrong argument : " + initialCapacity);
-        }
+        defaultCapacity = 10;
```
RomanDubovskyi, line 15: i was talking about size variable
default capcity should be constant as it was

AlexeyBavika: understood

PR #81, opened 2020-04-07
ArrayList implemented

NikolayAbysov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
 
+import java.util.NoSuchElementException;
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
+
 public class ArrayList<T> implements List<T> {
 
+    private static final int INITIAL_CAPACITY = 1;
```
RomanDubovskyi, line 11:
```suggestion
    private static final int INITIAL_CAPACITY = 10;
```
let's make capacity same as in original Arraylist

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
+    private int elementNum = 0;
```
RomanDubovskyi, line 12: it's better to initialize non constant fields inside of constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
+            throw new ArrayIndexOutOfBoundsException();
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        capacityCheck();
+        list[elementNum] = value;
+        elementNum = elementNum + 1;
```
RomanDubovskyi, line 41:
```suggestion
        elementNum++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexCheck(index);
+        capacityCheck();
+        System.arraycopy(list, index, list, index + 1, elementNum - index);
+        list[index] = value;
+        elementNum = elementNum + 1;
```
RomanDubovskyi, line 50:
```suggestion
        elementNum++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
+        return (T) list[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        indexCheck(index);
+        if (list[index] == null) {
+            list[index] = value;
+            elementNum = elementNum + 1;
```
RomanDubovskyi, line 71: with set you don't need to new element and increment your size 
you just need to override value, so you can remove 69-72 rows

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexCheck(index);
+        Object[] trimmed = new Object[elementNum - 1];
+        T removedElement = (T) list[index];
+        if (index == elementNum - 1) {
```
RomanDubovskyi, line 82: here is my bad, i gave you wrong answer: actual arraylist doesn't trim it's capacity after remove, that's for your information. But you can leave as it is this task will be counted

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,129 @@
...
-        return null;
+        int position = 0;
+
+        if (t == null) {
+            for (int i = 0; i < elementNum; i++) {
+                if (list[i] == null) {
+                    remove(i);
+                    return null;
+                }
+            }
```
RomanDubovskyi, line 104: try to optimize it like this: 
 ```
            if (list[i] == t || list[i].equals(t)) {
                return remove(i);
            }
        }
```

File: src/main/java/core/basesyntax/ArrayList.java (hunk 1 above)
Comment (NikolayAbysov) [reply src/main/java/core/basesyntax/ArrayList.java:11]: Sure, that's was for testing purposes :)

File: src/main/java/core/basesyntax/ArrayList.java [hunk 8]
```diff
@@ -1,53 +1,128 @@
...
     @Override
     public void set(T value, int index) {
-
+        indexCheck(index);
+        if (list[index] == null) {
+            list[index] = value;
+            elementNum++;
+        } else {
+            list[index] = value;
+        }
```
RomanDubovskyi, line 76:
```suggestion
        indexCheck(index);
         list[index] = value;
        }
```

RomanDubovskyi: set method only overrides value, you use add with index to insert value to particular index

NikolayAbysov: O, sure :)

src/main/java/core/basesyntax/ArrayList.java
RomanDubovskyi: i don't see you fixed this comment

NikolayAbysov:
Sorry, it seems i missed that :( Now it 100% fixed

PR #83, opened 2020-04-07
hw_7_1_done

IrinaDyakonova:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private int size = 0;
+    private T [] arrayList;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[CAPACITY];
+    }
+
+    public ArrayList(int size) {
```
RomanDubovskyi, line 18: did you test if this constructor work? 
it's not neccessary for this task, so i suggest we remove it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
+        ensureCapacity(size);
+    }
+
+    public void ensureCapacity(int size) {
+        if (size >= CAPACITY) {
```
RomanDubovskyi, line 23: CAPACITY will always stay 10, meaning you will do resize every add() method after your size is bigger than that
```suggestion
        if (size >= arryList.length) {
```

src/main/java/core/basesyntax/ArrayList.java
IrinaDyakonova: this constructor is optional for tests, but in general it should be in just the arraylist class

src/main/java/core/basesyntax/ArrayList.java
IrinaDyakonova: yes, it makes no sense to compare with the original long constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+    }
+
+    private void assertIndex(int index) {
+        if (index < 0 || index >= size()) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
     }
+
 }
+
```
RomanDubovskyi, line 105: unneeded empty lines

PR #85, opened 2020-04-07
HW-7.1

serg-ksv:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,118 @@
...
+        checkIndex(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedElement = elementData[index];
+        resize(elementData, index, size, false);
```
RomanDubovskyi, line 62: here is my bad, i gave you wrong answer: actual arraylist doesn't trim it's capacity after remove, that's for your information. But you can leave as it is this task will be counted

PR #89, opened 2020-04-07
result

MarberyUA:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,130 @@
...
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    static final Integer MIN_CAPACITY = 10;
```
RomanDubovskyi, line 11: should here be a default access modifier?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,130 @@
...
+            System.arraycopy(cloneOfArray, index, this.array, index + 1, elementData - index - 1);
+        } else if (index == elementData) {
+            this.add(value);
+            elementData += 1;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Index: " + index + " Size: " + elementData);
+        }
+    }
 
+    private void grow(int lengthOfElements, Integer length) {
```
RomanDubovskyi, line 42: you don't need to pass here anything, you have access to your array.length anyway, so you just can increase it in 1.5 time when you need to

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,130 @@
...
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (this.array.length < elementData + list.size()) {
+            grow(list.size(), this.array.length);
+        }
+        T[] bufferedArray = (T[]) new Object[list.size()];
```
RomanDubovskyi, line 55: you can just iterate list with for i loop and call add() method for each element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,130 @@
...
+        }
+        throw new ArrayIndexOutOfBoundsException("Index: " + index + " Size: " + elementData);
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < elementData) {
+            this.array[index] = value;
+        } else {
```
RomanDubovskyi, line 75: this else block is redundant 
if (check) { 
  throw new ....
} 
will be enough

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,130 @@
...
+                return result;
+            }
+        }
+        throw new NoSuchElementException("No such element in array.");
+    }
+
+    private Integer cutting(int index) {
+        if (index < elementData) {
+            T[] bufferedMassive = this.array.clone();
+            this.array = (T[]) new Object[this.array.length - 1];
```
RomanDubovskyi, line 101: here is my bad, i gave you wrong answer: actual arraylist doesn't trim it's capacity after remove, that's for your information. But you can leave as it is this task will be counted

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,133 @@
...
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    private static final Integer MIN_CAPACITY = 10;
```
RomanDubovskyi, line 11:
```suggestion
    private static final int MIN_CAPACITY = 10;
```

RomanDubovskyi: do we have any reason to use wrapper instead of primitive?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,133 @@
...
+    private T[] array;
+    private int elementDat;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[MIN_CAPACITY];
```
RomanDubovskyi, line 16: btw your code will work without "this."

PR #103, opened 2020-04-08
added solution

vbortnyk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,107 @@
...
+        dataArray[index] = value;
+        size++;
 
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (dataArray.length - size < list.size()) {
+            optimizeCapacity();
```
RomanDubovskyi, line 46: do we need it here?

src/main/java/core/basesyntax/List.java
```diff
@@ -14,7 +16,7 @@
 
     T remove(int index);
 
-    T remove(T t);
+    T remove(T t) throws FileNotFoundException;
```
RomanDubovskyi, line 19: should this method have this exception in signature?

VBortnyk: I don't even know how it appeared there)

RomanDubovskyi (approved):

PR #109, opened 2020-08-05
added solution

tarkaiev:
(sorry for that code)

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,53 +1,117 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+    private int index;
+
+    public ArrayList() {
+        elements = new Object[INITIAL_CAPACITY];
+    }
+
+    private void growArray() {
+        Object[] newArray = new Object[elements.length + elements.length / 2 + 1];
+        System.arraycopy(elements, 0, newArray, 0, elements.length);
+        elements = newArray;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= this.index) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+    }
 
     @Override
     public void add(T value) {
+        if (index == elements.length) {
+            growArray();
+        }
+        elements[index] = value;
+        index++;
+        size++;
 
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == elements.length) {
+            growArray();
+        }
+        System.arraycopy(elements, index, elements, index + 1, this.index - index);
+        elements[index] = value;
+        this.index++;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
+        checkIndex(index);
+        elements[index] = value;
+    }
 
+    private void removeAction(int index) {
+        System.arraycopy(elements, index + 1, elements, index, this.index - index);
+        size--;
+        this.index--;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T result = (T) elements[index];
+        removeAction(index);
+        return result;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        if (t == null) {
+            size--;
+            this.index--;
+            return null;
+        }
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (t == elements[i] || t.equals(elements[i])) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+    private int index;
```
dmytro-hryhoruk, line 13: do you need both of these variables?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+
+    private void growArray() {
+        Object[] newArray = new Object[elements.length + elements.length / 2];
+        System.arraycopy(elements, 0, newArray, 0, elements.length);
+        elements = newArray;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= this.index) {
+            throw new ArrayIndexOutOfBoundsException();
```
dmytro-hryhoruk, line 27: let's add message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+        }
+    }
 
     @Override
     public void add(T value) {
+        if (index == elements.length) {
+            growArray();
+        }
```
dmytro-hryhoruk, line 35: we can move this check inside this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,117 @@
...
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException();
+        }
+        T result = (T) elements[index];
+        removeAction(index);
+        return result;
```
dmytro-hryhoruk, line 105: https://mate-academy.github.io/jv-program-fulltime/01_core/checklist/hw_08.html

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: Actually, we don't.
Made it for better code understanding, forgot to remove.
Fixed

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: okay

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: moved

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: refactored

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
-
+        growArray();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == elements.length) {
```
dmytro-hryhoruk, line 27: let's add checking for index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("No such element in array");
+        }
+        T result = (T) elements[index];
+        removeAction(index);
+        return result;
```
dmytro-hryhoruk, line 77:
```suggestion
                return remove(i);
            }
        }
        throw new NoSuchElementException("No such element in array");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+        checkIndex(index);
+        T result = (T) elements[index];
+        removeAction(index);
+        return result;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
```
dmytro-hryhoruk, line 64:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Index is out of array bounds");
+        }
+    }
+
+    private void removeAction(int index) {
+        System.arraycopy(elements, index + 1, elements, index, size - index);
+        size--;
```
dmytro-hryhoruk, line 106: we can keep that in remove(int index) method

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: Makes sense. fixed

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: added
i had to add different checking6 not just using checkIndex() because index here can be the same value as size

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: done

src/main/java/core/basesyntax/ArrayList.java
tarkaiev: done

PR #113, opened 2020-08-05
done

ilya449:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        T removedElement;
+        for (int i = 0; i < nextEmptyPosition; i++) {
+            if (elements[i] == null ? elements[i] == t : elements[i].equals(t)) {
+                removedElement = elements[i];
+                shiftToTheLeftFrom(i);
+                nextEmptyPosition--;
+                return removedElement;
```
dmytro-hryhoruk, line 77: when you have index of element that you need to remove. Probably from that point we have functionality to do that?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+        return nextEmptyPosition == 0;
+    }
+
+    private void increaseCapacity() {
+        T[] temp = elements;
+        elements = (T[]) new Object[elements.length + (elements.length >> 1)];
+        System.arraycopy(temp, 0, elements, 0, temp.length);
+    }
+
+    private void shiftToTheLeftFrom(int i) {
```
dmytro-hryhoruk, line 99: you have to use this only in one place. No need to make a separate method for this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+        System.arraycopy(elements, i + 1, elements,
+                i, elements.length - i - 1);
+    }
+
+    private void shiftToTheRightFrom(int index) {
```
dmytro-hryhoruk, line 104: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+        nextEmptyPosition++;
+        elements[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (list.size() > elements.length - nextEmptyPosition - 1) {
+            increaseCapacity();
+        }
```
dmytro-hryhoruk, line 40: method add() will check and increase capacity if need.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+
+    private void checkAndIncreaseCapacity() {
+        if (elements.length <= nextEmptyPosition) {
+            increaseCapacity();
+        }
+    }
+
+    private void checkIndexInRange(int index) {
+        if (index < 0 || index >= nextEmptyPosition) {
+            throw new ArrayIndexOutOfBoundsException();
```
dmytro-hryhoruk, line 117: let's add message to exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+        T removedElement;
+        for (int i = 0; i < nextEmptyPosition; i++) {
+            if (elements[i] == null ? elements[i] == t : elements[i].equals(t)) {
+                removedElement = elements[i];
+                shiftToTheLeftFrom(i);
+                nextEmptyPosition--;
+                return removedElement;
+            }
+        }
+        throw new NoSuchElementException();
```
dmytro-hryhoruk, line 80: let's add message to exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,120 @@
...
+        checkIndexInRange(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexInRange(index);
+        T removedElement = elements[index];
+        if (index != nextEmptyPosition) {
```
dmytro-hryhoruk, line 62: try to test this without this check. Maybe it will work and you will just trim the empty space from the end of array

ilya449: yes, it works. it was redundant check

src/main/java/core/basesyntax/ArrayList.java
ilya449: Added

src/main/java/core/basesyntax/ArrayList.java
ilya449: Added

src/main/java/core/basesyntax/ArrayList.java
ilya449: I thought if the list is too big it would be better to increase capacity firstly and then just copy. Was I wrong?

src/main/java/core/basesyntax/ArrayList.java
ilya449: fixed

src/main/java/core/basesyntax/ArrayList.java
ilya449: fixed

src/main/java/core/basesyntax/ArrayList.java
ilya449: changed to ```remove(index)``` method

ilya449:
> Good job!

Thank you!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
+        nextEmptyPosition--;
+        return removedElement;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < nextEmptyPosition; i++) {
+            if (elements[i] == null ? elements[i] == t : elements[i].equals(t)) {
+                return this.remove(i);
```
dmytro-hryhoruk, line 75:
```suggestion
                return remove(i);
```

PR #116, opened 2020-08-05
added solution

MaliukDaria:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
+            increaseCapacity();
+        }
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size) {
```
KseniiaMakarova, line 27: what if index < 0?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
+            throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 28: let's make this more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size) {
```
KseniiaMakarova, line 47: what if index < 0? 
also, we use this check in get(), set() and remove()
let's make a helper private method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
     public void set(T value, int index) {
-
+        if (index >= size) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
```
KseniiaMakarova, line 62: please check index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
-        return null;
+        T value = (T) elements[index];
+        System.arraycopy(elements, index + 1, elements, index, size - index);
```
KseniiaMakarova, line 64:
```suggestion
        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
```
we need to move 1 element less because we are removing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,90 @@
...
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (t == null ? t == elements[i] : t.equals(elements[i])) {
+                return remove(i);
+            }
+        }
+        throw new java.util.NoSuchElementException();
```
KseniiaMakarova, line 76: let's import this class and make the exception informative

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done)

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,98 @@
...
+            increaseCapacity();
+        }
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size) {
```
KseniiaMakarova, line 27: again: what if index < 0? we should also include this condition, right?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,98 @@
...
-
+        indexCheck(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexCheck(index);
+        final T value = (T) elements[index];
```
KseniiaMakarova, line 61:
```suggestion
        T value = (T) elements[index];
```
what was the point of making it final?

MaliukDaria: If I don't make it final, maven check fails:
 ArrayList.java:61: Distance between variable 'value' declaration and its first usage is 4, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageDistance]

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -5,49 +5,98 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[size - 1] = null;
+        size--;
```
KseniiaMakarova, line 64:
```suggestion
        elements[--size] = null;
```
and you will be able to remove `final` :)

MaliukDaria: this is genius! Thank)

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: done

PR #120, opened 2020-08-05
added solution

matviichuk92:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,97 @@
...
     public void add(T value) {
-
+        if (array.length == size) {
+            resize();
+        }
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 29: let's make two separate checks: for index and for capacity

the check for index is slightly different from your checkIndex() method - index may be equal to size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,97 @@
...
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(array[i], t)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
KseniiaMakarova, line 75: let's make this more informative

matviichuk92:
done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,99 @@
...
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Index out of bounds!");
+        }
+    }
+
+    private void resize() {
```
KseniiaMakarova, line 88:
```suggestion
    private void checkSize() {
```

PR #122, opened 2020-08-05
Homework 8 completed

Borisko130:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
+
 public class ArrayList<T> implements List<T> {
 
+    private static final int DEFAULT_CAPACITY = 10;
+
+    private int currentSize;
+    private int capacity;
```
KseniiaMakarova, line 16: you might use `values.length` when you need this value
what do you think?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
+        return currentSize;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentSize == 0;
+    }
+
+    private void growList() {
```
KseniiaMakarova, line 94:
```suggestion
    private void ensureCapacity() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
-
+        growList();
+        values[currentSize++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > currentSize || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 33: let's make this exception informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
+        }
+        System.arraycopy(values, index, values, index + 1, capacity - index - 1);
+        values[index] = value;
+        currentSize++;
+        growList();
```
KseniiaMakarova, line 38: why are we ensuring capacity after we have already added the element?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
+        values[currentSize++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > currentSize || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        System.arraycopy(values, index, values, index + 1, capacity - index - 1);
```
KseniiaMakarova, line 35:
```suggestion
        System.arraycopy(values, index, values, index + 1, ??? );
```
let's review this part

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index > currentSize - 1 || index < 0) {
```
KseniiaMakarova, line 50:
```suggestion
        if (index > currentSize || index < 0) {
```
also, this check is the same for get(), set() and remove()
let's move it to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
+            throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 51: make this informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > currentSize - 1 || index < 0) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        T valueToReturn = values[index];
+        System.arraycopy(values, index + 1, values, index, currentSize--);
```
KseniiaMakarova, line 70:
```suggestion
        System.arraycopy(values, index + 1, values, index,  ??? );
```
please review this part and decrement size on a new line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < currentSize; i++) {
+            if (t == values[i] || (t != null && t.equals(values[i]))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
KseniiaMakarova, line 81: please make this informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T valueToReturn = values[index];
+        int toCopy = currentSize - index - 1;
+        System.arraycopy(values, index + 1, values, index, toCopy);
+        values[currentSize--] = null;
```
KseniiaMakarova, line 66:
```suggestion
        values[--currentSize] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        return currentSize;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentSize == 0;
+    }
+
+    private void checkIndex(int index) {
```
```suggestion
    private void checkIndexForAdd(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        }
+        return values[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= currentSize || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("List has no such index");
+        }
```
KseniiaMakarova, line 56: let's move this to a separate method
we should use the same logic for get(), set() and remove()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
     public void add(T value) {
-
+        ensureCapacity();
+        values[currentSize++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
```
KseniiaMakarova, line 30:
```suggestion
        checkIndexForAdd(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexForAddAndRemove(index);
+        T valueToReturn = values[index];
+        int toCopy = currentSize - index - 1;
+        System.arraycopy(values, index + 1, values, index, toCopy);
+        values[--currentSize] = null;
```
Borisko130, line 62: Changed from 
1) `values[currentSize--] = null;`
to
2) `values[--currentSize] = null;`
But I don't understand why.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
     public void set(T value, int index) {
-
+        checkIndexForGetAndSet(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexForAddAndRemove(index);
```
KseniiaMakarova, line 58: why are you doing the check for `add()` and not the check for `set() and get()`?
is it possible to remove the element from index 4 from the array of size 4?

src/main/java/core/basesyntax/ArrayList.java
KseniiaMakarova: if size is `n`, what will be the value of the last index?

I would suggest you to debug this method in both cases and look what is happening
if it will not be clear after that, please be free to ask again, ok?

Borisko130: I checked it with debugger, but still don't understand.
It does the opposite of what I expected.

PR #126, opened 2020-08-05
Hw8

fyarkiy:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
     public void add(T value) {
-
+        addElementToArray(value, currentIndex);
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > currentIndex) {
+            throw new ArrayIndexOutOfBoundsException(" Index above the size of Array");
```
dmytro-hryhoruk, line 28:
```suggestion
            throw new ArrayIndexOutOfBoundsException("Index above the size of Array");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+        }
+        sizeCheck(index);
+        if (index < currentIndex) {
+            System.arraycopy(arrayList, index, arrayList, index + 1, currentIndex + 1);
```
dmytro-hryhoruk, line 32: put into separate method

src/main/java/core/basesyntax/ArrayList.java
dmytro-hryhoruk, line 28: you have a very similar code for checking index. Try to make a method out of this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+        currentCapacity = currentCapacity + (currentCapacity >> 1);
+        Object[] arrayListTemporary = (T[]) new Object[currentCapacity];
+        System.arraycopy(arrayList, 0, arrayListTemporary, 0, currentIndex);
+        arrayList = (T[]) arrayListTemporary;
+    }
+
+    public void sizeCheck(int sizeForTest) {
+        if (sizeForTest == currentCapacity) {
+            increaseCapacity();
+        }
```
dmytro-hryhoruk, line 105: you can move this check to increaseCapacity() . And think if you really need to pass some parameters to this method to make this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[DEFAULT_CAPACITY];
+        currentCapacity = DEFAULT_CAPACITY;
+    }
 
     @Override
     public void add(T value) {
-
+        addElementToArray(value, currentIndex);
```
dmytro-hryhoruk, line 22: do you need to pass currentIndex as parameter?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+            System.arraycopy(arrayList, index, arrayList, index + 1, currentIndex + 1);
+        }
+        addElementToArray(value, index);
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            addElementToArray(list.get(i), currentIndex);
```
dmytro-hryhoruk, line 40: can we use method add(T value)?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+        return currentIndex;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentIndex == 0;
+    }
+
+    public void addElementToArray(T value, int index) {
```
dmytro-hryhoruk, line 89: you have to use this method only in one place. So we can keep this logic there

src/main/java/core/basesyntax/ArrayList.java
fyarkiy: Yes, just for one example when I need to add a new element somewhere in-between of existing elements. In the rest of cases I don't need it, but I think no point to create another method addElementToArray with one parameter, right?

src/main/java/core/basesyntax/ArrayList.java
fyarkiy: I just thought - on this line I'm just calling a method. Is there a value to have a method that just calls another one? But you are probably right (in case we decided to replace it with our own solution) .

src/main/java/core/basesyntax/ArrayList.java
fyarkiy: Which method? addElementToArray? If it is related to your comment above - I still keep using it in two places when we need to add an element into the middle (or somewhere inside) of the array.

src/main/java/core/basesyntax/ArrayList.java
fyarkiy: Just to clarify - I'm using addElementToArray in two cases. And in the second call (of this method) I need to pass one variable (index) from that method. 
On the 22 line (to which you pointed out) I don't really need to pass currrentIndex but because method expect "index" I have to stipulate it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,104 @@
...
+        sizeCheck();
+        arrayList[index] = value;
+        currentIndex++;
+    }
+
+    public void indexCheck(int index) {
+        if (index > currentIndex) {
```
dmytro-hryhoruk, line 90: less than zero?

PR #127, opened 2020-08-05
Hw 8

pavlokhomiak:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
     public T remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (t == null && elementData[i] == null || elementData[i].equals(t)) {
+                return remove(i);
+            }
+            if (i == size - 1) {
+                throw new NoSuchElementException("Element is not exists");
+            }
+        }
         return null;
```
dmytro-hryhoruk, line 94: try to make 2 cases: 1 you returning value in case remove has been done
2 you are throwing exception

PR #130, opened 2020-08-06
Initial commit

SetMyDream:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity = DEFAULT_CAPACITY;
```
KseniiaMakarova, line 11: do we need this field?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+    private T[] storage;
+    private int current = 0;
```
KseniiaMakarova, line 13: let's initialise non-constant fields in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        if (capacity == current) {
+            storage = resize();
+        }
+        storage[current++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        while (index > capacity) {
```
KseniiaMakarova, line 36: if in any method you receive an index that is larger than your current `storage.length`, you should throw an Exception

src/main/java/core/basesyntax/ArrayList.java
KseniiaMakarova, line 13:
```suggestion
    private int currentSize;
```
what do you think?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            this.add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < current) {
```
KseniiaMakarova, line 53: what if index < 0?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+            return (T) storage[index];
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Index out of bounds");
+        }
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < current) {
```
KseniiaMakarova, line 62: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
     public void add(T value) {
-
+        if (capacity == current) {
+            storage = resize();
+        }
+        storage[current++] = value;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 35: let's check if index is valid here as well?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+            storage[index] = value;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Index out of bounds");
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < current) {
```
KseniiaMakarova, line 71: what if index < 0?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < current; i++) {
+            if (storage[i] == t
+                    || storage[i] != null && storage[i].equals(t)) {
+                index = i;
```
KseniiaMakarova, line 88: let's just put a return statement on this line?

src/main/java/core/basesyntax/ArrayList.java
SetMyDream: Ok, i see

src/main/java/core/basesyntax/ArrayList.java
SetMyDream: don`t know exact reason, but yes:
If I don`t initialize capacity here instead of  constructor, many tests ruins, and i don`t understand why... Also i using DEFAULT_CAPACITY, not number directly because of checklist: 
Default capacity of the array should be imaged as a constant field.

SetMyDream: Found how to deal with

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
+            storage = resize();
+        }
+        storage[currentSize++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > currentSize || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("");
```
KseniiaMakarova, line 30: pass some informative message to exception constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
+        }
+        if (storage.length == currentSize) {
+            resize();
```
KseniiaMakarova, line 33: what does `resize()` return?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
+    private T[] resize() {
+        T[] newArray = (T[]) new Object[(storage.length * 3) / 2 + 1];
+        System.arraycopy(storage, 0, newArray, 0, currentSize);
+        return newArray;
+    }
+
+    private void outOfBoundsCheck(int index) {
+        if (index >= currentSize || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("Index " + index
+                    + " out of bounds of " + storage.length);
```
KseniiaMakarova, line 98:
```suggestion
                    + " out of bounds of " + currentSize);
```

SetMyDream: done

PR #131, opened 2020-08-06
added solution

AnaLanda:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+        return counter;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return counter == 0;
+    }
+
+    private void resize() {
```
```suggestion
    private void checkCapacity() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+        if (counter >= DEFAULT_CAPACITY) {
```
KseniiaMakarova, line 94: if you have already resized once your `values`, will this work for the second time

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
-
+        resize();
+        values[counter] = value;
+        counter++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0) {
```
KseniiaMakarova, line 28: if index > size we should also throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= counter || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("This index is out of bounds.");
+        }
```
KseniiaMakarova, line 48: let's move this to a separate method - you use this check in get(), set() and remove()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= counter || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("This index is out of bounds.");
+        }
+        T removedItem = values[index];
+        for (int i = index; i < counter - 1; i++) {
+            values[i] = values[i + 1];
+        }
```
KseniiaMakarova, line 68: let's use System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+        counter--;
+        return removedItem;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < counter; i++) {
+            if (values[i] == t || t != null && values[i].equals(t)) {
```
KseniiaMakarova, line 76: there is a risk of NPE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
 
     @Override
     public T remove(T t) {
-        return null;
+        if (t == null) {
+            counter--;
+            return null;
+        }
+        for (int i = 0; i < counter; i++) {
+            if (values[i] == t || t != null && values[i].equals(t)) {
```
KseniiaMakarova, line 74: the risk of NullPointerException is still here
you are calling `values[i].equals(t)`
is it safe?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return counter == 0;
+    }
+
+    private void checkCapacity() {
+        if (counter >= currentCapacity) {
+            currentCapacity = counter * 3 / 2 + 1;
```
```suggestion
            currentCapacity = values.length * 3 / 2 + 1;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
+            Object[] newValues = new Object[currentCapacity];
+            System.arraycopy(values, 0, newValues, 0, counter);
+            values = (T[]) newValues;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index >= counter) {
```
KseniiaMakarova, line 101: let's add another condition - `|| index < 0`

src/main/java/core/basesyntax/ArrayList.java
AnaLanda: Ohhh. Will this remove the risk of NPE? 
if (values[i] == t || values[i] != null && values[i].equals(t))
I need to make sure that the thing I'm calling equals() on is not null, right?

KseniiaMakarova: yes, you got it!)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,98 @@
...
+        checkIndex(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedItem = values[index];
+        System.arraycopy(values,index + 1, values, index, counter - index - 1);
```
KseniiaMakarova, line 60:
```suggestion
        System.arraycopy(values, index + 1, values, index, counter - index - 1);
```

PR #133, opened 2020-08-06
Add solution

OleksiiTromsa:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
- * реалізації ArrayList (default capacity, newCapacity...)</p>
- */
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+    private int capacity;
```
dmytro-hryhoruk, line 10: you can always get capacity from elementData

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
```
dmytro-hryhoruk, line 51: you are making the same check twice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
+        ensureCapacity(size);
+        System.arraycopy(elementData, index, elementData, index + 1, size - index - 1);
+        elementData[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        size += list.size();
+        ensureCapacity(size);
```
dmytro-hryhoruk, line 61: you have capacity of 10. You should add a another list with the size of 10. Total capacity should be 20, right?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elementData[i], t)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
dmytro-hryhoruk, line 97: let's add message

src/main/java/core/basesyntax/ArrayList.java
OleksiiTromsa: In my case it will be checked if capacity >20, if yes - ok, if no - it will be increased according to the formula: (20 * 3) / 2 + 1. 
I don't now how is it implemented in real ArrayList and don't want to see it until this task will be done.  And there were no instructions how should it be done. So I implemented it in this way.

src/main/java/core/basesyntax/ArrayList.java
OleksiiTromsa: fixed

src/main/java/core/basesyntax/ArrayList.java
OleksiiTromsa: fixed

src/main/java/core/basesyntax/ArrayList.java
OleksiiTromsa: It is not the same check.
Method add(T value, int index) is the only method for which (index = size) is valid. For all others should be (index < size). My method checkIndex() is universal for all methods except add(T value, int index). So I added one additional check to this method.

dmytro-hryhoruk: agree

PR #134, opened 2020-08-06
8 task solution

gradiviro:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] list;
+    private int index;
```
KseniiaMakarova, line 12: this variable has no connection to ArrayList
let's remove this field and refactor our solution using `size` and `list.length` values

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
+        index = 0;
+    }
 
+    public boolean checkCapacity() {
+        return (index >= list.length);
     }
 
-    @Override
-    public void add(T value, int index) {
+    public void resize() {
```
KseniiaMakarova, line 25:
```suggestion
    private void resize() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
 
-    @Override
-    public void add(T value) {
+    public ArrayList() {
+        list = new Object[DEFAULT_CAPACITY];
+        size = 0;
+        index = 0;
+    }
 
+    public boolean checkCapacity() {
```
KseniiaMakarova, line 21:
```suggestion
    private boolean checkCapacity() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
 
-    @Override
-    public void add(T value, int index) {
+    public void resize() {
+        Object[] newList = new Object[list.length + (list.length / 2)];
+        System.arraycopy(list, 0, newList, 0, this.index);
+        list = newList;
+    }
 
+    public void trimArray(int index) {
```
KseniiaMakarova, line 31:
```suggestion
    private void trimArray(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,110 @@
...
+        System.arraycopy(list, index + 1, list, index, this.index - index);
+        size--;
+        this.index--;
     }
 
-    @Override
-    public void addAll(List<T> list) {
+    private void checkIndexBounds(int index) {
+        if (0 > index || index >= this.index) {
+            throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 39: make your exceptions informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
-
+        if (checkCapacity()) {
+            resize();
+        }
+        list[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 29: let's check if index is valid
the logic is similar to that in your `checkIndexBounds()` method, but not the same - `index` may be equal to `size`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        return (T) list[index];
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexBounds(index);
+        T removed = (T) list[index];
+        trimArray(index);
+        list[size] = null;
```
KseniiaMakarova, line 62:
```suggestion
        list[--size] = null;
```
we should decrement the size, right?

KseniiaMakarova: I can see that you are decrementing it in `trim()`
nice!

PR #136, opened 2020-08-06
Implemented solution for ArrayList

andriylagun:


andriylagun:
At my computer tests are passed, what's wrong?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,99 @@
...
+            resize();
+        }
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index >= elements.length) {
```
KseniiaMakarova, line 32: make two separate checks: one for index (if it is valid) and the other for capacity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,99 @@
...
+        checkIndex(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T element = (T) elements[index];
+        System.arraycopy(elements, index + 1, elements, index, size - index);
```
KseniiaMakarova, line 63:
```suggestion
        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
```
we should move one less element because we are removing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        ensureCapacity(size);
+        checkIndex(index);
```
KseniiaMakarova, line 35:
```suggestion
        checkIndex(index);
        ensureCapacity(size);
```
it will be better to first check the index so that we could not make unnecessary resize

PR #140, opened 2020-08-06
hw

YaTut1901:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,123 @@
...
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index >= values.length) {
+            throw new java.lang.ArrayIndexOutOfBoundsException();
+        }
```
dmytro-hryhoruk, line 34: let's make a separate method for this check

dmytro-hryhoruk: what if i want to add element to the last index which will be equal values.length

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,123 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(t, values[i])) {
+                T[] arr = (T[])new Object[values.length];
+                System.arraycopy(values, 0, arr, 0, i);
+                System.arraycopy(values, i + 1, arr, i, size - i - 1);
+                values = arr;
+                size--;
```
dmytro-hryhoruk, line 91: if you have index , should we repeat code?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,123 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int firstEmptyCell() {
```
dmytro-hryhoruk, line 108: Do you need this method?

src/main/java/core/basesyntax/ArrayList.java
dmytro-hryhoruk: ?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        isValid(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isValid(index);
+        T element = values[index];
+        System.arraycopy(values, index + 1, values, index, size - index);
```
KseniiaMakarova, line 62:
```suggestion
        System.arraycopy(values, index + 1, values, index, size - index - 1);
```
we should move one element less here because we are removing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(t, values[i])) {
+                System.arraycopy(values, i + 1, values, i, size - i);
+                size--;
+                return t;
```
KseniiaMakarova, line 73: let's use `remove(i)` instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+            }
+        }
+        throw new NoSuchElementException();
```
KseniiaMakarova, line 76: please add message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void isValid(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 91: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        }
     }
+
+    private void extend() {
```
KseniiaMakarova, line 95:
```suggestion
    private void checkCapacity() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        if (size == values.length) {
+            T[] extendedValues = (T[]) new Object[(int) Math.ceil(values.length * 2)];
```
KseniiaMakarova, line 97: should we increase `values.length` twice?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void isValid(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Элемента с таким индексом нет в списке!");
```
RomanDubovskyi, line 90: eng please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(t, values[i])) {
+                remove(i);
+                return t;
+            }
+        }
+        throw new NoSuchElementException("Такого элемента нет в списке!");
```
RomanDubovskyi, line 75: same here and everywhere

PR #141, opened 2020-08-06
[ADD] solution hw-8.1

vicnikolaev:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+        myList = (T[]) new Object[CAPACITY];
+        sizeList = 0;
+    }
 
     @Override
     public void add(T value) {
-
+        if (sizeList >= myList.length) {
+            resizeList();
+        }
```
dmytro-hryhoruk, line 24: let's move this check to this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,100 @@
...
+        myList[sizeList] = value;
+        sizeList += 1;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (sizeList >= myList.length) {
+            resizeList();
```
dmytro-hryhoruk, line 32: what if i want to add to the invalid index?

vicnikolaev: done

src/main/java/core/basesyntax/ArrayList.java
vicnikolaev: done

src/main/java/core/basesyntax/ArrayList.java
vicnikolaev: done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,101 @@
...
-
+        checkResizeList();
+        myList[sizeList] = value;
+        sizeList += 1;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0) {
```
dmytro-hryhoruk, line 29: what if my size = 1 and i want to put an element to 20 index?

vicnikolaev: done

PR #147, opened 2020-08-06
task done

yutsevich:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    private static final int INITIAL_CAPACITY = 10;
+    private int size = 0;
+    private T[] arrayList = (T[]) new Object[INITIAL_CAPACITY];
```
dmytro-hryhoruk, line 14: let's make this in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+
     @Override
     public void add(T value) {
-
+        if (size == arrayList.length - 1) {
```
dmytro-hryhoruk, line 18:
```suggestion
        if (size == arrayList.length) {
```

dmytro-hryhoruk: can we move this check to the method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (t == null) {
+                System.arraycopy(arrayList, i + 1, arrayList, i, size - i);
+                size--;
+                return null;
+            }
+            if (t.equals(arrayList[i])) {
```
dmytro-hryhoruk, line 78: let's combine these conditions

dmytro-hryhoruk: can we use remove(index) here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexCheck(index);
+
+        T removedItem = arrayList[index];
+
+        System.arraycopy(arrayList, index + 1, arrayList, index, size - index);
```
KseniiaMakarova, line 66:
```suggestion
        System.arraycopy(arrayList, index + 1, arrayList, index, size - index - 1);
```
we should move one less element here because we are removing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resize() {
```
KseniiaMakarova, line 91:
```suggestion
    private void checkSize() {
```
you will resize if a certain condition is met, right?

PR #148, opened 2020-08-06
Hw8

voloshynamaryna11:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
 
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (!(index >= arraySize || index < 0)) {
```
KseniiaMakarova, line 51: you are making the same check in get(), put() and remove()
let's move it to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+            values[--arraySize] = null;
+            return removed;
+        }
+        throw new ArrayIndexOutOfBoundsException("wrong index");
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < values.length; i++) {
```
KseniiaMakarova, line 79:
```suggestion
        for (int i = 0; i < arraySize; i++) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+            if ((T) values[i] == t || t != null && t.equals((T) values[i])) {
```
KseniiaMakarova, line 80:
```suggestion
            if (values[i] == t || t != null && t.equals(values[i])) {
```
value is still an instance of type T, it doesn't matter that it is stored in Object[]

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+                int numMoved = arraySize - i - 1;
+                System.arraycopy(values, i + 1, values, i, numMoved);
+                values[--arraySize] = null;
+                return t;
```
KseniiaMakarova, line 84: let's make use of the code that we already have

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,111 @@
...
+            }
+        }
+        throw new NoSuchElementException();
```
KseniiaMakarova, line 87: please make this more informative

PR #150, opened 2020-08-06
HW-8

vremere-a:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
 
+import java.util.NoSuchElementException;
+
 /**
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
 public class ArrayList<T> implements List<T> {
 
+    static final int DEFAULT_CAPACITY = 10;
```
KseniiaMakarova, line 11: what is wrong here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
+        if (size == elementData.length) {
+            grow();
+        }
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > elementData.length) {
```
KseniiaMakarova, line 29:
```suggestion
        if (index < 0 || index > size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        } else {
```
KseniiaMakarova, line 31: you don't need `else` in this case

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
+            if (size == elementData.length) {
+                grow();
+            } else {
```
KseniiaMakarova, line 34: why using `else`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
+    //// как создать проверку на не существующий объект
```
KseniiaMakarova, line 49: let's remove this line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
     @Override
     public T get(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        }
```
KseniiaMakarova, line 54: the same check is used in put(), get() and remove()
move it to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
+        for (int i = 0; i < elementData.length; i++) {
+            if (elementData[i] == elementData[index]) {
+                return elementData[i];
+            }
+        }
         return null;
```
KseniiaMakarova, line 60: this is too complex
you have direct access to the value in your `elementData` array by index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        }
+        for (int i = 0; i < elementData.length; i++) {
+            if (elementData[i] == elementData[index]) {
+                elementData[index] = value;
+            }
+        }
```
KseniiaMakarova, line 72: same here: too complex

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,116 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        T temp;
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        }
+        temp = elementData[index];
+        elementData[index] = null;
```
KseniiaMakarova, line 82:
```suggestion
        T temp = elementData[index];
```
we don't need line 82 - we will shift values to the left on the next line, so the value will be changed

src/main/java/core/basesyntax/ArrayList.java
vremere-a: done. Thanks. I didn't even think

src/main/java/core/basesyntax/ArrayList.java
vremere-a: ready

src/main/java/core/basesyntax/ArrayList.java
vremere-a: I understand. done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        return size == 0;
     }
+
+    private void grow() {
+        T[] newElementData = (T[]) new Object[elementData.length + (elementData.length >> 1)];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        elementData = newElementData;
+    }
+
+    private void checkIndex(int index) {
```
```suggestion
    private void checkIndexForAdd(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        }
+    }
+
+    private void checkIndexWithEquel(int index) {
```
KseniiaMakarova, line 99:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,106 @@
...
+        checkIndexWithEquel(index);
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Your ArrayIndex are out of bounds");
+        }
```
KseniiaMakarova, line 54: do you have this logic in some method already?

PR #153, opened 2020-08-06
added solution

vladislavshvets:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,102 @@
...
+    }
+
+    public ArrayList(int capacity) {
+        array = (T[]) new Object[capacity];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= ((int) array.length)) {
```
dmytro-hryhoruk, line 26: do we need that cast? you may answer in your chat

PR #154, opened 2020-08-06
solution

Vas35:


Vas35:
But why I passed all tests, just have two fails from Maven which one I didn't understand...?

Vas35:
I fixed already all bugs in Maven and got  BUILD SUCCESS!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,125 @@
...
-/**
+import java.util.NoSuchElementException;
+
+/*
  * <p>Реалізувати свій ArrayList який імплементує інтерфейс List. Дотриматися основних вимог щодо
  * реалізації ArrayList (default capacity, newCapacity...)</p>
  */
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPASITY = 10;
```
```suggestion
    private static final int CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,125 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public T[] resize() {
```
KseniiaMakarova, line 94:
```suggestion
    private T[] resize() {
```
please make all "helper" methods private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,125 @@
...
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index > size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException(
+                    "This index is not suitable for the given array");
+        }
```
KseniiaMakarova, line 36: this check is all right
but on line 29 you already cover the case if `index == size`
so, here it is possible to just call `checkIndex()` instead of lines 33-36

is it clear?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,125 @@
...
+        arrayGeneric[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() >= arrayGeneric.length) {
+            arrayGeneric = resize();
+        }
```
KseniiaMakarova, line 50: do we need this if we are calling `add()` for every element in the list?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,125 @@
...
+    }
+
+    public T[] resize() {
+        T[] currentArray = (T[]) new Object[(arrayGeneric.length * 3) / 2 + 1];
+        System.arraycopy(arrayGeneric, 0, currentArray, 0, size);
+        return currentArray;
     }
+
+    public int findRemoveElement(T element) {
+        for (int i = 0; i < arrayGeneric.length; i++) {
```
KseniiaMakarova, line 101: do we need to iterate over the whole array or only the filled part?

src/main/java/core/basesyntax/ArrayList.java
Vas35: Sorry, here something wrong?

src/main/java/core/basesyntax/ArrayList.java
Vas35: Ok, I will do.

src/main/java/core/basesyntax/ArrayList.java
Vas35: I'm not sure, because Dmytro yesterday on QA  told me didn't touch this "method public void add(T value, int index)"  like here we have other logic that why I didn't touch this method with the same exception?!

src/main/java/core/basesyntax/ArrayList.java
Vas35: Ok, thanks I will change on size

src/main/java/core/basesyntax/ArrayList.java
Vas35: Sorry, what do you mean add() I have this if in addAll(), P.S - I did everything for the implementation as described in the article: https://habr.com/ru/post/128269/

PR #156, opened 2020-08-06
hw8

mikhail771:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
 
     @Override
     public void add(T value) {
-
+        if (!ensureCapacity()) {
+            elementData = grow(size);
+        }
```
KseniiaMakarova, line 24: you have a method that does this check, right?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkCapacity();
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
+        if (ensureCapacity()) {
```
KseniiaMakarova, line 34: you are checking capacity already in this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkCapacity();
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
```
KseniiaMakarova, line 33: let's move index check to the start of this method - before capacity check
we don't want to resize the array if index is not valid

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() > elementData.length) {
+            elementData = grow(size + list.size());
+        }
```
KseniiaMakarova, line 45: do we need this if we call `add()` for each element next?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        checkIndex(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
```
KseniiaMakarova, line 67: let's call `checkIndex()` instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        T removedValue = (T) elementData[index];
+        int newSize;
+        if ((newSize = size - 1) > index) {
```
KseniiaMakarova, line 70:
```suggestion
```
let's try to remove this variable and this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+            System.arraycopy(elementData, index + 1,
+                    elementData, index, newSize - index);
```
KseniiaMakarova, line 72:
```suggestion
                    elementData, index, size - index - 1);
```
we should move one element less because we are removing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,121 @@
...
+        }
+        elementData[size = newSize] = null;
```
KseniiaMakarova, line 74:
```suggestion
        elementData[ ??? ] = null;
```
let's decrement `size` and use it to make the previously last element `null`

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
mikhail771: done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
```
KseniiaMakarova, line 31:
```suggestion
        checkCapacity();
        System.arraycopy(elementData, index, elementData, index + 1, size - index);
```
we need to checkCapacity in all `add()` methods, right?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedValue = (T) elementData[index];
+        System.arraycopy(elementData, index + 1,
+                elementData, index, size - index - 1);
+        elementData[size--] = null;
```
KseniiaMakarova, line 61:
```suggestion
        elementData[--size] = null;
```
try to debug this
you will see that `size` value is different from the last element's index
in case of any questions after debugging, be free to ask!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,109 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean ensureCapacity() {
+        return size + 1 < elementData.length;
+    }
```
KseniiaMakarova, line 89: do we use this method?

PR #157, opened 2020-08-07
solution added

margarita-ignatieva:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,112 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private int numberOfElements;
+    private T[] elementData;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        numberOfElements = 0;
+    }
+
+    public T[] grow() {
```
KseniiaMakarova, line 15:
```suggestion
    private T[] grow() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,112 @@
...
+        T[] arrayT = (T[]) new Object[elementData.length * 2];
```
KseniiaMakarova, line 16: should the capacity increase twice?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,112 @@
...
+        System.arraycopy(elementData, 0, arrayT, 0, numberOfElements);
+        return arrayT;
+
+    }
+
+    public boolean checkIndex(int index) {
```
KseniiaMakarova, line 22:
```suggestion
    private boolean checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,112 @@
...
+        if (index < 0 || index >= numberOfElements) {
+            throw new ArrayIndexOutOfBoundsException();
+        } else {
```
KseniiaMakarova, line 25: we don't need `else` after `throw`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,112 @@
...
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < numberOfElements; i++) {
+            if (elementData[i] == t
+                    || elementData[i] != null && elementData[i].equals(t)) {
+                index = i;
```
KseniiaMakarova, line 90: if we know the index, can we just use some other code that we already have in this class?

margarita-ignatieva: yes, done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,108 @@
...
+            }
+        }
+        if (index != -1) {
+            T value = (T) elementData[index];
+            return value = remove(index);
```
KseniiaMakarova, line 94: let's remove `index` variable from this method
in our loop, when we find a match for the value on lines 87-88, let's call `remove(i)`

margarita-ignatieva: done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,53 +1,105 @@
...
-        return null;
+        int index = -1;
+        for (int i = 0; i < numberOfElements; i++) {
+            if (elementData[i] == t
+                    || elementData[i] != null && elementData[i].equals(t)) {
+                index = i;
+                return t = remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found!");
```
```suggestion
        for (int i = 0; i < numberOfElements; i++) {
            if (elementData[i] == t
                    || elementData[i] != null && elementData[i].equals(t)) {
                return t = remove(i);
            }
        }
        throw new NoSuchElementException("Element not found!");
```

PR #167, opened 2020-12-08
Task completed

NikolayFatkullin:
- Overridden methods: add(), addAll(), get(), set(), remove(), size(), isEmpty();
- Created method enlargeCapacity() to enlarge our ArrayList;
- Created method findElement() to find element by value, and if it is not found, it throws NoSuchElementException;
- Created method isCorrectIndex() which throw ArrayIndexOutOfBOundException, if index incorrect;

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[CAPACITY];
+        this.size = 0;
+    }
 
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            enlargeCapacity();
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == elementData.length) {
+            enlargeCapacity();
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            this.add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isCorrectIndex(index);
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        isCorrectIndex(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isCorrectIndex(index);
+        T removedElement = (T) elementData[index];
+        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T t) {
-        return null;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
-
+        if (size == elementData.length) {
+            enlargeCapacity();
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 25: let’s also check the index - it should not be less than 0 or greater than size (it can be equal to size though)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        return removedElement;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = findElement(t);
+        if (index == - 1) {
+            return t;
+        }
```
KseniiaMakarova, line 67: I think this can be removed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+
+    private void isCorrectIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index - " + index);
+        }
+    }
+
+    private int findElement(T elementToFind) {
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == elementToFind || elementData[i].equals(elementToFind)) {
```
KseniiaMakarova, line 95: what if `elementData[i]` is null and `elementToFind` is not null?

KseniiaMakarova (changes requested):
overall very good job) 👍

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == elementData.length) {
+            enlargeCapacity();
+        } else if (index > size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index - " + index);
+        }
```
KseniiaMakarova, line 30: why else if? they are 2 independent checks
index check is more reasonable to be put first

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
...
+    }
+
+    private int findElement(T elementToFind) {
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == null && elementToFind != null) {
+                continue;
+            }
+            if (elementData[i] == elementToFind || elementData[i].equals(elementToFind)) {
```
KseniiaMakarova, line 97: let's write everything in one `if` (remember how we compared objects in previous tasks)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,102 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[CAPACITY];
+        this.size = 0;
```
KseniiaMakarova, line 12:
```suggestion
        elementData = new Object[CAPACITY];
        size = 0;
```

KseniiaMakarova (approved):

PR #169, opened 2020-12-08
Array list by Oleh Kravets

Ketch92:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
+    }
 
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            expandArray();
+        }
+        array[size] = value;
+        size += 1;
```
dmytro-hryhoruk, line 23:
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        size += 1;
+        if (array.length <= size) {
+            expandArray();
+        }
+        indexValidation(index);
+        T[] newArray = Arrays.copyOf(array, array.length);
```
dmytro-hryhoruk, line 33: you don't need this copy, you can use the same array in System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
     @Override
-    public T remove(T t) {
-        return null;
+    public T remove(T element) {
+        int index = getIndex(element);
+        if (index == -1) {
+            throw new NoSuchElementException("The "
+                    + element.toString() + " wasn't found in the list!");
+        }
+        return index == -1 ? null : remove(index);
```
dmytro-hryhoruk, line 77: what will happen if index == -1 at 74 line? is there a chance that at this line (77) we can get index = -1?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
+                    + " is out of the range of list size " + size);
+        }
+    }
+
+    private void expandArray() {
+        if (array.length == LIST_MAX_SIZE) {
+            throw new ArrayStoreException("List size reached maximum");
+        }
+        int newLength = Math.min((array.length >> 1) + array.length, LIST_MAX_SIZE);
+        array = Arrays.copyOf(array, newLength);
```
dmytro-hryhoruk, line 111: let's use System.arraycopy() here too

dmytro-hryhoruk (changes requested):
Good job!

src/main/java/core/basesyntax/ArrayList.java
Ketch92: Thanks, nice tip

src/main/java/core/basesyntax/ArrayList.java
Ketch92: Did

src/main/java/core/basesyntax/ArrayList.java
Ketch92: true, this condition is redundant

src/main/java/core/basesyntax/ArrayList.java
Ketch92: ok, but system.arraycopy doesn't create new array, so I have to create new array and copy there. Then reasignt array to new array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_LENGTH = 10;
+    private static final int LIST_MAX_SIZE = Integer.MAX_VALUE - 8;
+    private int size;
+    private T[] array;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_LENGTH];
```
dmytro-hryhoruk, line 12:
```suggestion
        array = (T[]) new Object[DEFAULT_LENGTH];
```

dmytro-hryhoruk (approved):

PR #170, opened 2020-12-08
ArrayList implementation

Artificial-Friend:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_SIZE = 8;
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE;
+    private Object[] array = new Object[ARRAY_SIZE];
+    private int counter = 0;
+
+    private void resize(int minSize) {
+        if (minSize < 0) {
+            throw new OutOfMemoryError();
```
boroda4436, line 13: what?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_SIZE = 8;
```
boroda4436, line 6: why 8?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+    private Object[] array = new Object[ARRAY_SIZE];
+    private int counter = 0;
+
+    private void resize(int minSize) {
+        if (minSize < 0) {
+            throw new OutOfMemoryError();
+        }
+        int newSize = minSize + (minSize >> 2);
+        if (newSize - MAX_ARRAY_SIZE >= 0) {
+            throw new RuntimeException("Array is out of integer bounds and "
```
boroda4436, line 17: In the rest of your implementation, you are using `ArrayIndexOutOfBoundsException`. Why here you are using `RuntimeException`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
     @Override
     public int size() {
-        return 0;
+        return counter;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return counter <= 0;
```
boroda4436, line 99: could counter be less than 0?

boroda4436 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: I tried to implement my ArrayList as far as possible from original implementation. Didn't knew it is mandatory.

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: In case if...

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: It's just my wild imagination.

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: Since it is impossible to have < 0 array size I thought it would be suitable exception, since it's not user's fault but programmer (me) that didn't created implementation to go beyond integer size.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,98 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_SIZE = 10;
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE;
+    private Object[] array = new Object[ARRAY_SIZE];
+    private int counter = 0;
+
+    private void resize(int minSize) {
+        int newSize = minSize + (minSize >> 1);
+        if (newSize - MAX_ARRAY_SIZE >= 0) {
```
boroda4436, line 13: can we have `newSize` greater than `Integer.MAX_VALUE`?

boroda4436 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: I'm blindly stupid, sorry. Now I understand.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_SIZE = 10;
+    private Object[] array;
+    private int counter = 0;
```
KseniiaMakarova, line 8: let's initialise this variable in the constructor as well

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+
+    public ArrayList() {
+        array = new Object[ARRAY_SIZE];
+    }
+
+    private void resize(int minSize) {
```
KseniiaMakarova, line 14: let’s move private methods to the bottom of the file, or at least after the place where we call them

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        }
+        Object[] oldArray = array;
+        array = new Object[newSize];
+        System.arraycopy(oldArray, 0, array,0, counter);
+    }
 
     @Override
     public void add(T value) {
-
+        if (counter >= array.length - 1) {
```
KseniiaMakarova, line 27:
```suggestion
        if (counter == array.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+            resize(counter);
+        }
+        array[counter] = value;
+        counter = counter + 1;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (counter >= array.length - 1) {
```
KseniiaMakarova, line 36: same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+            resize(counter);
+        }
+        array[counter] = value;
+        counter = counter + 1;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 35: let’s also check the index - it should not be less than 0 or greater than size (it can be equal to size though)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        array[index] = value;
+        counter = counter + 1;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list.size() > array.length) {
+            resize(list.size());
+        }
```
KseniiaMakarova, line 48: this is redundant - you already have capacity check in your add() method that you use in a loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= counter) {
+            throw new ArrayIndexOutOfBoundsException("Out of bounds for index: " + index);
+        } else {
```
KseniiaMakarova, line 58: the index check for `get()`, `set()` and `remove()` methods may be moved to the private method
also, what if index < 0?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        System.arraycopy(array, index + 1, array, index, counter - index - 1);
+        counter = counter - 1;
+        return oldValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < counter; i++) {
+            if (array[i] == null || array[i].equals(t)) {
```
KseniiaMakarova, line 82: is this the right way to compare two objects
check your previous solutons

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+                T oldValue = get(i);
+                System.arraycopy(array, i + 1, array, i, counter - i - 1);
+                counter = counter - 1;
+                return oldValue;
```
KseniiaMakarova, line 86: can we use the existing method here?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: Thank you very much for that review!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+        System.arraycopy(array, index + 1, array, index, counter - index - 1);
+        counter = counter - 1;
+        return oldValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < counter; i++) {
+            if (array[i] == null || array[i].equals(t)) {
```
KseniiaMakarova, line 66: let's refactor this line - check how we compared two objects in the previous solutions

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+                T oldValue = get(i);
+                System.arraycopy(array, i + 1, array, i, counter - i - 1);
+                counter = counter - 1;
+                return oldValue;
```
KseniiaMakarova, line 70: now when you have the index, let's call `remove(int index)` method here and return its result

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+                throw new ArrayIndexOutOfBoundsException("Array is out of integer bounds and "
+                    + "require hugeCapacity() implementation. It's can't be bigger right now!");
+            }
+            Object[] oldArray = array;
+            array = new Object[newSize];
+            System.arraycopy(oldArray, 0, array, 0, counter);
+        }
+    }
+
+    private void indexValidator(int index) {
```
KseniiaMakarova, line 99:
```suggestion
    private void validateIndex(int index) {
```
let's use verbs for method names

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Artificial-Friend: Thank you, done!

KseniiaMakarova (approved):

PR #171, opened 2020-12-08
created ArrayList, implemented all methods that is in "List" interfac…

YuraBondarenko:
…e and own methods such as "listIncrease", "arrayIndexOutOfBoundException" and "isListFull"

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_LIST_SIZE = 10;
+    private T[] list;
+    private int size;
+
+    public ArrayList() {
+        this.list = (T[]) new Object[DEFAULT_LIST_SIZE];
+        this.size = 0;
+    }
+
+    private void listIncrease() {
+        T[] newList = (T[]) new Object[(int) (list.length * 1.5)];
+        System.arraycopy(list, 0, newList, 0, size);
+        list = newList;
+    }
+
+    private void arrayIndexOutOfBoundException(int index) {
+        if (index >= size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
+    }
+
+    private void isListFull() {
+        if (size == list.length) {
+            listIncrease();
+        }
+    }
 
     @Override
     public void add(T value) {
-
+        isListFull();
+        list[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == list.length) {
+            isListFull();
+        }
+        System.arraycopy(list, index, list, index + 1, size - index);
+        list[index] = value;
+        size++;
+        arrayIndexOutOfBoundException(index);
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        arrayIndexOutOfBoundException(index);
+        return list[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        arrayIndexOutOfBoundException(index);
+        if (index < size) {
+            list[index] = value;
+        } else {
+            add(value);
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        arrayIndexOutOfBoundException(index);
+        T remove = get(index);
+        System.arraycopy(list, index + 1, list, index, size - index);
+        size--;
+        return remove;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
+        if (t == null) {
+            size--;
+            return null;
+        }
+        for (int i = 0; i < size; i++) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
+        this.size = 0;
+    }
+
+    private void listIncrease() {
+        T[] newList = (T[]) new Object[(int) (list.length * 1.5)];
+        System.arraycopy(list, 0, newList, 0, size);
+        list = newList;
+    }
+
+    private void arrayIndexOutOfBoundException(int index) {
```
dmytro-hryhoruk, line 21:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
+        if (index >= size) {
```
dmytro-hryhoruk, line 22: what if less than zero?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
+        isListFull();
+        list[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == list.length) {
+            isListFull();
```
dmytro-hryhoruk, line 43: same check twice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
+        }
+        System.arraycopy(list, index, list, index + 1, size - index);
+        list[index] = value;
+        size++;
+        arrayIndexOutOfBoundException(index);
```
dmytro-hryhoruk, line 48: do we need it here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        arrayIndexOutOfBoundException(index);
+        if (index < size) {
+            list[index] = value;
+        } else {
+            add(value);
+        }
```
dmytro-hryhoruk, line 71: set should not add a new element. Either replace the value of the old one or throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        arrayIndexOutOfBoundException(index);
+        T remove = get(index);
```
dmytro-hryhoruk, line 77: you have access to the array. If the index is valid you can just get element by it;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,110 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        int index = -1;
+        if (t == null) {
+            size--;
+            return null;
+        }
+        for (int i = 0; i < size; i++) {
+            if ((t.equals(list[i]) || t == list[i])) {
```
dmytro-hryhoruk, line 91: let's combine two if conditions and decrease size there

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
YuraBondarenko: I need to save element in position "index" before call `System.arraycopy(list,` index + 1, list, index, size - index);` otherwise I can't return this element because it will deleted

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
-        return null;
+        checkIndex(index);
+        return list[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        if (index < size) {
```
dmytro-hryhoruk, line 64: checkIndex isn't enough?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
+            list[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedElement = get(index);
```
dmytro-hryhoruk, line 72:
```suggestion
        T removedElement = list[index];
```
You have access to the array and you have checked the index already

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (t == null) {
+                size--;
+                return null;
+            } else if (t.equals(list[i])) {
+                return remove(i);
+            }
```
dmytro-hryhoruk, line 86: let's combine these too conditions

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
YuraBondarenko: ok, i understood what did you mean

dmytro-hryhoruk (approved):

PR #177, opened 2020-12-09
added solution

Fedich09:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
     public void add(T value) {
-
+        elements[size] = value;
+        size++;
+        if (elements.length == size) {
+            T[] newElements = (T[]) new Object[(int) (elements.length * 1.5)];
+            for (int i = 0; i < elements.length; i++) {
+                newElements[i] = elements[i];
+            }
+            elements = newElements;
```
KseniiaMakarova, line 23: you moved this logic to the private method - let's use it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
     @Override
     public void add(T value) {
-
+        elements[size] = value;
+        size++;
+        if (elements.length == size) {
+            T[] newElements = (T[]) new Object[(int) (elements.length * 1.5)];
+            for (int i = 0; i < elements.length; i++) {
+                newElements[i] = elements[i];
+            }
```
KseniiaMakarova, line 22: use System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index >= 0 && index <= size) {
+            if (elements.length == size) {
+                grow();
+            }
+            for (int i = size; i > index; i--) {
+                elements[i] = elements[i - 1];
+            }
```
KseniiaMakarova, line 35: use System.arraycopy() please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= 0 && index < size) {
+            return elements[index];
+        } else {
```
KseniiaMakarova, line 54: do we need else after return?

also, the index check for `get()`, `set()` and `remove()` methods may be moved to the private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+            throw new ArrayIndexOutOfBoundsException("Wrong index!");
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
+            T deletedElement = elements[index];
+            if (index >= 0 && index < size) {
```
KseniiaMakarova, line 72: do we need the second if?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+                for (int i = index; i < size - 1; i++) {
+                    elements[i] = elements[i + 1];
+                }
```
KseniiaMakarova, line 75: use System.arraycopy() please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+            return deletedElement;
+        } else {
+            throw new ArrayIndexOutOfBoundsException();
+        }
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        int deleteIndex = -1;
```
KseniiaMakarova, line 86: we don't need this variable - we can just call `return` on line 93

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+        T returnElement = (T) new Object();
+        for (int i = 0; i < elements.length; i++) {
```
KseniiaMakarova, line 88: should we iterate over the filled part only?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+            if ((t == null && elements[i] == null)
+                    || ((t != null && elements[i] != null) && t.equals(elements[i]))) {
```
```suggestion
            if (t == elements[i]
                    || t != null && t.equals(elements[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        if (size > 0) {
+            return false;
+        } else {
+            return true;
+        }
```
KseniiaMakarova, line 115: this can be simplified

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,125 @@
...
+    }
+
+    public void grow() {
```
KseniiaMakarova, line 118: make this method private

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Fedich09: yes, we should, because I should will find find element which must equal element that I should delete

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_CAPACITY = 10;
+    private T[] elements;
+    private int size = 0;
```
KseniiaMakarova, line 8: let's initialise this field in the constructor as well

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
...
+        if (index >= 0 && index < size) {
+            return elements[index];
+        }
+        throw new ArrayIndexOutOfBoundsException("Wrong index!");
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= 0 && index < size) {
```
KseniiaMakarova, line 54: let's move the logic (if index for `get()`, `set()` and `remove()` methods is invalid) -> throw the exception
to the private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
...
+            return deletedElement;
+        }
+        throw new ArrayIndexOutOfBoundsException();
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        T returnElement = (T) new Object();
+        for (int i = 0; i < elements.length; i++) {
```
KseniiaMakarova, line 75: if we should iterate over the filled part only, let's use our `size` instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
...
+            if (t == elements[i]
+                    || ((t != null && elements[i] != null) && t.equals(elements[i]))) {
+                returnElement = elements[i];
+                remove(i);
+                return returnElement;
```
KseniiaMakarova, line 80:
```suggestion
                return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        if (size > 0) {
+            return false;
+        }
+        return true;
```
KseniiaMakarova, line 96: let's simplify this please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,106 @@
...
+    }
+
+    private void grow() {
+        T[] newElements = (T[]) new Object[(int) (elements.length * 1.5)];
+        for (int i = 0; i < size; i++) {
+            newElements[i] = elements[i];
+        }
```
KseniiaMakarova, line 103: use System.arraycopy

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
...
+    private int size = 0;
+
+    public ArrayList() {
+        this.elements = (T[]) new Object[ARRAY_CAPACITY];
+    }
+
+    public ArrayList(T[] elements, int size) {
+        this.elements = elements;
+        this.size = size;
+    }
```
KseniiaMakarova, line 17: we don't need this constructor - let's remove it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
...
+        return size <= 0;
+    }
+
+    private void grow() {
+        T[] newElements = (T[]) new Object[(int) (elements.length * 1.5)];
+        System.arraycopy(elements, 0, newElements, 0, size);
+        elements = newElements;
+    }
+
+    private void exception(int index) {
```
KseniiaMakarova, line 98: please make the method name more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int ARRAY_CAPACITY = 10;
+    private T[] elements;
+    private int size = 0;
```
KseniiaMakarova, line 8: assign the value to this field in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,103 @@
...
+        size--;
+        return deletedElement;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (t == elements[i]
+                    || ((t != null && elements[i] != null) && t.equals(elements[i]))) {
```
KseniiaMakarova, line 75:
```suggestion
                    || t != null && t.equals(elements[i])) {
```
this may be simplified

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+        wrongIndexException(index);
+        return elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        wrongIndexException(index);
+        elements[index] = value;
+        return;
```
KseniiaMakarova, line 55: do you need this return?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+        return size <= 0;
+    }
+
+    private void grow() {
+        T[] newElements = (T[]) new Object[(int) (elements.length * 1.5)];
+        System.arraycopy(elements, 0, newElements, 0, size);
+        elements = newElements;
+    }
+
+    private void wrongIndexException(int index) {
```
KseniiaMakarova, line 94: method names should be verbs
let's name it something like `checkIndex`, `validateIndex`

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #178, opened 2020-12-09
yegor sereda first commit

entertainment-veks:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
 
+    public static final int DEFAULT_CAPACITY = 10;
+
+    private T[] values = (T[]) new Object[DEFAULT_CAPACITY];
+    private int size = 0;
```
KseniiaMakarova, line 10: let’s initialise non-static fields in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean isHaveSpace() {
```
KseniiaMakarova, line 88:
```suggestion
    private boolean isSpaceAvailable() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
-
+        if (!isHaveSpace()) {
+            grow();
+        }
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
```
KseniiaMakarova, line 22: let’s also check the index - it should not be less than 0 or greater than size (it can be equal to size though)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < size) {
```
KseniiaMakarova, line 41: what if index < 0?
the index check for `get()`, `set()` and `remove()` methods may be moved to the private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T oldValue = values[index];
+        if (size == 0) {
+            return oldValue;
+        }
```
KseniiaMakarova, line 61: do we need this check? let's check if the index is valid instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        size--;
+        return oldValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < values.length; i++) {
```
KseniiaMakarova, line 69: should we iterate over the filled part only?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,118 @@
...
+            throw new ArrayIndexOutOfBoundsException();
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isIndexInvalid(index)) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
```
KseniiaMakarova, line 77: index also can't be equal or greater to size, like in `get()` and `set()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,118 @@
...
+
+        if (index < size) {
+            return values[index];
+        }
+        throw new ArrayIndexOutOfBoundsException();
     }
 
     @Override
     public void set(T value, int index) {
+        if (isIndexInvalid(index)) {
```
KseniiaMakarova, line 62: let's add another condition here as well

so that we could simplify this method and use only one `if` and one `throw`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,118 @@
...
     public T get(int index) {
-        return null;
+        if (isIndexInvalid(index)) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+
+        if (index < size) {
+            return values[index];
+        }
+        throw new ArrayIndexOutOfBoundsException();
```
KseniiaMakarova, line 57: same here

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #180, opened 2020-12-09
added realised methods for my array list

ShemsheyMaxim:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private static final double COEFFICIENT = 0.5;
+    private Object[] array = new Object[CAPACITY];
+    private int size = 0;
```
KseniiaMakarova, line 9: let's initialise non-static fields in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
 
     @Override
     public void add(T value) {
-
+        if (size < array.length) {
```
KseniiaMakarova, line 13: let's move resizing logic in a separate private method

if (size is equal to array length) -> resize
after we resized or not, we add the value

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
+            array[size] = value;
+            size++;
+            return;
+        }
+        Object[] oldArray = new Object[array.length];
+        System.arraycopy(array, 0, oldArray, 0, array.length);
+        array = new Object[(int) (array.length * COEFFICIENT) + array.length];
+        System.arraycopy(oldArray, 0, array, 0, oldArray.length);
```
KseniiaMakarova, line 21: try creating the new larger array and use one System.arraycopy()
then make `array` field point on the new array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size) {
```
KseniiaMakarova, line 28: what if index < 0?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
+            throw new ArrayIndexOutOfBoundsException("We can't add value on this position because "
+                    + "index > size" + index + " > " + size);
+        }
+        System.arraycopy(array, index, array, index + 1, size - index);
```
KseniiaMakarova, line 32: should we also check the size here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size) {
```
KseniiaMakarova, line 46: same question about negative index (< 0) goes for all your index checks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        if (index < size) {
+            array[index] = value;
+            return;
+        }
+        throw new ArrayIndexOutOfBoundsException("We can't add value on this position because "
+                + "index > size" + index + " > " + size);
```
KseniiaMakarova, line 59:
```suggestion
        throw new ArrayIndexOutOfBoundsException("The index " + index + " is invalid");
```
the index check for `get()`, `set()` and `remove()` methods may be moved to the private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size) {
+            throw new ArrayIndexOutOfBoundsException("We can't add value on this position because "
+                    + "index > size" + index + " > " + size);
+        }
+        T result = (T) array[index];
+        System.arraycopy(array, index + 1, array, index, size - index);
```
KseniiaMakarova, line 69:
```suggestion
        System.arraycopy(array, index + 1, array, index, ???);
```
the last part is not 100% correct, please reconsider it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
+        size--;
+        return result;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
```
KseniiaMakarova, line 76: should we iterate over the filled part only?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,93 @@
...
+            if (array[i] == t || (array[i] != null && array[i].equals(t))) {
+                return (T) remove(i);
+            }
+        }
+        throw new NoSuchElementException("Value don't found");
```
KseniiaMakarova, line 81:
```suggestion
        throw new NoSuchElementException("Value not found");
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private static final double COEFFICIENT = 0.5;
+    private T[] array;
+    private int size;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[CAPACITY];
+        this.size = 0;
```
KseniiaMakarova, line 13:
```suggestion
        array = (T[]) new Object[CAPACITY];
        size = 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+    private void indexCheckForGet(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("The index " + index + " is invalid");
+        }
+    }
+
+    private void indexCheckForSet(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException("The index " + index + " is invalid");
+        }
```
KseniiaMakarova, line 105: are these methods identical?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        return size == 0;
+    }
+
+    private T[] resize(T[] array) {
+        T[] newArray = (T[]) new Object[(int) (array.length * COEFFICIENT) + array.length];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
+    }
+
+    private void indexCheckForAddAndRemove(int index) {
```
```suggestion
    private void indexCheckForAdd(int index) {
```
remove() method does not allow index == size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexCheckForAddAndRemove(index);
+        if (size + 1 == array.length) {
```
KseniiaMakarova, line 28:
```suggestion
        if (size == array.length) {
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
ShemsheyMaxim: done

src/main/java/core/basesyntax/ArrayList.java
ShemsheyMaxim: done

src/main/java/core/basesyntax/ArrayList.java
ShemsheyMaxim: done

src/main/java/core/basesyntax/ArrayList.java
ShemsheyMaxim: done

KseniiaMakarova (approved):

PR #182, opened 2020-12-09
created own ArrayList

Ilya-brinchuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
```
dmytro-hryhoruk, line 8:
```suggestion
    private T[] elements;
```
Let's not call the inner array of ArrayList as arrayList)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+    private int size;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private T[] grow() {
+        return Arrays.copyOf(arrayList, arrayList.length + arrayList.length / 2);
```
dmytro-hryhoruk, line 17: Let's make this method void(it's private so it's okay) and reassign our array inside

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isIndexValid(index + 1);
```
dmytro-hryhoruk, line 58: let's make two checks: One for methods where index should be in range : index >= 0 && index < size(get, set, remove);
And the second one: index >= 0 && index <= size
 which you will use only in add method(le'ts not make a method for this one)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        return removeValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((t == arrayList[i]) || (arrayList[i] != null && arrayList[i].equals(t))) {
+                T val = remove(i);
+                return val;
```
dmytro-hryhoruk, line 82:
```suggestion
                return remove(i);
```

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+            return true;
+        }
+        throw new ArrayIndexOutOfBoundsException("Wrong index. The index must be between"
+                + " 0 and size: " + size);
+    }
 
     @Override
     public void add(T value) {
-
+        if (size == elements.length - 1) {
```
dmytro-hryhoruk, line 30: do you need to compare with elements.length - 1 or with just  elements.length?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 ^ index > size) {
+            throw new ArrayIndexOutOfBoundsException("Wrong index. The index must be between"
+                    + " 0 and size: " + size);
+        }
+        if (size == elements.length - 1) {
```
dmytro-hryhoruk, line 43: same

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Ilya-brinchuk: I understood. Under this condition, the last element of the array is never used. Thank you)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        size--;
+        return removeValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((t == elements[i]) || (elements[i] != null && elements[i].equals(t))) {
```
dmytro-hryhoruk, line 83:
```suggestion
            if (t == elements[i] || elements[i] != null && elements[i].equals(t)) {
```

dmytro-hryhoruk (approved):

PR #184, opened 2020-12-09
Realized Array List

AndreiYazadzhi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_SIZE = 10;
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        this.array = new Object[INITIAL_SIZE];
+        this.size = 0;
```
KseniiaMakarova, line 12: do we need to use `this`? consider all your methods, not only this one

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+
+    private void grow() {
+        Object[] resizedArray = new Object[(array.length * 3) / 2];
+        System.arraycopy(array, 0, resizedArray, 0, array.length);
+        this.array = resizedArray;
+    }
+
+    private boolean isIndexInBound(int index) {
+        return index < size && index >= 0;
+    }
```
KseniiaMakarova, line 34: let's move private methods to the bottom of the file, or at least after the place where we call them

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
     @Override
     public void add(T value) {
-
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == array.length - 1) {
```
KseniiaMakarova, line 43:
```suggestion
        if (size == array.length) {
```
there still will be one empty slot if size == array.length - 1

KseniiaMakarova, line 43: let's also check the index - it should not be less than 0 or greater than size (it can be equal to size though)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (!isIndexInBound(index)) {
+            throw new ArrayIndexOutOfBoundsException();
+        } else {
```
KseniiaMakarova, line 62: do we need else after throw?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+            throw new ArrayIndexOutOfBoundsException();
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (!isIndexInBound(index)) {
+            throw new ArrayIndexOutOfBoundsException();
+        } else {
```
KseniiaMakarova, line 80: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+            System.arraycopy(array, index + 1, array, index, size - index - 1);
+            size--;
+            return deletedElement;
+        }
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0;i < size;i++) {
```
```suggestion
        for (int i = 0; i < size; i++) {
```

KseniiaMakarova (changes requested):
overall very nice!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        if (size < 0) {
+            throw new RuntimeException("Provide size less than 0");
+        } else if (size == 0) {
+            array = new Object[INITIAL_SIZE];
+        } else {
+            array = new Object[size];
+        }
+    }
+
+    private void grow() {
```
KseniiaMakarova, line 25: move this method to the bottom of the file

KseniiaMakarova (approved):

PR #188, opened 2020-12-09
added solution

ivlytvynets:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (!isIndexValid(index) && index != size) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+        if (size == elementData.length) {
+            elementData = grow();
+        }
```
dmytro-hryhoruk, line 32: let's move this check to addElementInMiddle()(in add() method you are doing this anyway;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+        T removedElement = get(index);
+        arrayShift(index);
+        return removedElement;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        if (!isIndexValid(getIndex(t))) {
```
dmytro-hryhoruk, line 78: let's save the result of getIndex() to local variable in order not to invocate it twice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int newCapacity(int oldCapacity) {
+        return (int) (oldCapacity * 1.5);
+    }
```
dmytro-hryhoruk, line 97: don't see the point of making it a separate method and use only in onve place

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+
+    private T[] grow() {
```
dmytro-hryhoruk, line 99: let's make this method void and reassign the array here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+        return Arrays.copyOf(elementData, newCapacity(size));
+    }
+
+    private boolean isIndexValid(int index) {
+        return index >= 0 && index < size;
+    }
+
+    private int getIndex(T value) {
+        int index = 0;
+        for (T element: elementData) {
```
dmytro-hryhoruk, line 109: Optional: you may use for i loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+            }
+            index++;
+        }
+        return -1;
+    }
+
+    private void addElementInMiddle(T value, int index) {
+        for (int i = size; i > index; i--) {
+            elementData[i] = elementData[i - 1];
+        }
```
dmytro-hryhoruk, line 121: use System.arraycopy for that

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+        elementData[index] = value;
+        size++;
+    }
+
+    private void arrayShift(int index) {
+        for (int i = 0; i + index + 1 < size; i++) {
+            elementData[index + i] = elementData [index + i + 1];
+        }
```
dmytro-hryhoruk, line 129: same

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
ivlytvynets: done

ivlytvynets (commented):
all is fixed

dmytro-hryhoruk (approved):

PR #190, opened 2020-12-10
Implement my own ArrayList

Resci:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+        array[index + 1] = array[index];
+    }
+
+    private void moveLeft(int index) {
+        array[index - 1] = array[index];
+    }
+
+    private T[] grow() {
+        return array = Arrays.copyOf(array, (int) (array.length * 1.5));
     }
```
dmytro-hryhoruk, line 115:
```suggestion
    private void grow() {
         array = Arrays.copyOf(array, (int) (array.length * 1.5));
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_SIZE];
+    }
 
     @Override
     public void add(T value) {
-
+        if (array.length == size) {
+            array = grow();
```
dmytro-hryhoruk, line 18:
```suggestion
         grow();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+        }
+        array[size] = value;
+        size++;
```
dmytro-hryhoruk, line 21:
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+            add(value);
+        } else if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Add element failed. Index out of bounds");
+        } else {
+            if (size == array.length) {
+                array = grow();
+            }
+            for (int i = size - 1; i >= index; i--) {
+                moveRight(i);
+            }
```
dmytro-hryhoruk, line 36: use System.arraycopy() method. Same in remove

dmytro-hryhoruk: Let's make checks in this order:
check index range(throw exception;
if index = size use add();
add by index  with shift of all elements;
Dont use else block after throw keyword. It's redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+        if (index < size && index >= 0) {
+            return array[index];
+        }
+        throw new ArrayIndexOutOfBoundsException("Getting element failed. Index out of bounds");
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < size && index >= 0) {
```
dmytro-hryhoruk, line 61: let's make a boolean method(checkIndex) and use it in set(), remove(), get()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,116 @@
...
+        }
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == t || (array[i] != null && array[i].equals(t))) {
+                remove(i);
+                return t;
```
dmytro-hryhoruk, line 89:
```suggestion
                return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
dmytro-hryhoruk: And let's use this order everywhere:
if(condition) {throw new Exception}
//rest part of the method

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

boroda4436 (approved):

PR #191, opened 2020-12-10
done

AleksandrPok:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+            grow();
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (array.length == size) {
```
KseniiaMakarova, line 26: let’s also check the index - it should not be less than 0 or greater than size (it can be equal to size though)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (array.length < size + list.size()) {
+            grow();
```
KseniiaMakarova, line 37: there still will be capacity checks in add() method that you later use in for loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+        }
+        addition(list);
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < size) {
```
KseniiaMakarova, line 44: what if index < 0?
also, the index check for `get()`, `set()` and `remove()` methods may be moved to the private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+            array[index] = value;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index");
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T tmp = array[index];
```
KseniiaMakarova, line 62: reconsider the variable name

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (t == null) {
+                size--;
+                return null;
+            }
+            if (array[i].equals(t)) {
+                return remove(i);
+            }
```
KseniiaMakarova, line 77: we have already compared two objects considering nulls multiple times in previous tasks
let's recall this)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        T[] newArray = (T[]) new Object[newLength];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        array = newArray;
+    }
+
+    private void addition(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
+    }
```
KseniiaMakarova, line 94: do you need the separate method? let's move this logic to addAll()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,101 @@
...
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
+        size--;
+        return returnedValue;
     }
 
     @Override
     public T remove(T t) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == null || array[i].equals(t)) {
```
KseniiaMakarova, line 66: this is not 100% right

please remember how we compared objects in the previous tasks (for example, keys in storage)

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #230, opened 2021-04-08
hw-array-list

Trident1998:
There was created ArrayList class and override the methods!

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,128 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double INCREASING_COEFFICIENT = 10;
+    private static final int IMPORTANT_VALUE = 1;
+
+    private T[] elementData = (T[]) new Object[DEFAULT_SIZE];
+    private int size = 0;
+
+    private T[] grow(int oldSize) {
+        return (T[]) new Object[(int) (oldSize * INCREASING_COEFFICIENT)];
+    }
+
+    private T[] copiedList(T[] oldList, T[] newList) {
+        for (int i = 0; i < oldList.length; i++) {
+            newList[i] = oldList[i];
+        }
+        return newList;
+    }
+
+    private void addListInList(List<T> sourceList, T[] destList,
+                              int sizeOfSourceList, int sizeOfDestList) {
+        for (int i = sizeOfDestList, j = 0; j < sizeOfSourceList; j++, i++) {
+            destList[i] = sourceList.get(j);
+        }
+    }
+
+    private void droppedValue(T[] elementData, int index) {
+        int numMoved = size - index - IMPORTANT_VALUE;
+        System.arraycopy(elementData, index + IMPORTANT_VALUE, elementData, index, numMoved);
+        elementData[size - IMPORTANT_VALUE] = null;
+        size -= IMPORTANT_VALUE;
+    }
 
     @Override
     public void add(T value) {
-
+        if (size < elementData.length) {
+            elementData[size] = value;
+        } else {
+            elementData = copiedList(elementData, grow(elementData.length));
+            elementData[size] = value;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index >= size + IMPORTANT_VALUE || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        if (DEFAULT_SIZE > elementData.length || size < elementData.length) {
+            System.arraycopy(elementData, index, elementData,
+                    index + IMPORTANT_VALUE, size - index);
+            elementData[index] = value;
+        } else {
+            T[] widerList = copiedList(elementData, grow(elementData.length));
+            System.arraycopy(elementData, index, widerList, index + IMPORTANT_VALUE, size - index);
+            widerList[index] = value;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int sizeOfSourceList = list.size();
+        if (size + list.size() < elementData.length) {
+            addListInList(list, elementData, sizeOfSourceList, size);
+        } else {
+            addListInList(list, grow(elementData.length), sizeOfSourceList, size);
+        }
+        size += sizeOfSourceList;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        T oldValue = elementData[index];
+        droppedValue(elementData, index);
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = 0;
+        for (int i = 0; i < size; i++) {
+            if (element == elementData[i] || element.equals(elementData[i])) {
+                index = i;
+                break;
+            } else if (size - IMPORTANT_VALUE == i && index == 0) {
+                throw new NoSuchElementException("There is no such element present");
+            }
+        }
+        T oldValue = elementData[index];
+        droppedValue(elementData, index);
+        return oldValue;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double INCREASING_COEFFICIENT = 10;
+    private static final int IMPORTANT_VALUE = 1;
+
+    private T[] elementData = (T[]) new Object[DEFAULT_SIZE];
+    private int size = 0;
```
Max-Pochepets, line 11:
```suggestion
    private T[] elementData = (T[]) new Object[DEFAULT_SIZE];
    private int size;
```
non-constant fields should be set in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+
+    private T[] grow(int oldSize) {
```
Max-Pochepets, line 13: private methods should be placed after public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+        return (T[]) new Object[(int) (oldSize * INCREASING_COEFFICIENT)];
```
Max-Pochepets, line 14: no need for a separate method only for `new ...`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+    }
+
+    private T[] copiedList(T[] oldList, T[] newList) {
```
Max-Pochepets, line 17: you can just use `System.arraycopy`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double INCREASING_COEFFICIENT = 10;
+    private static final int IMPORTANT_VALUE = 1;
```
Max-Pochepets, line 8: no need to make it consant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+        elementData[size - IMPORTANT_VALUE] = null;
+        size -= IMPORTANT_VALUE;
+    }
 
     @Override
     public void add(T value) {
-
+        if (size < elementData.length) {
+            elementData[size] = value;
+        } else {
```
Max-Pochepets, line 42: can we manage without else?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+            elementData = copiedList(elementData, grow(elementData.length));
+            elementData[size] = value;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index >= size + IMPORTANT_VALUE || index < 0) {
```
Max-Pochepets, line 51:
```suggestion
        if (index > size || index < 0) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        if (DEFAULT_SIZE > elementData.length || size < elementData.length) {
+            System.arraycopy(elementData, index, elementData,
+                    index + IMPORTANT_VALUE, size - index);
+            elementData[index] = value;
+        } else {
+            T[] widerList = copiedList(elementData, grow(elementData.length));
+            System.arraycopy(elementData, index, widerList, index + IMPORTANT_VALUE, size - index);
```
Max-Pochepets, line 60: can we move this part to method grow entirely?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        int sizeOfSourceList = list.size();
+        if (size + list.size() < elementData.length) {
+            addListInList(list, elementData, sizeOfSourceList, size);
+        } else {
+            addListInList(list, grow(elementData.length), sizeOfSourceList, size);
+        }
+        size += sizeOfSourceList;
```
Max-Pochepets, line 74: can we just use `add(T)` here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size || index < 0) {
```
Max-Pochepets, line 95:
```suggestion
        if (index >= size || index < 0) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,128 @@
...
+            if (element == elementData[i] || element.equals(elementData[i])) {
+                index = i;
+                break;
+            } else if (size - IMPORTANT_VALUE == i && index == 0) {
+                throw new NoSuchElementException("There is no such element present");
+            }
+        }
+        T oldValue = elementData[index];
+        droppedValue(elementData, index);
+        return oldValue;
```
Max-Pochepets, line 116: there is a code duplication

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: Yes we can!! I threw "else" out.

src/main/java/core/basesyntax/ArrayList.java
Trident1998: Yes, we can!! I did that!

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: I don't think so. Because I have already realized other type of System.arraycopya in grow method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double INCREASING_COEFFICIENT = 1.5;
+
+    private T[] elementData = (T[]) new Object[DEFAULT_SIZE];
```
KseniiaMakarova, line 9: let's initialise `elementData` field in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        if (DEFAULT_SIZE > elementData.length || size < elementData.length) {
+            System.arraycopy(elementData, index, elementData,
+                    index + 1, size - index);
+            elementData[index] = value;
+        } else {
+            T[] widerList = grow(elementData.length);
+            System.arraycopy(elementData, index, widerList, index + 1, size - index);
+            widerList[index] = value;
+        }
+        size++;
```
KseniiaMakarova, line 35: can we simplify this?
you can do the same check for as in `add(T value)` and see if the array needs to `grow()`
after this, let's just call `System.arraycopy` once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
```
KseniiaMakarova, line 49: this check appears in 3 methods - should we move the code into a private method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size || index < 0) {
```
KseniiaMakarova, line 63:
```suggestion
        if (index >= size || index < 0) {
```
can you remove the element on index 5 if the size is 5?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        return newElementData;
+    }
+
+    private T droppedValue(T[] elementData, int index) {
+        final T oldValue = elementData[index];
+        int numMoved = size - index - 1;
+        System.arraycopy(elementData, index + 1, elementData, index, numMoved);
+        elementData[size - 1] = null;
+        size -= 1;
+        return oldValue;
```
KseniiaMakarova, line 105: good job! here is how we can reduce the number of code lines a bit
```suggestion
        T oldValue = elementData[index];
        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
        elementData[--size] = null;
        return oldValue;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        return size == 0;
+    }
+
+    private T[] grow(int oldSize) {
+        T[] newElementData = (T[]) new Object[(int) (oldSize * INCREASING_COEFFICIENT)];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        return newElementData;
+    }
+
+    private T droppedValue(T[] elementData, int index) {
```
KseniiaMakarova, line 99: let's use verbs for method names
also, do you need to pass `elementData`?
```suggestion
    private T removeValue(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        return droppedValue(elementData, index);
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = 0;
```
KseniiaMakarova, line 71: let's remove this variable - we already have `i` counter in a loop
if we find a match - we can call a private method with `i` as index from the loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+        for (int i = 0; i < size; i++) {
+            if (element == elementData[i] || element.equals(elementData[i])) {
```
KseniiaMakarova, line 73: what if `element` is null and `elementData[i]` is not null?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,107 @@
...
+                index = i;
+                break;
+            } else if (size - 1 == i && index == 0) {
+                throw new NoSuchElementException("There is no such element present");
```
KseniiaMakarova, line 77: we may throw NoSuchElementException if we iterated over the array and did not find a match

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: Hahahah! I'm sorry! No we can't do it

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: if elment == nul and elementData[i] != null, then in first part of condition will "false". This means that in this case, this loop iteration will be skipped. And this is exactly what we need! Or am I wrong?

src/main/java/core/basesyntax/ArrayList.java
Trident1998: done

src/main/java/core/basesyntax/ArrayList.java
Trident1998: if elment == nul and elementData[i] != null, then in first part of condition will "false". This means that in this case, this loop iteration will be skipped. And this is exactly what we need! Or am I wrong?

KseniiaMakarova (approved):

PR #229, opened 2021-04-08
added solution

Vlad-Andr:

PR #232, opened 2021-04-09
added solution/ second attempt

Vlad-Andr:


Vlad-Andr:
https://github.com/mate-academy/jv-array-list/pull/229 - previous PR

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound");
+        }
 
+        int arrayLength = arrayOfObjects.length;
+        Object[] tempo = new Object[arrayLength + 1];
+        System.arraycopy(arrayOfObjects, 0, tempo, 0, arrayLength);
+        arrayOfObjects = null;
+        arrayOfObjects = new Object[arrayLength + 1];
```
Max-Pochepets, line 34: why do you explicitly set array to null?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound");
+        }
 
+        int arrayLength = arrayOfObjects.length;
+        Object[] tempo = new Object[arrayLength + 1];
+        System.arraycopy(arrayOfObjects, 0, tempo, 0, arrayLength);
```
Max-Pochepets, line 32: `System.arraycopy` creates a buffer array inside, no need to create it here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
+        System.arraycopy(tempo, index, arrayOfObjects,
+                index + 1, size - index);
+
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        Object[] listArray = toArray(list);
```
Max-Pochepets, line 45: don't make an array, list has a method to get an element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,132 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Object objectByIndex;
+        if (index >= 0 && index < size) {
+            objectByIndex = arrayOfObjects[index];
+            Object[] tempArrayFirstPart = new Object[index];
+            Object[] tempArraySecondPart = new Object[size - index - 1];
```
Max-Pochepets, line 75: System.arraycopy handles all this, read about how this method works

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < size && index >= 0) {
```
KseniiaMakarova, line 51: please see the checklist - If the logic of your code repeats - move it to the separate private method. For example, the index check for get(), set() and remove() methods should be moved to the separate method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
+            objectByIndex = arrayOfObjects[index];
+            Object[] temp = new Object[size];
+            System.arraycopy(arrayOfObjects, 0, temp, 0, size);
+            arrayOfObjects = new Object[size - 1];
+            System.arraycopy(temp, 0, arrayOfObjects,
+                    0, index);
+            System.arraycopy(temp, index + 1, arrayOfObjects,
+                    index, size - index - 1);
+            size--;
+            return (T) objectByIndex;
```
KseniiaMakarova, line 80: let's remove the `temp` variable here and use `System.arraycopy` only once
we can move part of the elements inside the same `arrayOfObjects` array, and we don't need to decrease its size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growCapacity() {
+        Object[] tempArray = new Object[size + size / 2];
+        System.arraycopy(arrayOfObjects,0,tempArray,0,arrayOfObjects.length);
+        arrayOfObjects = new Object[tempArray.length];
+        System.arraycopy(tempArray,0,arrayOfObjects,0,tempArray.length);
```
KseniiaMakarova, line 111: can we replace these 2 lines with a simpler solution? let's call `System.arraycopy` only once in this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,113 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound");
+        }
+
+        int arrayLength = arrayOfObjects.length;
+        Object[] tempo = new Object[arrayLength + 1];
```
KseniiaMakarova, line 31: same comment as for remove() method - let's remove the `tempo` variable here and use `System.arraycopy` only once
we can move part of the elements inside the same arrayOfObjects array, and we don't need to increase its size - instead we might do the same check as for `add(T value)` method to be sure that the new element will fit

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
+        }
+        checkSize();
+        if (index != size) {
+            System.arraycopy(arrayOfObjects, index, arrayOfObjects, index + 1, size);
+            arrayOfObjects[index] = value;
+        } else {
+            arrayOfObjects[index] = value;
+        }
 
+        size++;
```
KseniiaMakarova, line 34:
```suggestion
        System.arraycopy(arrayOfObjects, index, arrayOfObjects, index + 1, // why `size` here?);
        arrayOfObjects[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,104 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growCapacity() {
+        Object[] tempArray = new Object[size + size / 2];
+        System.arraycopy(arrayOfObjects,0,tempArray,0,arrayOfObjects.length);
```
KseniiaMakarova, line 89: add whitespaces

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,99 @@
...
+        Object objectByIndex = arrayOfObjects[index];
+        System.arraycopy(arrayOfObjects, index + 1, arrayOfObjects, index, size - index);
+        size--;
+        return (T) objectByIndex;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < arrayOfObjects.length; i++) {
```
KseniiaMakarova, line 62: do we need to iterate over the whole array, or over only the filled part?

KseniiaMakarova (approved):

PR #250, opened 2021-04-11
Overrided all methods in ArrayList class + created new methods and fi…

Kharlanava:
…elds.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,96 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity;
```
Resci, line 7:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+    private Object[] elementData;
```
Resci, line 8: Let's use generic type for this variable

Resci (changes requested):

Resci (approved):

PR #273, opened 2021-06-09
ArrayList.java class implemented. Tests passed.

nikdagaev:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] arrayList;
+
+    public ArrayList() {
+        arrayList = new Object[DEFAULT_CAPACITY];
+    }
+
+    private String errorOutOfBoundsMsg(int index) {
```
KseniiaMakarova, line 15:
```suggestion
    private String getErrorOutOfBoundsMessage(int index) {
```

please move private methods below public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        return "Index: " + index + ", Size: " + size;
+    }
+
+    private Object[] grow() {
+        int oldCapacity = arrayList.length;
+        if (oldCapacity > 0) {
```
KseniiaMakarova, line 21: do we need this if-else?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+            int newCapacity = (oldCapacity >> 1) + oldCapacity;
```
KseniiaMakarova, line 22: please write what does `>> 1` mean?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+            return arrayList = Arrays.copyOf(arrayList, newCapacity);
```
KseniiaMakarova, line 23: can we use `Systems.arrayCopy()` instead?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        } else {
+            return arrayList = new Object[DEFAULT_CAPACITY];
+        }
+    }
+
+    public void ensureCapacity() {
```
KseniiaMakarova, line 29: let's make this method private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        if (size + 1 >= arrayList.length) {
```
KseniiaMakarova, line 30:
```suggestion
        if (size >= arrayList.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        }
+    }
+
+    public Object[] toArray(List<T> list) {
+        Object[] listToArray = new Object[list.size()];
+        for (int i = 0; i < listToArray.length; i++) {
+            listToArray[i] = list.get(i);
+        }
+        return listToArray;
+    }
```
KseniiaMakarova, line 41: let's better remove this method) 
or alternatively - if you still want to keep it - add it to interface as well

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+
+    private void rangeCheck(int index) {
```
KseniiaMakarova, line 43:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(errorOutOfBoundsMsg(index));
```
KseniiaMakarova, line 45: there is a small problem - if you pass `index + 1` to this method, and the index will be invalid - the message will be incorrect

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        Object[] listArr = toArray(list);
+        System.arraycopy(listArr, 0,
+                arrayList, size,
+                listArr.length);
+        size += listArr.length;
```
KseniiaMakarova, line 71: what about capacity check?

let's better simply call `add()` method for each element in `list`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
-
+        rangeCheck(index + 1);
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        rangeCheck(index);
+        T toReturn = (T)arrayList[index];
```
KseniiaMakarova, line 89:
```suggestion
        T returnValue = (T) arrayList[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,115 @@
...
+        System.arraycopy(arrayList, index + 1,
+                arrayList, index,
+                size-- - index);
```
KseniiaMakarova, line 92: I am not sure about this last parameter - let's double check how much elements do we move

KseniiaMakarova (changes requested):
can you please attach the screenshot with tests results to your PR?

src/main/java/core/basesyntax/ArrayList.java
nikdagaev: oldCapacity * 1.5

src/main/java/core/basesyntax/ArrayList.java
nikdagaev: image

src/main/java/core/basesyntax/ArrayList.java
nikdagaev: According to the logic I was trying to put into my code, we always put index into this method, not index + 1

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,111 @@
...
+        System.arraycopy(arrayList, index,
+                arrayList, index + 1,
+                size++ - index);
+        arrayList[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        Object[] listArr = toArray(list);
```
KseniiaMakarova, line 40: the name of the variable is not very good( let's not shorten words

also, do we need to make an array at all here?
can we get value from list?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,111 @@
...
+        for (int i = 0; i < list.size(); i ++) {
+            add((T) listArr[i]);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index + 1);
```
KseniiaMakarova, line 48: I meant this case when I said that the exception message would not be correct

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,111 @@
...
     public void set(T value, int index) {
-
+        checkIndex(index + 1);
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
```
KseniiaMakarova, line 60: will we be able to remove element on index 5 if the size is 5?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,111 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (size + 1 > arrayList.length) {
```
KseniiaMakarova, line 89: why `size + 1` and not size?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,111 @@
...
+            grow();
+        }
+    }
+
+    private String getErrorOutOfBoundsMessage(int index) {
+        return "Index: " + index + ", Size: " + size;
+    }
```
KseniiaMakarova, line 96: I would suggest to remove this method
and generify the message in this case
like `The index is invalid!`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,108 @@
...
+                arrayList, index + 1,
+                size++ - index);
+        arrayList[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add((T) list.get(i));
```
KseniiaMakarova, line 41:
```suggestion
            add(list.get(i));
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,108 @@
...
+        checkIndex(index + 1);
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index is invalid!");
+        }
```
KseniiaMakarova, line 61: can you use your private method?

KseniiaMakarova (changes requested):
almost done! just please use private method instead logic duplication

KseniiaMakarova (approved):

PR #274, opened 2021-06-09
hw-8 added solution

OlesiaSmitiukh:
Override methods (add, addAll, get, set, remove, size, isEmpty), created helped methods expandArray and isIndexExist.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size = 0;
+    private transient T[] elementData;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
 
     @Override
     public void add(T value) {
-
+        elementData = expandArray();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not within the array");
+        }
+        elementData = expandArray();
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isIndexExist(index);
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        isIndexExist(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isIndexExist(index);
+        final T value = elementData[index];
+        T[] newElementData = elementData;
+        elementData = (T[]) new Object[newElementData.length - 1];
+        System.arraycopy(newElementData, 0, elementData, 0, index);
+        System.arraycopy(newElementData, index + 1, elementData, index, newElementData.length
+                - index - 1);
+        size--;
+        return value;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == elementData[i] || element != null && element.equals(elementData[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Can’t find the element " + element
+                + " in the array");
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] expandArray() {
+        if (size == elementData.length) {
+            int newArraySize = (elementData.length) + (elementData.length >> 1);
+            T[] newElementData = (T[]) new Object[newArraySize];
+            System.arraycopy(elementData, 0, newElementData, 0, elementData.length);
+            return newElementData;
+        }
+        return elementData;
+    }
+
+    private void isIndexExist(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not within the array");
+        }
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size = 0;
+    private transient T[] elementData;
```
KseniiaMakarova, line 8: what does `transient` mean and why is this field transient? :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size = 0;
```
KseniiaMakarova, line 7:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
+    private transient T[] elementData;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
```
```suggestion
        elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] expandArray() {
```
KseniiaMakarova, line 85:
```suggestion
    private T[] checkArraySize() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
+        if (size == elementData.length) {
+            int newArraySize = (elementData.length) + (elementData.length >> 1);
+            T[] newElementData = (T[]) new Object[newArraySize];
+            System.arraycopy(elementData, 0, newElementData, 0, elementData.length);
+            return newElementData;
+        }
+        return elementData;
+    }
+
+    private void isIndexExist(int index) {
```
KseniiaMakarova, line 95:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,100 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isIndexExist(index);
+        final T value = elementData[index];
+        T[] newElementData = elementData;
+        elementData = (T[]) new Object[newElementData.length - 1];
+        System.arraycopy(newElementData, 0, elementData, 0, index);
```
KseniiaMakarova, line 57: we don't need to make the capacity smaller when we remove the element

let's just shift the elements within the same array

KseniiaMakarova (changes requested):
overall very nice!

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: How I understand this term, This field is like "temporary", won’t  save to file(won’t be part of the serialization, but I haven’t delved into this topic).

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: It’s logical 💁🏼 because default value is 0. Done ✔️

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: Done ✔️ Is it connected with transient field?

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: done ✔️

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: done ✔️

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: done ✔️ I had some trouble with this method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private transient T[] elementData;
+
+    public ArrayList() { elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
```
```suggestion
    public ArrayList() { 
        elementData = (T[]) new Object[DEFAULT_CAPACITY];
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        checkIndex(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T value = elementData[index];
+        T[] newElementData = elementData;
```
KseniiaMakarova, line 54: can we go without this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
+        System.arraycopy(newElementData, index + 1, elementData, index, newElementData.length
+                - index - 1);
```
KseniiaMakarova, line 56: let's double-check the last parameter here

should we count from the capacity or the actual size?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] checkArraySize() {
+        if (size == elementData.length) {
+            int newArraySize = (elementData.length) + (elementData.length >> 1);
```
KseniiaMakarova, line 84:
```suggestion
            int newArraySize = elementData.length + (elementData.length >> 1);
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: ✔️  I don't know how it's happened🤯

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: Can

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: Changed. I should use size.

src/main/java/core/basesyntax/ArrayList.java
OlesiaSmitiukh: ✔️

KseniiaMakarova (approved):

PR #284, opened 2021-06-10
HW-8 Created my ArrayList

docer95:
Overridden methods add/get/set/remove/size/isEmpty

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,49 +1,96 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static int MAX_SIZE_ARRAY = 10;
+    private T[] myList;
+    private int size;
+
+    public ArrayList() {
+        myList = (T[]) new Object[MAX_SIZE_ARRAY];
+    }
 
     @Override
     public void add(T value) {
-
+        if (size + 1 >= myList.length) {
+            expandingMaxSize();
+        }
+        myList[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
+        if (size + 1 == MAX_SIZE_ARRAY) {
+            expandingMaxSize();
+        }
+        System.arraycopy(myList, index, myList, index + 1, size++ - index);
+        myList[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index + 1);
+        return myList[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index + 1);
+        myList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T newTempList = myList[index];
+        System.arraycopy(myList, index + 1, myList, index, size - 1 - index);
+        size--;
+        return newTempList;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (myList[i] == element || (myList[i] != null && myList[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("List haven't this element!");
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index incorrect");
+        }
+    }
+
+    private T[] expandingMaxSize() {
+        int oldCapacity = myList.length;
+        if (oldCapacity > 0) {
+            int newCapacity = (oldCapacity >> 1) + oldCapacity;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static int MAX_SIZE_ARRAY = 10;
```
KseniiaMakarova, line 7: is everything ok here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+    private int size;
+
+    public ArrayList() {
+        myList = (T[]) new Object[MAX_SIZE_ARRAY];
+    }
 
     @Override
     public void add(T value) {
-
+        if (size + 1 >= myList.length) {
```
KseniiaMakarova, line 17: why `size + 1` and not just `size`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+            expandingMaxSize();
+        }
+        myList[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
+        if (size + 1 == MAX_SIZE_ARRAY) {
```
KseniiaMakarova, line 26: same
also - is it correct to compare with MAX_SIZE_ARRAY?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
     public void set(T value, int index) {
-
+        checkIndex(index + 1);
+        myList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
```
KseniiaMakarova, line 54: should you be able to remove element at index that equals size?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+        T newTempList = myList[index];
```
KseniiaMakarova, line 55: is the variable name informative?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (myList[i] == element || (myList[i] != null && myList[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("List haven't this element!");
```
KseniiaMakarova, line 68:
```suggestion
        throw new NoSuchElementException("List doesn't have this element!");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index incorrect");
```
KseniiaMakarova, line 83:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("This index is incorrect");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+        }
+    }
+
+    private T[] expandingMaxSize() {
```
KseniiaMakarova, line 87:
```suggestion
    private T[] expandArray() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+        int oldCapacity = myList.length;
+        if (oldCapacity > 0) {
```
KseniiaMakarova, line 89: why do you have this check?
can it be 0 or less?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,96 @@
...
+            int newCapacity = (oldCapacity >> 1) + oldCapacity;
+            return myList = Arrays.copyOf(myList, newCapacity);
```
KseniiaMakarova, line 91: let's use `System.arraycopy()`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
docer95: No I can't. This value is not included in the index check condition. But I made a mistake in the verification and now, wherever I used it in the methods, everything needs to be corrected.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,49 +1,97 @@
...
     public void set(T value, int index) {
-
+        checkIndex(index);
+        myList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size || index < 0) {
```
KseniiaMakarova, line 55: should you be able to remove from index 5 when your size is 5?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
docer95: yes. We must use checkIndex instead if

KseniiaMakarova (approved):
nice!

PR #305, opened 2021-08-20
Created own ArrayList class and implemented some methods

kovalskiy7777:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+    private T fastRemove(int index) {
+        T elementToRemove = array[index];
+        int lengthLastPartArray = size - index;
+        int correctLastElement = lengthLastPartArray == 1 ? 1 : 0;
+        System.arraycopy(array, index + 1, array, index, lengthLastPartArray - correctLastElement);
+        size--;
+        return elementToRemove;
+    }
+
+    private T[] reSize() {
```
boroda4436, line 109:
```suggestion
    private T[] resize() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexForAdd(int index, int adjustment) {
```
boroda4436, line 94:
```suggestion
    private void checkIndexForAdd(int index) {
```
Consider not using `adjustment` parameter in this method

boroda4436 (changes requested):

PR #312, opened 2021-08-25
implemented own class ArrayList

VladGural:
implemented own class ArrayList

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity;
+    private int size;
+    private Object[] elements;
+
+    public ArrayList() {
+        capacity = DEFAULT_CAPACITY;
+        elements = new Object[capacity];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        checkAdd();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexAddRemove(index);
+        checkAdd();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexGetSet(index);
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndexGetSet(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexAddRemove(index);
+        T removedItem = get(index);
+        int length = size == capacity ? size - index - 1 : size - index;
+        System.arraycopy(elements, index + 1, elements, index, length);
+        elements[--size] = null;
+        return removedItem;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == null ? element == null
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int capacity = 10;
```
MaliukDaria, line 6: let's make it as constant and call like INITIAL_CAPACITY or smth like that

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private int size = 0;
```
MaliukDaria, line 7: we dont need to initialize `size` by 0, it initialized to 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private Object[] elements = new Object[capacity];
```
MaliukDaria, line 8: why array of Objects? Lets initialize it in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        checkGrow();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexAddRemove(index);
+        checkGrow();
```
MaliukDaria, line 20: do we need this check here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        capacity += capacity >> 1;
```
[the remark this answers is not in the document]
MaliukDaria: maybe its better to use `size` variable instead of capacity?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
VladGural: yes we need. If index out of range we throw exception and if size will be more than capacity we need grow our array

VladGural:
maybe its better to use size variable instead of capacity?
No we increase our real array (capacity = capacity * 1.5) and make new array[new capacity]. But new remain our logical array and size is same.

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
VladGural: No we increase our real array (capacity = capacity * 1.5) and make new array[new capacity]. But new remain our logical array and size is same.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private int capacity;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        capacity = INITIAL_CAPACITY;
+        size = 0;
```
MaliukDaria, line 13: we don't need to initialize `size` variable explicitly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity;
```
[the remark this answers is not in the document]
MaliukDaria: Lets remove `capacity` variable

MaliukDaria (changes requested):

VladGural:
I've understanded about variable 'capacity' and deleted it!!!

MaliukDaria (approved):
Good Job! 🔥

PR #317, opened 2021-08-28
Array list homework

romuth:
In the process of correcting inaccuracies

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -18,12 +18,16 @@ public class ArrayList<T> implements List<T> {
 
     @Override
     public void add(T value) {
-        if (checkForEmpty(value)) {
+        if (!arrayIsEmpty) {
             if (!ensureCapacity()) {
                 grow();
             }
             arrayData[size] = value;
             size++;
+        } else {
+            arrayData[size] = value;
+            size++;
+            arrayIsEmpty = false;
         }
     }
 
@@ -38,13 +42,12 @@ public class ArrayList<T> implements List<T> {
             size++;
             }
         else if (indexCapacity(index) && ensureCapacity()) {
-            arrayCopySet(index);
+            arrayCopyAdd(index);
             arrayData[index] = value;
             size++;
-        }
-        else if (indexCapacity(index) && !ensureCapacity()) {
+        } else {
             grow();
-            arrayCopySet(index);
+            arrayCopyAdd(index);
             arrayData[index] = value;
             size++;
         }
@@ -86,7 +89,8 @@ public class ArrayList<T> implements List<T> {
         if (indexCapacity(index) && ensureCapacity()) {
             arrayCopyRemove(index);
             size--;
-        } else if (index == size && ensureCapacity()) {
+        } else if ((index == size && !ensureCapacity())
+                 || (indexCapacity(index) && !ensureCapacity())) {
             grow();
             arrayCopyRemove(index);
             size--;
@@ -101,9 +105,6 @@ public class ArrayList<T> implements List<T> {
             if ((element == null
                     && element == arrayData[i])
                     || element != null && element.equals(arrayData[i])) {
-                if (!ensureCapacity()) {
-                    grow();
-                }
                 arrayCopyRemove(i);
                 size--;
                 return element;
@@ -131,17 +132,7 @@ public class ArrayList<T> implements List<T> {
         currentSizeOfArray = (int) (currentSizeOfArray * GROW_SIZE);
     }
 
-    private boolean checkForEmpty(T value) {
-        if (arrayIsEmpty) {
-            arrayData[0] = value;
-            size++;
-            arrayIsEmpty = false;
-            return false;
-        }
-        return true;
-    }
-
-    private void arrayCopySet(int index) {
+    private void arrayCopyAdd(int index) {
         int numberOfCopiedElements = size +1;
             System.arraycopy(arrayData, index, arrayData, ++index, numberOfCopiedElements - index);
     }
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -27,20 +27,20 @@ public class ArrayList<T> implements List<T> {
     @Override
     public void add(T value, int index) {
         if (checkForEmpty(value)) {
-            if (index < currentSizeOfArray
-            && index < size && ensureCapacity()) {
-                copyValues(index);
+            if (index > currentSizeOfArray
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_SIZE = 10;
+    private static final int INITIAL_SRC_POS = 0;
+    private static final double GROW_SIZE = 1.5;
+    private boolean arrayIsEmpty = true;
```
MaliukDaria, line 9: We dont need this variable, we just can check size for 0

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+    private T[] arrayData;
+    private int size;
+    private int currentSizeOfArray = INITIAL_SIZE;
```
MaliukDaria, line 12: What is the difference between currentSizeOfArray and size?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+
+    @SuppressWarnings({"unchecked"})
+    public ArrayList() {
+        arrayData = (T[]) new Object[INITIAL_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        if (!arrayIsEmpty) {
```
MaliukDaria, line 21: we can do it without this check?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        }
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (indexIsNegative(index) || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index are not exist, your index is "
+                    + index + " last index is " + size);
+        } else if (indexIsLegal(index) && ensureCapacity()) {
```
MaliukDaria, line 39: you check index two times here. Lets make index check in indexIsLegal method. And we can throw ArrayListIndexOutOfBoundsException in indexIsLegal method, this will reduce the amount of duplicate code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() >= currentSizeOfArray) {
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            arrayData[size] = list.get(i);
+            size++;
+        }
```
MaliukDaria, line 57: lets use your `public void add(T value, int index)` method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+
+    private boolean ensureCapacity() {
+        return (size + 1) >= currentSizeOfArray;
+    }
+
+    private boolean indexIsLegal(int index) {
+        return (index >= 0 && index < size);
+    }
+
+    private boolean indexIsNegative(int index) {
```
MaliukDaria, line 137: We can add this check in indexIsLegal method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+
+    private void arrayCopyRemove(int index) {
+        System.arraycopy(arrayData, ++index, arrayData, --index, size-- - index);
+    }
+
+    private boolean ensureCapacity() {
+        return (size + 1) >= currentSizeOfArray;
+    }
+
+    private boolean indexIsLegal(int index) {
```
MaliukDaria, line 133:
```suggestion
    private boolean isIndexLegal(int index) {
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
romuth: currentSizeOfArray - это общий размер массива, без учета есть ли в нем элементы.
size - количество элементов в массиве, конечно же.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_SRC_POS = 0;
+    private static final double GROW_SIZE = 1.5;
+    private T[] arrayData;
+    private int size;
+    private int fullSizeOfArray = 10;
```
MaliukDaria, line 10: let's make this variable constant field and rename it to INITIAL_CAPASITY or smth like that

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        return value;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null
+                    && arrayData[i] == null)
+                    || element != null && element.equals(arrayData[i])) {
```
MaliukDaria, line 64:
```suggestion
            if ((element == arrayData[i])
                    || element != null && element.equals(arrayData[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        return size <= 0;
```
MaliukDaria, line 78: how could be size a negative number?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+    }
+
+    @SuppressWarnings({"unchecked"})
+    private void grow() {
+        T[] arrayDataTemp = (T[]) new Object[(int) (fullSizeOfArray * GROW_SIZE)];
+        System.arraycopy(arrayData, INITIAL_SRC_POS,
+                arrayDataTemp, INITIAL_SRC_POS, arrayData.length);
+        arrayData = arrayDataTemp;
+        fullSizeOfArray = (int) (fullSizeOfArray * GROW_SIZE);
```
MaliukDaria, line 87: we don't need to store this data in a separate variable as we can get in from `arrayData.length`

MaliukDaria (changes requested):

MaliukDaria (approved):
Goog Job! 👍🏻

PR #320, opened 2021-08-28
Exceptions informative, no redundant variables, resized array is sepa…

Evgen-Krasnenkov:
…rated, all checked

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,127 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int STANDARD_CAPACITY = 10;
+    private T[] arrayList;
+    private int size;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[STANDARD_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (hasArraySpace()) {
+            arrayList[size] = value;
+            size++;
+            return;
+        }
+        extendCapacity();
+        arrayList[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (!hasArraySpace()) {
+            extendCapacity();
+        }
+        if (index < 0 || index > arrayList.length || index > size + 1) {
+            throw new ArrayListIndexOutOfBoundsException("Your index does not suit rules");
+        }
+        System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
+        arrayList[index] = value;
+        size++;
 
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list != null) {
+            T[] addArray = (T[]) new Object[list.size()];
+            for (int i = 0; i < list.size(); i++) {
+                addArray[i] = list.get(i);
+            }
+            while (arrayList.length < size + addArray.length) {
+                extendCapacity();
+            }
+            System.arraycopy(addArray, 0, arrayList, size, addArray.length);
+            size = size + addArray.length;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Adding list is null");
+        }
     }
 
     @Override
     public T get(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("An element you"
+                    + " want to get does not exist");
+        }
+        if (size != 0) {
+            return arrayList[index];
+        }
         return null;
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("An element you"
+                    + " want to remove does not exist");
+        }
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("An element you"
+                    + " want to remove does not exist");
+        }
+        final T removedElement = arrayList[index];
+        System.arraycopy(arrayList, index + 1, arrayList, index, size - index - 1);
+        size--;
+        arrayList[size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, arrayList[i])) {
+                System.arraycopy(arrayList, i + 1, arrayList, i, size - i - 1);
+                arrayList[size] = null;
+                size--;
+                return element;
+            }
+        }
+        throw new NoSuchElementException("The value you want to remove does not exist");
     }
 
     @Override
     public int size() {
-        return 0;
+
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void extendCapacity() {
+        int tempLength = arrayList.length;
+        T[] tempArray = (T[]) new Object[tempLength];
+        tempArray = arrayList;
+        arrayList = (T[]) new Object[(tempLength * 3) / 2 + 1];
+        System.arraycopy(tempArray, 0, arrayList, 0, size);
+    }
+
+    private boolean hasArraySpace() {
+        return size < arrayList.length;
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int STANDARD_CAPACITY = 10;
```
Max-Pochepets, line 7:
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+
+    private void extendCapacity() {
+        int tempLength = arrayList.length;
+        T[] tempArray = (T[]) new Object[tempLength];
+        tempArray = arrayList;
+        arrayList = (T[]) new Object[(tempLength * 3) / 2 + 1];
+        System.arraycopy(tempArray, 0, arrayList, 0, size);
+    }
+
+    private boolean hasArraySpace() {
```
Max-Pochepets, line 124: no need to create a private method for one line of code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void extendCapacity() {
+        int tempLength = arrayList.length;
+        T[] tempArray = (T[]) new Object[tempLength];
```
Max-Pochepets, line 118: let's check for capacity here. if it's not enough, resize the array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (!hasArraySpace()) {
+            extendCapacity();
+        }
+        if (index < 0 || index > arrayList.length || index > size + 1) {
+            throw new ArrayListIndexOutOfBoundsException("Your index does not suit rules");
```
Max-Pochepets, line 33:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Wrong index.");
```
if to keep it simple

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        arrayList[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (!hasArraySpace()) {
+            extendCapacity();
+        }
+        if (index < 0 || index > arrayList.length || index > size + 1) {
```
Max-Pochepets, line 32: this check can be modified and moved to a private method to be used by all methods. try to think about checking if index is **inside** the bounds, not outside

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        if (list != null) {
+            T[] addArray = (T[]) new Object[list.size()];
+            for (int i = 0; i < list.size(); i++) {
+                addArray[i] = list.get(i);
+            }
+            while (arrayList.length < size + addArray.length) {
+                extendCapacity();
+            }
+            System.arraycopy(addArray, 0, arrayList, size, addArray.length);
+            size = size + addArray.length;
```
Max-Pochepets, line 52: this can be simplified up to several lines. HINT: you have `add(T value)` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, arrayList[i])) {
+                System.arraycopy(arrayList, i + 1, arrayList, i, size - i - 1);
+                arrayList[size] = null;
+                size--;
+                return element;
```
Max-Pochepets, line 99:
```suggestion
                return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        size--;
+        arrayList[size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, arrayList[i])) {
```
Max-Pochepets, line 95: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#dont-use-class-objects)

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: Renamed

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: simplified

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: new method created

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: done

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: add by value has been used

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: remove by index has been used in remove by element

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: Objects.equals - replaced by logic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -113,15 +84,21 @@ public boolean isEmpty() {
...
     }
 
-    private boolean hasArraySpace() {
-        return size < arrayList.length;
+    private void checkIndexValidity(int index) {
+        if (index >= 0 && index < size) {
+            return;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index.");
+        }
```
Max-Pochepets, line 102:
```suggestion
        if (index < 0 && index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Wrong index.");
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -113,15 +84,21 @@ public boolean isEmpty() {
...
 
-    private void extendCapacity() {
-        int tempLength = arrayList.length;
-        T[] tempArray = (T[]) new Object[tempLength];
-        tempArray = arrayList;
-        arrayList = (T[]) new Object[(tempLength * 3) / 2 + 1];
-        System.arraycopy(tempArray, 0, arrayList, 0, size);
+    private void extendCapacityIfNecessary() {
+        if (size == arrayList.length) {
+            int tempLength = arrayList.length;
```
Max-Pochepets, line 89:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,87 +1,61 @@
...
         }
-        System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
-        arrayList[index] = value;
-        size++;
-
     }
 
     @Override
     public void addAll(List<T> list) {
         if (list != null) {
```
Max-Pochepets, line 35: it's better to use `for-each` loop here, in that way you won't need to check list for null

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,87 +1,61 @@
...
-            extendCapacity();
-        }
-        if (index < 0 || index > arrayList.length || index > size + 1) {
-            throw new ArrayListIndexOutOfBoundsException("Your index does not suit rules");
+        extendCapacityIfNecessary();
+        if (index >= 0 && index <= size) {
+            System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
+            arrayList[index] = value;
+            size++;
+        } else {
```
Max-Pochepets, line 28: how can we do this with only if, without else?

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: return?

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: checkIndexValidity - method created

src/main/java/core/basesyntax/ArrayList.java
Evgen-Krasnenkov: http://prntscr.com/1qvche9 - if I change this the logic will break

Evgen-Krasnenkov: done

Max-Pochepets (approved):

PR #322, opened 2021-08-29
realized jv-array-list App

sashahordenin:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] array;
```
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    private int size;
+    private int minCapacity;
```
MaliukDaria, line 9: do we need this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+        this.minCapacity = DEFAULT_CAPACITY;
```
MaliukDaria, line 14: do we need `this` keyword here(and check another places in the code)? We don't need to initialize `size` variable explicitly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        } else {
+            add(value);
+        }
+    }
 
+    private T[] grow(int minCapacity) {
+        T[] arrayWithNewCapacity = (T[]) new Object[getNewCapacity(minCapacity)];
+        copyArray(this.array, arrayWithNewCapacity);
+        return arrayWithNewCapacity;
     }
```
MaliukDaria, line 44: lets make this method void and with no input parameters, we can just use `size` variable and increase the size of the array in 1,5 times

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+        this.minCapacity = DEFAULT_CAPACITY;
+    }
+
     @Override
     public void add(T value) {
-
+        this.array = grow(minCapacity);
```
MaliukDaria, line 19: Our array grows every time we add new element?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-    @Override
-    public void addAll(List<T> list) {
+    private int getNewCapacity(int minCapacity) {
+        if (size + 1 > array.length) {
+            return this.minCapacity = minCapacity + (minCapacity >> 1);
+        }
+        return minCapacity;
+    }
 
+    private void checkIndexForAddMethod(int index) {
```
MaliukDaria, line 53: As you use this method only ones we can move this logic to `add` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-
+        this.array = grow(minCapacity);
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexForAddMethod(index);
+        if (index + 1 <= minCapacity) {
```
MaliukDaria, line 27: You have already checked index in `checkIndexForAddMethod` why do we need this check?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+            this.array = grow(minCapacity);
+            T[] newArray = grow(minCapacity);
+            System.arraycopy(this.array, 0, newArray, 0, index);
+            newArray[index] = value;
+            System.arraycopy(this.array, index, newArray, index + 1, minCapacity - index - 1);
+            copyArray(newArray, this.array);
+            size++;
```
MaliukDaria, line 34: I can see 2 array grows and 3 System.arraycopy in this method... we can use only 1 System.arraycopy method here and we don't need grow method here at all.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        System.arraycopy(array, index + 1, array, index, minCapacity - index - 1);
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if ((array[i] == null && element == null)
```
MaliukDaria, line 95: do we need this check `element == null`?

MaliukDaria (changes requested):
Lets place private methods after public

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: changed on values[i] == element

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: deleted variable minCapacity

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: done

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: Done

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: removed

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: removed

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: Removed

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: Simplified almost all. I left one grow method call here and one System.arraycopy. If I'll remove this one, I can't grow capacity for this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        System.arraycopy(values, index, values, index + 1, size - index);
+        values[index] = value;
+        size++;
+    }
 
+    private void grow() {
+        T[] arrayWithNewCapacity = (T[]) new Object[getNewCapacity(values.length)];
+        System.arraycopy(values, 0, arrayWithNewCapacity, 0, values.length);
+        values = arrayWithNewCapacity;
+    }
```
MaliukDaria, line 37: let's check first if our array is full and grow it only if there is no free space left. (There is no need to copy array every time). Than we should probably rename the method to `ensureCapacity` or smth like this. And lets place private methods after public

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
sashahordenin: @MaliukDaria Done, thanks!)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexForRemoveGetSetMethods(index);
+        T removedElement = values[index];
+        System.arraycopy(values, index + 1, values, index, values.length - index - 1);
+        size--;
+        return removedElement;
```
MaliukDaria, line 58:
```suggestion
        size--;
        values[size] = null;
        return removedElement;
```

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #323, opened 2021-08-29
Hw ArrayList

osa240:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final double GROW_COEFFICIENT = 1.5;
+    private int currentCapacity = INITIAL_CAPACITY;
```
MaliukDaria, line 8: do we need this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == currentCapacity) {
+            resize();
+        }
```
MaliukDaria, line 20: lets move `size == currentCapacity` check into `resize()` method (this will reduce number of duplicated code)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        T[] temporaryArray = (T[]) new Object[size];
+        System.arraycopy(elementData, 0, temporaryArray, 0, size);
+        if (size == currentCapacity) {
+            resize();
+        }
```
MaliukDaria, line 31: do we need resize here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        isIndexValid(index, "ADD");
```
MaliukDaria, line 32: i think this check should be in the beginning of the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        elementData[index] = value;
+        System.arraycopy(temporaryArray,
```
MaliukDaria, line 34: we can implement this method without temporary array and use `System.arraycopy` method only ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        T[] temporaryArray = (T[]) new Object[list.size()];
+        for (int i = 0; i < temporaryArray.length; i++) {
+            temporaryArray[i] = list.get(i);
+        }
+        System.arraycopy(temporaryArray,
+                0,
+                elementData,
+                size,
+                list.size());
+        size += list.size();
```
MaliukDaria, line 53: lets use `add(T value, int index)` here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        if (temporaryIndex != -1) {
+            System.arraycopy(elementData,
+                    temporaryIndex + 1,
+                    elementData,
+                    temporaryIndex,
+                    size - temporaryIndex - 1);
+            size--;
+            return oldRecord;
+        }
+        throw new NoSuchElementException("Element not find");
```
MaliukDaria, line 103: lets use our `remove(int index)` method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void resize() {
+        currentCapacity *= GROW_COEFFICIENT;
+        T[] temporaryArray = (T[]) new Object[size];
+        System.arraycopy(elementData, 0, temporaryArray, 0, size);
+        elementData = (T[]) new Object[currentCapacity];
+        System.arraycopy(temporaryArray, 0, elementData, 0, size);
```
MaliukDaria, line 121: we can make it without `temporaryArray` and with only one `System.arraycopy` method call

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    }
+
+    private void isIndexValid(int index, String operation) {
+        if ((operation.equals("ADD") && (index > size || index < 0))
+                || (operation.equals("noADD") && (index < 0 || index >= size))) {
```
MaliukDaria, line 126: lets move "ADD" "noADD" to constants

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        System.arraycopy(elementData, index,elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i), size);
```
MaliukDaria, line 36:
```suggestion
            add(list.get(i));
```

i think it's better to use `add(T value)` method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+                    || (elementData[i] != null
+                    && elementData[i].equals(element))) {
+                temporaryIndex = i;
+                break;
+            }
+        }
+        if (temporaryIndex != -1) {
+            return remove(temporaryIndex);
+        }
+        throw new NoSuchElementException("Element not find");
```
MaliukDaria, line 79:
```suggestion
        int temporaryIndex = -1;
        for (int i = 0; i < size; i++) {
            if (elementData[i] == element
                    || (elementData[i] != null
                    && elementData[i].equals(element))) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element not found");
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
osa240: exactly)

src/main/java/core/basesyntax/ArrayList.java
osa240: you are right - like this will be better.

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #326, opened 2021-08-29
Methods realization

denyssmei:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
```
Max-Pochepets, line 19: let's place this check inside `resize()` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((array[i] == null && value == null)
+                    || (array[i] != null && array[i].equals(value))) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private T[] addingElementInside(T value, int index, T[] oldArray) {
```
Max-Pochepets, line 114: this code is used only once, no need to create private method for it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        T[] newArray = (T[]) new Object[array.length];
```
Max-Pochepets, line 115: can we manage without `newArray`? can we work with the original one?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
-
+        if (size == array.length) {
+            array = resizeArr(array);
+        }
+        if (index < size && index >= ZERO_INDEX) {
+            array = addingElementInside(value, index, array);
+        } else if (index == size) {
+            add(value);
+            return;
+        } else {
```
Max-Pochepets, line 35: try to simplify this if-else construction

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == ZERO_INDEX;
+    }
+
+    private int indexOf(T value) {
```
Max-Pochepets, line 104: same, code used only once

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Max-Pochepets: we can have have one if to check if the index is out of bounds to throw exception. and one if-else construction: if index == size - add to the end; else - line 27 to 32. **AND** think about using only one `System.arraycopy()`, you don't need two of them

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+            if ((array[i] == null && element == null)
+                    || (array[i] != null && array[i].equals(element))) {
+                indexOfElement = i;
+                break;
+            }
+        }
+        if (indexOfElement == -1) {
+            throw new NoSuchElementException();
+        }
+        return remove(indexOfElement);
```
Max-Pochepets, line 97: can we place it somewhere on lines 88-93? and in that way we won't need second if, but we'll need the exception

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #330, opened 2021-08-30
Implemented lists methods.

VladyslavLekhter:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double INCREASE_CAPACITY = 1.5;
+    private static final int ARRAYS_ZERO_BORDER = 0;
+    private static final int INCREASE_INDEX_FOR_COPY = 1;
+    private T[] elementData = (T[]) new Object[DEFAULT_CAPACITY];
```
MaliukDaria, line 10: lets initialize this array in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+    private int size = 0;
```
MaliukDaria, line 11: we don't need to initialize `size` variable explicity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public T[] grow() {
```
MaliukDaria, line 108: lets make this method `void`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        if (elementData.length < size + list.size()) {
+            elementData = grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            elementData[size] = list.get(i);
+            size++;
+        }
```
MaliukDaria, line 46: lets use `add(T value, int index)` here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
 
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < size && index >= ARRAYS_ZERO_BORDER) {
```
MaliukDaria, line 52: lets move this check to separate method `if (index < size && index >= ARRAYS_ZERO_BORDER)`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+            elementData = grow();
+        }
+        size++;
+        if (index < size && index >= ARRAYS_ZERO_BORDER) {
+            System.arraycopy(elementData, index, elementData,
+                    index + INCREASE_INDEX_FOR_COPY,
+                    size - (index + INCREASE_INDEX_FOR_COPY));
+            elementData[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Index is bigger than arrays size");
```
MaliukDaria, line 34: if index is negative, than this message became not informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        }
+        T oldElement = elementData[index];
+        if (size != elementData.length) {
+            System.arraycopy(elementData, index + INCREASE_INDEX_FOR_COPY,
+                    elementData, index, size - index);
+            size--;
+        } else {
+            elementData[index] = null;
+            size--;
+        }
```
MaliukDaria, line 82:
```suggestion
        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
        size--;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public T[] grow() {
+        int oldCapacity = elementData.length;
+        int newCapacity = (int) (oldCapacity * INCREASE_CAPACITY);
+        T[] newElementData = (T[]) new Object[newCapacity];
```
MaliukDaria, line 111:
```suggestion
        int newCapacity = (int) (elementData.length * INCREASE_CAPACITY);
        T[] newElementData = (T[]) new Object[newCapacity];
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double INCREASE_CAPACITY = 1.5;
```
```suggestion
    private static final double INCREASE_COEFFICIENT = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (elementData.length == size) {
+            grow();
```
MaliukDaria, line 20: we could move this check `if (elementData.length == size)` into `grow` method(this will reduce number if duplicated code). If we do this, then it might be worth to rename `grow` method to `ensureCapacity` or smth like this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (elementData.length == size) {
+            grow();
+        }
+        size++;
+        if (checkIndex(index)) {
```
MaliukDaria, line 32: we can place `throw new ArrayListIndexOutOfBoundsException("Index out of bounds array size");` into `checkIndex` method, this will reduce number of duplicated code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (elementData.length == size) {
+            grow();
+        }
+        size++;
```
MaliukDaria, line 31: first, you increase the size of the array, and then you check for the index, and what if the check fails and throws `ArrayListIndexOutOfBoundsException`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds array size");
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (elementData.length < size + list.size()) {
+            grow();
+        }
```
MaliukDaria, line 46: you have already do this check in the `add(T value)` method, do we need it here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void grow() {
+
```
MaliukDaria, line 105: redundant empty line

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T oldElement = elementData[index];
+        System.arraycopy(elementData, index + INCREASE_INDEX_FOR_COPY,
+                elementData, index, size - index - INCREASE_INDEX_FOR_COPY);
+        size--;
+        return oldElement;
```
MaliukDaria, line 63:
```suggestion
        element[--size] = null;
        return oldElement;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void ensureCapacity() {
```
MaliukDaria, line 88: why this method is public?

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #331, opened 2021-08-30
Hw-array-list-rostyslav

RostyslavF5R:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size == 0;
+    }
+
+    private void chekIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds.");
+        }
+    }
+
+    private T[] grow() {
```
MaliukDaria, line 95: lets make `grow()` as void method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (elementData.length == size) {
+            elementData = grow();
+        }
```
MaliukDaria, line 18: we can add this check `if (elementData.length == size)` into `grow()` method and rename it to `ensureCapacity` or smth like this, this will reduce number of duplicated code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        for (int i = 0; i < size; i++) {
+            if (element == elementData[i] || (element != null
+                    && element.equals(elementData[i]))) {
+                index = i;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("No such element");
+        }
+        return remove(index);
```
MaliukDaria, line 76: can we simplify this? We can make it without `index` variable

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        chekIndex(index);
+        T removedElement = elementData[index];
+        System.arraycopy(elementData,index + 1, elementData, index, size - index - 1);
+        size--;
+        return removedElement;
```
MaliukDaria, line 57:
```suggestion
        size--;
        elementData[size] = null;
        return removedElement;
```

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #334, opened 2021-08-30
Create ArrayList class which implements the List interface.

SonyaPakhomova:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private T[] array;
```
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+    private int size = 0;
```
MaliukDaria, line 8: we don't need to initialize `size` variable explicitly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_SIZE];
+    }
+
+    private void grow() {
+        T[] arrayCopy = (T[]) new Object[array.length * 3 / 2];
```
MaliukDaria, line 15: let's make constant GROV_COEFFICIENT = 1.5 and use it to create array with new capacity

MaliukDaria, line 15:
```suggestion
        T[] newArray = (T[]) new Object[array.length * 3 / 2];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        System.arraycopy(array, 0, arrayCopy, 0, size);
+        array = arrayCopy;
+    }
+
+    private void testIndex(int index) {
```
MaliukDaria, line 20: let's place private methods after public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private T[] array;
+    private int size = 0;
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_SIZE];
+    }
+
+    private void grow() {
```
MaliukDaria, line 14: we could move this check `if (size == array.length)` into `grow` method(this will reduce number if duplicated code). If we do this, then it might be worth to rename `grow` method to `ensureCapacity` or smth like this

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Element does not exist");
+        }
+        if (size == values.length) {
+            ensureCapacity();
+        }
```
MaliukDaria, line 28:
```suggestion
            ensureCapacity();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        testIndex(index);
+        return (T) values[index];
```
MaliukDaria, line 44:
```suggestion
        return values[index];
```

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #335, opened 2021-08-30
array-list-hw-solution

MiaHrynda:
… value) method which checks the element count in array, and if count equals array length, then it calls grow(), which copies this array to new array with length bigger in 1.5 times. Else it just adds new element.

Implemented add(T value, int index), which shifts all elements of the array from the current index to the next cell and writes a new value to the input index.
Implemented addAll(List<T> list), which adds all elements from list to elementData[].
Implemented get(int index), which gets element by index from elementData[].
Implemented set(T value, int index), which set value by this index to elementData[].
Implemented remove(int index), which shifts all elements of the array from the current index to the previous cell and rewrite the previous index element.
Implemented remove(T element), which checks if such element exists, and if it is true, then it removes it.
Implemented size(), which returns the number of elements in the array.
Implemented isEmpty(), which checks if array is empty.
Implemented checkIfIndexIsValid(int index, int length), which checks if index is > 0 and is < the number of elements in the array.
Implemented grow(), which copies elementData array to new array with length bigger in 1.5 times.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,112 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] elementData;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        checkSize();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
+        }
+        checkSize();
+        System.arraycopy(elementData,
+                index,
+                elementData,
+                index + 1,
+                size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
+
 
     @Override
     public T get(int index) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] elementData;
```
netiazhevskyi, line 8: I think we need to use generics instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
```
netiazhevskyi, line 11: generics here as well

netiazhevskyi (approved):
looks good to me :)

src/main/java/core/basesyntax/ArrayList.java
MiaHrynda: Ye, maybe in this case it's better to make it T[] type not to cast all the time to T[]. But the idea was to make this elementData[] Object type like in the real realization of ArrayList.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == element
+                    || elementData[i] != null && elementData[i].equals(element)) {
+                remove(i);
+                return element;
```
Max-Pochepets, line 74:
```suggestion
                return remove(i);
```

Max-Pochepets (approved):

PR #337, opened 2021-08-30
Implemented all methods in ArrayList class. Created two addition methods: grow(), checkIndex().

sich-mykhailo:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] defaultArray;
```
MaliukDaria, line 8:
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] grow() {
```
MaliukDaria, line 86: let's make `grow` method as void. And we could move this check `if (size == defaultArray.length)` into `grow` method(this will reduce number if duplicated code). If we do all of this, then it might be worth to rename `grow` method to `ensureCapacity` or smth like this

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index is not correct!");
+        }
+        ensureCapacity();
+        System.arraycopy(values, index, values, index + 1, size - index);
+        this.values[index] = value;
```
MaliukDaria, line 28: do we need `this` keyword here. (Check it in other places in code)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T element = values[index];
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        size--;
+        return element;
```
MaliukDaria, line 57:
```suggestion
        size--;
        values[size] = null;
        return element;
```

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #350, opened 2021-08-31
implement ArrayList

Nazar-Hetun:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+            arrayList = addSize(arrayList);
+        }
+        arrayList[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == arrayList.length) {
```
MiaHrynda, line 24: You use this comparison logic twice. It would be better to put it in a separate method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= 0 && index < size) {
```
MiaHrynda, line 45: Here is the same. You do not follow the DRY principle, which tells us "Don't Repeat Yourself". (https://dzone.com/articles/software-design-principles-dry-and-kiss) The logic of your code repeats. You should move it to the separate private method.

src/main/java/core/basesyntax/NoSuchElementException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public class NoSuchElementException extends RuntimeException {
```
MiaHrynda, line 3: You shouldn't create your own exception. It's a custom exception, so you can just import it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     public void add(T value, int index) {
-
+        if (size == arrayList.length) {
+            arrayList = addSize(arrayList);
+        }
+        if (index >= 0 && index <= size) {
+            System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
+            arrayList[index] = value;
+            size++;
+            return;
```
MiaHrynda, line 31: You don't need return statement here

MiaHrynda (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private T[] arrayList;
+    private int size;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[10];
```
MiaHrynda, line 10: You should create constant variable of Array capacity.

MiaHrynda (changes requested):

MiaHrynda (approved):
Looks good now!

PR #355, opened 2021-09-12
My homework jv-array-list is done

roman-hrechanyi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
```
RomanDubovskyi, line 42: if condition with throwing exception repeats itself several times, it's better to move it to the separate private method to avoid code duplication

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return (T) removedValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T removedValue = null;
+        boolean isFind = false;
+        for (int i = 0; i < size; i++) {
+            if (values[i] != null && ((T) values[i]).equals(element) || values[i] == element) {
```
RomanDubovskyi, line 72:
```suggestion
            if (values[i] != null && (values[i]).equals(element) || values[i] == element) {
```
Let's try to remove casting of values[i], it should work without it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+                removedValue = remove(i);
+                isFind = true;
+                break;
```
RomanDubovskyi, line 75:
```suggestion
                return remove(i);
```
and after loop just throw an exception - if loop finished, that means we don't have the same value inside our arr

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #357, opened 2021-09-19
Home work is done

romanNov-creator:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int capacity = 10;
```
RomanDubovskyi, line 6:
```suggestion
    private static final int INITIAL_CAPACITY = 10;
    private int capacity;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+    public ArrayList() {
+        this.elementData = (T[]) new Object[capacity];
+    }
+
+    public ArrayList(int arrayListCapacity) {
+        this.elementData = (T[]) new Object[arrayListCapacity];
+        capacity = arrayListCapacity;
+    }
+
+    private void increaseCapacity() {
```
RomanDubovskyi, line 20: private methods should be located after public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        if (this.size == elementData.length) {
```
RomanDubovskyi, line 21:
```suggestion
        if (size == elementData.length) {
```
`this` keyword is redundant when we have no shadowing (2 variables with same name but different scopes)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            T[] tempElementData = (T[]) new Object[capacity];
+            System.arraycopy(elementData, 0, tempElementData, 0, this.size);
+            elementData = tempElementData;
+            for (int i = 0; i < list.size(); i++) {
+                elementData[this.size + i] = list.get(i);
+            }
+            this.size += list.size();
+        } else {
+            for (int i = 0; i < list.size(); i++) {
+                elementData[this.size + i] = list.get(i);
```
RomanDubovskyi, line 61: in this case you can just use you add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            }
+            this.size += list.size();
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        for (int i = 0; i < size; i++) {
```
RomanDubovskyi, line 69: get() method should have constant algorithm complexity. check the index - if it's in bounds just `return elementData[index]` without iteration

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            if (i == index) {
+                return elementData[i];
+            }
+        }
+        throw new ArrayListIndexOutOfBoundsException(index + " is invalid index");
     }
 
     @Override
     public void set(T value, int index) {
-
```
RomanDubovskyi, line 26: same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
```
RomanDubovskyi, line 124: use ternary operator here, that will shorten the code

RomanDubovskyi (changes requested):

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #359, opened 2021-09-27
Wrote an assignment jv-array-list.

ASelivonchyk:
Override the methods of ArrayList class.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,93 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void resizeArray() {
+        if (size == array.length - 1) {
+            array = Arrays.copyOf(array, array.length + 10);
```
RomanDubovskyi, line 82: default behaviour is to increase size in 1.5 times

RomanDubovskyi (approved):

PR #360, opened 2021-10-03
the required methods for ArrayList were override

mykolaskrypka:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
+        System.arraycopy(list, index, list, index + 1, size - index);
+        list[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() > capacity) {
```
RomanDubovskyi, line 44: the same this will be done in add method, so we can skip this first loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        if (index > size - 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index to get value!");
+        }
+        return list[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index > size - 1 || index < 0) {
```
RomanDubovskyi, line 62: let's move this code to prive method checkIndex, cause it's duplicated in several methods

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #415, opened 2022-03-18
implemented task

tanya-logatskaya:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] array = new Object[INITIAL_CAPACITY];
```
Sofasmile, line 8: better do initialization in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
+    private int size = 0;
+
     @Override
     public void add(T value) {
-
+        if (array.length == size) {
+            Object[] newArray = new Object[(int) (array.length * 1.5)];
+            System.arraycopy(array, 0, newArray, 0, size);
+            array = newArray;
+        }
```
Sofasmile, line 17: this code is the same in `add(T value)` method and in the `add(T value, int index)`, so we can move it to the separate private method

Sofasmile (approved):

PR #458, opened 2022-04-16
working solution

alexgash8:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+    public ArrayList(int size) {
+        if (size > 0) {
+            this.elementData = new Object[size];
+            this.size = 0;
+        } else if (size == 0) {
+            this.elementData = EMPTY_ELEMENTDATA;
+            this.size = 0;
+        } else {
+            throw new IllegalArgumentException("Illegal Capacity: " + size);
+        }
```
MaliukDaria, line 20:
```suggestion
    public ArrayList(int size) {
        if (size > 0) {
            this.elementData = new Object[size];
        } else if (size == 0) {
            this.elementData = EMPTY_ELEMENTDATA;
        } else {
            throw new IllegalArgumentException("Illegal Capacity: " + size);
        }
```

we don't need to initialize size by 0 as it will be initialized by 0 by dafault

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+    }
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
MaliukDaria, line 25: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+        Object[] newElemData = new Object[elementData.length + 1];
+        System.arraycopy(oldElemData, 0, newElemData, 0, index);
+        newElemData[index] = value;
+        System.arraycopy(oldElemData, index, newElemData, index + 1, size - index);
+        size++;
+        elementData = newElemData;
+        return;
+    }
+
+    private void indexInRange(int index, int end) {
```
MaliukDaria, line 52: private methods should be after public as helper ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+        if (index == 0) {
+            return;
+        }
```
MaliukDaria, line 55: why do we need this method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+            grow();
+        }
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size) {
+            throw new ArrayListIndexOutOfBoundsException("index " + index + "size " + size);
+        }
```
MaliukDaria, line 40: I think we should also grow array here if it is full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+        indexInRange(index, elementData.length - 1);
+        Object[] oldElemData = elementData;
+        Object[] newElemData = new Object[elementData.length + 1];
+        System.arraycopy(oldElemData, 0, newElemData, 0, index);
+        newElemData[index] = value;
+        System.arraycopy(oldElemData, index, newElemData, index + 1, size - index);
+        size++;
+        elementData = newElemData;
+        return;
```
MaliukDaria, line 49:
```suggestion
        System.arraycopy(elementData, index, elementData, index + 1, size - index);
        elementData[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+        if (index == 0) {
+            return;
+        }
+        if (index < 0 || index > end) {
+            throw new ArrayListIndexOutOfBoundsException("index " + index + "size " + size);
+        }
+    }
 
+    private void grow() {
+        int newSize = (int) (elementData.length + (elementData.length * 0.5));
```
MaliukDaria, line 62: lets move 0.5 to constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        indexInRange(index, size - 1);
+        final T oldValue = (T) elementData[index];
+        Object[] newArr = new Object[elementData.length];
+        System.arraycopy(elementData, 0, newArr, 0, index);
+        if (index != size - 1) {
+            System.arraycopy(elementData, index + 1, newArr, index, elementData.length - index - 1);
+        }
```
MaliukDaria, line 102: same, why do we need to copy array? We could just use System.arraycopy for elementData

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
+                }
+            }
+            if (elementData[i].equals(element)) {
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("there is no such element " + element);
+        }
```
MaliukDaria, line 127: could we simplify this? We just need to go through all elements in array and check for equals

src/test/java/core/basesyntax/ArrayListTest.java
```diff
@@ -186,7 +186,7 @@ public void removeElementFromArrayListByValue() {
         Assert.assertEquals(6, arrayList.size());
         String actualResult = arrayList.remove("Java");
         Assert.assertEquals(String.format("Test failed! Returned value should be \"%s\", "
-                        + "but was \"%s\"\n", "Java", actualResult), "Java", actualResult);
+                + "but was \"%s\"\n", "Java", actualResult), "Java", actualResult);
```
MaliukDaria, line 189: we dont need to change test classes

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] EMPTY_ELEMENTDATA = {};
```
MaliukDaria, line 7: Where do you use EMPTY_ELEMENTDATA?

MaliukDaria (changes requested):
I think it's better not to copy jdk implementation... you should have made your own implementation based on basic ArrayList principles

MaliukDaria (changes requested):
test

src/main/java/core/basesyntax/ArrayList.java
alexgash8: i've read that and implemented exactly so
image

src/main/java/core/basesyntax/ArrayList.java
alexgash8: ok, added extra check before and it works, such way is much shorter and obviouse, thanks
image

src/main/java/core/basesyntax/ArrayList.java
alexgash8: agreed

src/main/java/core/basesyntax/ArrayList.java
alexgash8: yuhoooo, it works!!!

MaliukDaria (approved):

PR #463, opened 2022-04-19
1th

ihor-dmytrush:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final String INDEX_EXCEPTION_TEXT = "Index is out of bounds!";
+    private static final String NO_SUCH_ELEMENT_EXCEPTION = "Element not found in list!";
```
MaliukDaria, line 9: I dont think we need exception messages as constnats

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+
+    private T[] items = (T[]) new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
-
+        if (size == items.length) {
+            items = grow();
```
MaliukDaria, line 17:
```suggestion
            items = grow();
```

lets move `if (size == items.length) {` check to grow() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_EXCEPTION_TEXT);
+        }
```
MaliukDaria, line 47: lets move this check to separate method (as we use it several times in code)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        items[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_EXCEPTION_TEXT);
+        }
+        T[] modifyItems = (T[]) new Object[size];
```
MaliukDaria, line 64: could we make it without temp modifyItems array?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] grow() {
```
MaliukDaria, line 95: I think we should make this method void and just set new array to items in it

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #466, opened 2022-04-22
Jv-array-list-maxlut

maxlutovinov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static int CAPACITY = 10;
```
Nick97-git, line 8:
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+@SuppressWarnings("unchecked")
```
Nick97-git, line 6: do you need this here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
 public class ArrayList<T> implements List<T> {
+    private static int CAPACITY = 10;
+    private static final double CAPACITY_EXPAND_FACTOR = 1.5;
```
Nick97-git, line 9:
```suggestion
    private static final double GROW_COEFFICIENT = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
Nick97-git: you don't need a separate global variable for storing current array's length, you can get current length using method of array itself

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+        objects = new Object[CAPACITY];
+        if (list.size() <= CAPACITY) {
+            copyListToArrayListAndResize(list);
+            return;
+        }
+        do {
+            expandObjectsArray(objects);
+        } while (list.size() <= CAPACITY);
+        copyListToArrayListAndResize(list);
+    }
```
Nick97-git, line 27: the only function of constructors is an initialization of fields, you can't use such complex logic here, so remove this constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+
     @Override
     public void add(T value) {
-
+        if (size < CAPACITY) {
+            objects[size] = value;
+            size++;
```
Nick97-git, line 33: you have the same lines below, rewrite your logic, so you could remove duplicates

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+
+    private void checkIndex(int index, int correctionFactor) {
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index has to be positive");
+        } else if (index >= size + correctionFactor) {
+            throw new ArrayListIndexOutOfBoundsException("The index is out of range");
+        }
+    }
+
+    private void expandObjectsArray(Object[] objects) {
```
Nick97-git, line 129:
```suggestion
    private void resize() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index, int correctionFactor) {
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index has to be positive");
+        } else if (index >= size + correctionFactor) {
+            throw new ArrayListIndexOutOfBoundsException("The index is out of range");
+        }
```
Nick97-git, line 126: let's use one if condition with error message `Invalid index`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+    }
+
+    private void copyListToArrayListAndResize(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            objects[i] = list.get(i);
+        }
+        size = list.size();
+    }
+
+    private void copyArrayToObjectsAndResizeArrayList(
```
Nick97-git, line 143: it's a redundant method, use this lines directly in methods where you need them

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+        return (T) objects[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index, 0);
+        if (objects[index] == value) {
+            return;
+        }
```
Nick97-git, line 86: remove this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,148 @@
...
+        if (size + arrList.size <= CAPACITY) {
+            copyArrayToObjectsAndResizeArrayList(
+                    arrList.objects, 0, size, arrList.size, arrList.size);
+            return;
+        }
+        do {
+            expandObjectsArray(objects);
+        } while (size + arrList.size <= CAPACITY);
+        copyArrayToObjectsAndResizeArrayList(
+                arrList.objects, 0, size, arrList.size, arrList.size);
```
Nick97-git, line 72: use for loop to go through the passed list and add its elements using add method inside of this class

Nick97-git (changes requested):

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: When I remove the "@SuppressWarnings("unchecked")" annotation, the "(T) objects[index]" warning appears. Which practice is better - with or without this annotation?

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: Thanks, done.

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: Done

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: Done

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: Done

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: Clear

src/main/java/core/basesyntax/ArrayList.java
maxlutovinov: The best!

src/main/java/core/basesyntax/ArrayList.java
Nick97-git: This annotation should be put if you are sure that the compiler's warning can't happen. In this case, It would be better to use it at least not on the class level, but on the method level. For this task let's not use this annotation.

Nick97-git (approved):

PR #468, opened 2022-05-05
task completed

olesyakhalakk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private T[] values;
+    private int size = 0;
```
MaliukDaria, line 10:
```suggestion
    private int size;
```

It will be initialized by 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_SIZE];
+    }
+
+    public void resizeArray() {
```
MaliukDaria, line 16: lets make all helper methods private (and remember, private methods should follow after public)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        int newSize = (size + (size >> 1));
+        values = Arrays.copyOf(values, newSize);
```
MaliukDaria, line 18: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements): Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    }
+
+    public void checkIndex(int index) throws ArrayListIndexOutOfBoundsException {
```
MaliukDaria, line 21: Why do we need throws here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index doesn't exist");
```
MaliukDaria, line 23:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index " + index + " doesn't exist in list of size " + size);
```

lets make exception messages more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        }
+    }
+
     @Override
     public void add(T value) {
+        if (size == values.length) {
+            resizeArray();
```
MaliukDaria, line 30: lets move `if (size == values.length) {` check to resizeArray() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index doesn't exist");
+        }
+
+        if (size == values.length) {
+            resizeArray();
+        }
+
+        System.arraycopy(values, index, values, index + 1, size - index);
```
MaliukDaria, line 47: same, Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        T removedValue = values[index];
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        size--;
+        return removedValue;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
```
MaliukDaria, line 84: why do we need throws here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        for (int l = 0; l <= size; l++) {
+            if (Objects.equals(values[l], element)) {
```
MaliukDaria, line 86: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements): Don’t use class Objects

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+                size--;
+                System.arraycopy(values, l + 1, values, l, size - l);
+                return element;
+            }
+        }
+        throw new NoSuchElementException("There is no valid value in array");
```
MaliukDaria, line 92:
```suggestion
        throw new NoSuchElementException("There is no such value in array " + element);
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: what about this comment?

PR #479, opened 2022-05-20
implement task

gdpbalt:
- add implementation of interface List in class ArrayList;
- data stores in array Object[] elementData;
- field size has information about size;
- add implementation of public methods in class ArrayList: add(), addAll(), get(), set(), remove(), size(), isEmpty()

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,133 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int GROW_BY = 2;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            elementData = grow();
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        rangeCheckForAdd(index);
+        if (size == elementData.length) {
+            elementData = grow();
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
+                s.append(", ");
+            }
+            s.append(elementData[i]);
+        }
+        return String.format("[%s] (%d)", s, size());
+    }
+
+    private Object[] grow() {
+        int newCapacity = elementData.length * GROW_BY;
+        return Arrays.copyOf(elementData, newCapacity);
```
liliia-ponomarenko, line 102: Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
+    }
+
+    private void fastRemove(Object[] data, int index) {
```
liliia-ponomarenko, line 105: Does this method make fast removal? Let's create another name

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    @Override
+    public String toString() {
```
liliia-ponomarenko, line 89: this method is unnecessary, let’s  remove it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
+        for (int i = 0; i < size; i++) {
+            if (i != 0) {
+                s.append(", ");
+            }
+            s.append(elementData[i]);
+        }
+        return String.format("[%s] (%d)", s, size());
+    }
+
+    private Object[] grow() {
```
liliia-ponomarenko, line 100: this method can be void

liliia-ponomarenko (changes requested):
Good job 👍  Let’s improve your solution ;)

PR #486, opened 2022-05-24
Class ArrayList was created.

Ivan-Yuriv:
All methods were Override.
Also were added methods grow() and checkIndex();

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAYLIST_CAPACITY = 10;
+    private int size = 0;
+    private T[] elements;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_ARRAYLIST_CAPACITY];
+    }
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
+
```
liliia-ponomarenko, line 91: Let’s  remove  empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null && elements[i] == null || elements[i] != null
+                    && elements[i].equals(element)) {
+                return remove(i);
+            }
+
```
liliia-ponomarenko, line 72: the same here (check it in your solution)

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 😊

PR #492, opened 2022-05-27
All tests were passed. CheckStyle was checked.

MKaramavrov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+        increaseSize();
+        newArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        if (size == newArray.length) {
```
zim10093, line 20: in increaseSize() you have the same check, you can check only in increaseSize method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+            increaseSize();
+        }
+        if (index < size && index >= 0) {
```
zim10093, line 23: you have repeated code in get() and remove() can you move to a separate method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+            newArray[index] = value;
+        } else {
+            getArrayException(index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        increaseSize();
```
zim10093, line 33: redundant check, you make same in the add()

zim10093 (changes requested):
you have repeated  line"return getArrayException(index)".
You can check and can throw new exceptions in one method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -4,7 +4,7 @@
 
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
-    private T[] newArray = (T[]) new Object[DEFAULT_CAPACITY];
+    private Object[] newArray = (T[]) new Object[DEFAULT_CAPACITY];
```
zim10093, line 7: you don`t need cast to T type. store Object in Object

zim10093 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,89 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] newArray = new Object[DEFAULT_CAPACITY];
```
KostyaNikiforov, line 7:
```suggestion
    private Object[] values = new Object[DEFAULT_CAPACITY];
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #508, opened 2022-06-02
added solution

DooMLeoN:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Out of bout exception");
+        }
+        grow();
+        System.arraycopy(objects,index,objects,index + 1,objects.length - index - 1);
+        objects[index] = value;
+        size++;
+    }
 
+    private void grow() {
```
MaliukDaria, line 33: private methods should follow after public as helper ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        int size = this.size;
+        int listIndex = 0;
+        for (int j = size; j < list.size() + size; j++) {
+            objects[j] = list.get(listIndex);
+            listIndex++;
+            this.size++;
+        }
```
MaliukDaria, line 51: what about grow method? What if we need to increase the initial array to add the whole list? Maybe we can just use our `add(T value)`  method for every element in the incoming list?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        System.arraycopy(objects,index + 1, objects,index, objects.length - index - 1);
+        size--;
+        return element;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T value) {
+        T[] arrayContainer = objects;
```
MaliukDaria, line 84: can we make it without creating additional array(arrayContainer)?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        T element = objects[0];
+        for (int j = 0; j < size(); j++) {
+            if (Objects.equals(objects[j], value)) {
+                element = objects[j];
+                System.arraycopy(arrayContainer,j + 1, objects,j, objects.length - j - 1);
+                break;
+            }
+            if (j == size() - 1) {
+                throw new NoSuchElementException();
+            }
```
MaliukDaria, line 94: why do we need this check?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int BASED_ARRAY_SIZE = 10;
```
```suggestion
    private static final int INITIAL_CAPASITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        grow();
+        System.arraycopy(objects,index,objects,index + 1,objects.length - index - 1);
+        objects[index] = value;
+        size++;
+    }
 
+    private void grow() {
+        if (size == objects.length) {
+            T[] arraySours;
+            arraySours = objects;
```
MaliukDaria, line 36:
```suggestion
            T[] arraySours = objects;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
-        return null;
+    public T remove(T value) {
+        T element;
+        for (int j = 0; j < size(); j++) {
+            if (objects[j] == value || objects[j] != null && objects[j].equals(value)) {
+                element = objects[j];
+                System.arraycopy(objects,j + 1, objects,j, objects.length - j - 1);
+                size--;
+                return element;
+            }
```
MaliukDaria, line 70:
```suggestion
            if (objects[j] == value || objects[j] != null && objects[j].equals(value)) {
               return  remove(j);
            }
```

we can reuse our `remove(int index)` method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size == objects.length) {
+            T[] arraySours = objects;
+            objects = (T[])new Object[arraySours.length + 5];
+            System.arraycopy(arraySours,0,objects,0,objects.length - 5);
```
MaliukDaria, line 89: Why +5? We need to increase capacity 1.5 times

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
```
MaliukDaria, line 90: redundant empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        }
     }
+
+    private void checkOutOfBoundsException(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Out of bout exception for index: "
+                    + index);
+        }
+    }
+
```
MaliukDaria, line 100: redundant empty line

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #510, opened 2022-06-04
ArrayList implements List implemented

Illinois357:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+    private static final int ADD_ID = 1;
+    private static final int REMOVE_ID = 0;
+    private T[] storage;
+    private int size;
+    private int currentCapacity;
+
+    {
+        storage = (T[]) new Object[INITIAL_SIZE];
+        currentCapacity = storage.length;
+    }
```
Ivan-Yuriv, line 16: You can wrap these two lines into
public ArrayList() {
} for example. To make this part of code more readable.

Ivan-Yuriv (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        if (operation == ADD_ID) {
+            System.arraycopy(storage, index, storage, index + 1, size - index);
+        }
+        if (operation == REMOVE_ID) {
+            System.arraycopy(storage, index + 1, storage, index, size - index - 1);
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index > size || index < 0) {
```
liliia-ponomarenko, line 123:
```suggestion
        if (index >= size || index < 0) {
```
if you write this conditional and you can  use this method in 3 cases (without duplicate cal error when index == size)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void resize() {
+        T[] temporary = (T[]) new Object[currentCapacity];
+        System.arraycopy(storage, 0, temporary, 0, size);
+        storage = (T[]) new Object[(int) (currentCapacity * 1.5)];
+        System.arraycopy(temporary, 0, storage, 0, currentCapacity);
+        currentCapacity = (int) (currentCapacity * 1.5);
```
liliia-ponomarenko, line 110: Can we copy the array only once?

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

PR #512, opened 2022-06-08
Hw array list

MarkGodnyi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[10];
```
Nick97-git, line 11: create a constant from 10

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        ensureCapacity(size + 1);
+        try {
+            System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        } catch (IndexOutOfBoundsException e) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element to this index");
+        }
```
Nick97-git, line 27: instead of try catch better check index before you try to execute arraycopy, the same for other similar cases

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(size + list.size());
+        System.arraycopy(toArray(list), 0, elementData, size, list.size());
```
Nick97-git, line 35: you need just go through all elements of the list using for loop and call add method inside

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return elementData[0] == null;
```
Nick97-git, line 88:
```suggestion
        return size == 0;
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int LIST_CAPACITY = 10;
```
Nick97-git, line 7:
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Can't replace the element on the specified position");
+        }
```
Nick97-git, line 52: you have the same logic in several places, let's create a separate method and move this logic there

Nick97-git (changes requested):

Nick97-git (changes requested):
I mean to create a method only for this block of code
Screenshot 2022-06-09 at 22 04 25

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    }
+
+    private void removeElement(int index) {
+        int numMoved = size - index - 1;
+        System.arraycopy(elementData, index + 1, elementData, index, numMoved);
+        elementData[--size] = null;
+    }
+
+    private void throwArrayListIndexOutOfBoundsException(int index) {
```
Nick97-git, line 96:
```suggestion
    private void checkIndex(int index) {
```

Nick97-git (changes requested):

PR #513, opened 2022-06-11
added hw10-solution

ArinaSaara:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+
```
MaliukDaria, line 8: redundant empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elements;
+    private int size = 0;
```
MaliukDaria, line 11:
```suggestion
    private int size;
```

it will be initialized by 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resizeArray() {
+        int newSize = elements.length + (elements.length >> 1);
+        elements = Arrays.copyOf(elements, newSize);
```
MaliukDaria, line 81: Why do we need to use Arrays.copyOf here?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
ArinaSaara: Fixed

src/main/java/core/basesyntax/ArrayList.java
ArinaSaara: I wanted to avoid creating another array with a larger length and redefining the old one. Is it possible to do this at all or is it a duplication?

MaliukDaria (approved):

PR #529, opened 2022-07-24
Please check my solution for task jv-array-list

andmalenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == list.length) {
+            list = increase(list);
+        }
+        if ((index >= 0) && (index <= size)) {
```
KostyaNikiforov, line 28: Let's remove redundant brackets

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            Object[] temporary = new Object[list.length];
+            System.arraycopy(list, 0, temporary,0, index);
+            temporary[index] = value;
+            System.arraycopy(list, index, temporary,index + 1, size - index);
+            list = temporary;
```
KostyaNikiforov, line 33: Let's try to use `arraycopy` method only once, you can avoid using temporary array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            size++;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Your index is over the size");
```
KostyaNikiforov, line 36: Le's make exception messages more informative. It is a good practice to add info about the size and the wrong index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        if ((index >= 0) && (index < size)) {
+            list[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Your index is over the size");
+        }
     }
```
KostyaNikiforov, line 63: In both of these methods, you have similar logic. Let's move it to the separate private one

src/main/java/core/basesyntax/ArrayList.java
KostyaNikiforov, line 28: also you can try to reuse that method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        if ((index >= 0) && (index < size)) {
```
KostyaNikiforov, line 67: and here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                System.arraycopy(list, i + 1, list, i, size - i);
+                size--;
+                isFound = true;
+                break;
+            }
+        }
+        if (!isFound) {
+            throw new NoSuchElementException("The element being requested does not exist");
+        }
+        return element;
```
KostyaNikiforov, line 95: try to avoid using isFound variable and break operator

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
andmalenko: I've done it by explanation of Bogdan Chupika:
https://youtu.be/YhSF8_hVIU4?t=74
As I understand, you want that I changed 29-33 to:
System.arraycopy(list, index, list,index + 1, size - index);
list[index] = value;
Right?

src/main/java/core/basesyntax/ArrayList.java
andmalenko: I moved this logic to the new method - isValid(index).
Please check is it correct to use it like this?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((list[i] == element) || (list[i] != null) && (list[i].equals(element))) {
+                System.arraycopy(list, i + 1, list, i, size - i);
+                size--;
+                return element;
+            }
+        }
+        throw new NoSuchElementException("The element being requested does not exist");
```
MaliukDaria, line 77:
```suggestion
        throw new NoSuchElementException("The element being requested does not exist " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size == 0;
+    }
+
+    private Object[] increase(Object[] list) {
+        Object[] increasedList = new Object[list.length + (list.length >> 1)];
+        System.arraycopy(list, 0, increasedList, 0, list.length);
+        return increasedList;
+    }
+
+    private void isValid(int index) {
```
MaliukDaria, line 96:
```suggestion
    private void checkIndex(int index) {
```

MaliukDaria (approved):

PR #530, opened 2022-07-25
Solution

VitaliiShkliarskyi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double SIZE_INCREASE_FACTOR = 1.5;
+    private Object[] elementData; // should we use transient keyword?
```
MaliukDaria, line 9: | should we use transient keyword?
for what purpose?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Object[] buildUp() {
```
MaliukDaria, line 93:
```suggestion
    private void resize() {
```

could we make this method void?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        if (index == size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of range: " + index);
+        }
+    }
+
+    private void checkList(List<T> list) {
+        if (list == null) {
+            throw new NullPointerException("List : " + list + " can't be null");
+        }
+    }
```
MaliukDaria, line 117: Why do we need this check?

src/main/java/core/basesyntax/ArrayList.java
KostyaNikiforov, line 9: Array names must be plural

KostyaNikiforov, line 9: For example
```suggestion
    private Object[] elements; // should we use transient keyword?
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        checkIndexEqualsSize(index);
```
MaliukDaria, line 48: could we combine these checks into one method?

KostyaNikiforov (changes requested):

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ArrayList.java
VitaliiShkliarskyi: If the specified collection is null

src/main/java/core/basesyntax/ArrayList.java
VitaliiShkliarskyi: For serialization. I don't know for sure, but original ArrayList uses it. That's why I'm asking

VitaliiShkliarskyi: Isn't "data" word plural?

src/main/java/core/basesyntax/ArrayList.java
VitaliiShkliarskyi: We can do this, but it will require extra conditions in some methods.

src/main/java/core/basesyntax/ArrayList.java
MaliukDaria: I dont think we need this check for this task, and even if we provide it, we should throw some RuntineException, not NPE

MaliukDaria (approved):

PR #534, opened 2022-07-27
ArrayList_ver1

Maksym-Kulakov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_INITIAL_ARRAY_CAPACITY = 10;
+    private Object[] values;
+    private int size;
+    private int capacityOfArray;
```
MaliukDaria, line 9: why do we need `capacityOfArray` variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        }
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
```
MaliukDaria, line 28:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index  "+index+"is invalid for size " + size);
```

lets make exception messages more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+            size++;
+        } else if (index < size()) {
+            Object[] newValuesOne = new Object[capacityOfArray];
+            Object[] newValuesTwo = new Object[capacityOfArray];
+            System.arraycopy(values, 0, newValuesOne, 0, index);
+            System.arraycopy(values, index, newValuesTwo, 0, capacityOfArray - index);
+            values = newValuesOne;
+            values[index] = value;
+            System.arraycopy(newValuesTwo, 0, values, index + 1, capacityOfArray - index - 1);
+            size++;
```
MaliukDaria, line 44: could we use `System.arraycopy` method only ones here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        System.arraycopy(valuesFmArrayList, 0, values, size(), list.size());
+        size = size + list.size();
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
+        }
```
MaliukDaria, line 65: I think we can move this logic to separate private method(since it is repeated several times in the code)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
+        }
+        final T removedValue = (T) values[index];
+        Object[] newValuesOne = new Object[capacityOfArray];
+        System.arraycopy(values, index + 1, newValuesOne, 0, capacityOfArray - index - 1);
+        System.arraycopy(newValuesOne, 0, values, index, capacityOfArray - index - 1);
```
MaliukDaria, line 85: could we use `System.arraycopy` only ones here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int index = 0; index < size; index++) {
+            if ((values[index] == element) || (values[index] != null
+                    && values[index].equals(element))) {
+                return remove(index);
+            }
+        }
+        throw new NoSuchElementException("Element was not found");
```
MaliukDaria, line 98:
```suggestion
        throw new NoSuchElementException("Element was not found " + element);
```

MaliukDaria (changes requested):
The task should not be implemented in the master/main branch, a separate branch should be created for this

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: changed with value.length

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: msg expanded

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: checkIndex method created

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: I think I should)

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: added as well

src/main/java/core/basesyntax/ArrayList.java
Maksym-Kulakov: one remained only

Maksym-Kulakov:
Thanks for yr good corrections!  And I think I have forgot to push "fork". Should I make fork for this task?

MaliukDaria (approved):
👍🏻

PR #537, opened 2022-07-30
Hw array list

dmytrovb02:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_OF_SIZE = 10;
```
Nick97-git, line 6:
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Object[] increaseInSize(Object[] elements) {
```
Nick97-git, line 102:
```suggestion
    private void increaseSize(Object[] elements) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isIndexCorrect(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cannot remove element. "
+                    + "Index " + index + " is not correct");
+        }
```
Nick97-git, line 71: create a method for this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size - 1; i++) {
+            if (elements[i] == element || (elements[i] != null && elements[i].equals(element))) {
+                System.arraycopy(elements, i + 1, elements, i, size - 1);
+                size--;
+                return element;
```
Nick97-git, line 86:
```suggestion
                return remove(i);
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        Object[] increaseSizeLine = new Object[elements.length + (elements.length / 2)];
+        System.arraycopy(elements, 0, increaseSizeLine, 0, elements.length);
+        elements = increaseSizeLine;
+    }
+
+    private boolean isIndexCorrect(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is not correct");
+        }
         return false;
```
aaananas, line 95:
```suggestion
    private boolean isIndexCorrect(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is not correct");
        }
        return true;
```
This logic is not breaking anything because you don't use value returned from this method
The method throws an exception if the index is not correct, but in other cases (index is correct) the method should return true because the index is correct
Or you can just change the name of method to checkIndex and don't return any value

aaananas (approved):

PR #539, opened 2022-08-01
created solution

semwithuncle:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
dSkarzh, line 37: use `foreach` here instead of `fori`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+
+    private void acceptIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Wrong index out of bounds, please check index!");
+        }
+    }
+
+    private void increaseArray() {
+        if (size == values.length) {
```
dSkarzh, line 91: move this check out of this method, it would be better if this method only increased capacity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void acceptIndex(int index) {
```
dSkarzh, line 83:
```suggestion
    private void validateIndex(int index) {
```

dSkarzh (changes requested):
Overall great job!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        }
+        System.arraycopy(values, index, values, index + 1, size - index);
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        IntStream.range(0, list.size()).mapToObj(list::get).forEach(this::add);
```
dSkarzh, line 39:
```suggestion
        for (T element : list) {
            add(element);
        }
```
i meant something like this :)

dSkarzh (approved):

PR #543, opened 2022-08-02
create own ArrayList, implement behavior, throw exceptions

verbanov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,160 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private int size;
+    private Node<T>[] items;
+
+    public ArrayList(T value) {
```
Oleksandr-Tovstyi, line 10: this constructor is not needed

Oleksandr-Tovstyi, line 10: this constructor is not needed

Oleksandr-Tovstyi, line 10:
```suggestion
    public ArrayList(T value) {
```
this constructor is not needed

Oleksandr-Tovstyi (changes requested):

src/main/java/core/basesyntax/ArrayList.java
verbanov: Agree

Oleksandr-Tovstyi (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private int size;
+    private Node<T>[] items;
```
Sofasmile, line 8: should we use Node for array list? can we save element directly to the array?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+
+    public ArrayList() {
+        items = new Node[MAX_ARRAY_LENGTH];
+        size = 0;
```
Sofasmile, line 12: 0 is default value for int, let's remove initialization

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
     public void add(T value) {
-
+        if (size < items.length) {
+            items[size] = new Node<>(value);
+            size++;
+        } else {
+            growArray(size);
+            items[size] = new Node<>(value);
+            size++;
+        }
```
Sofasmile, line 24: try to remove code duplication from if-else block

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index");
+        } else if (index == size) {
```
Sofasmile, line 31: should we have `else` block after `throw` in the `if` statement?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+            add(value);
+        } else {
+            if (size + 1 <= items.length) {
+                addInside(index, value);
+            } else {
+                growArray(size);
+                addInside(index, value);
+            }
```
Sofasmile, line 39: same here, we do the same (addInside(index, value)) in two blocks, try to refactor this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
+        if (getIndexInRange(index)) {
+            return items[index].getValue();
+        }
         return null;
```
Sofasmile, line 55: it's optional to have `if` statement, because in method getIndexInRange() you throw exception, so it should work without `if` statement
```suggestion
        getIndexInRange(index);
        return items[index].getValue();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (getIndexInRange(index)) {
```
Sofasmile, line 60: same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+            items[index] = new Node<>(value);
+        }
     }
 
     @Override
     public T remove(int index) {
+        if (getIndexInRange(index)) {
```
Sofasmile, line 67: and here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+            tempArray[i] = items[i];
+        }
+        items = new Node[newArrayLength];
+        for (int i = 0; i < countOfElements; i++) {
+            items[i] = tempArray[i];
+        }
+    }
+
+    private int getNewLength(int oldLength) {
+        if (oldLength == 1) {
```
Sofasmile, line 117: int this implementation can oldLength be 1?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+    private void growArray(int countOfElements) {
+        int newArrayLength = getNewLength(countOfElements);
+        Node<T> [] tempArray = new Node[newArrayLength];
+        for (int i = 0; i < countOfElements; i++) {
+            tempArray[i] = items[i];
+        }
+        items = new Node[newArrayLength];
+        for (int i = 0; i < countOfElements; i++) {
+            items[i] = tempArray[i];
+        }
```
Sofasmile, line 113: you can use System.arrayCopy() for optimization

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,157 @@
...
+
+    private void addInside(int index, T value) {
+        Node<T>[] tempArray = new Node[items.length - index - 1];
+        for (int i = 0; i < tempArray.length; i++) {
+            tempArray[i] = items[index + i];
+        }
+        items[index] = new Node<>(value);
+        for (int i = 0; i < tempArray.length; i++) {
+            items[index + 1 + i] = tempArray[i];
+        }
```
Sofasmile, line 131: in this case you can use System.arraycopy too

Sofasmile (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        items[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index");
```
Noct2000, line 26: add index to error message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        items[index] = value;
+        for (int i = 0; i < tempArray.length; i++) {
+            items[index + 1 + i] = tempArray[i];
+        }
+        size++;
+    }
+
+    private boolean getIndexInRange(int index) {
+        if (index < 0 || index >= size()) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index");
```
Noct2000, line 115: add index to error message

Noct2000 (changes requested):
pay attention to old comments

Noct2000 (approved):

PR #547, opened 2022-08-04
Hw solution is developed

amarkelovUA:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        if (incorrectIndex) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR);
+        }
+    }
+
+    private void addArray(Object[] temp) {
+        for (Object element : temp) {
+            add((T) element);
+        }
     }
```
dSkarzh, line 112: you need to use `System.arraycopy()` instead of this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private void increaseCapasity() {
+        capacity *= GROWTH_FACTOR;
+        values = Arrays.copyOf(values, capacity);
+    }
+
+    private void checkIndex(boolean incorrectIndex) throws ArrayListIndexOutOfBoundsException {
+        if (incorrectIndex) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR);
+        }
+    }
```
dSkarzh, line 106: move whole logic of index checking and exception throwing in or out of this method
do not split it in half

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void increaseCapasity() {
+        capacity *= GROWTH_FACTOR;
+        values = Arrays.copyOf(values, capacity);
+    }
```
dSkarzh, line 100: `System.arraycopy()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                remove(i);
+                isExist = true;
+                break;
+            }
+        }
+        if (!isExist) {
+            throw new NoSuchElementException(MISSING_ELEMENT);
+        }
+        return element;
     }
```
dSkarzh, line 85:
```suggestion
    public T remove(T element) throws NoSuchElementException {
        for (int i = 0; i < size; i++) {
            if (element == values[i] || element != null && element.equals(values[i])) {
                return remove(i);
            }
        }
        throw new NoSuchElementException(MISSING_ELEMENT);
    }
```

dSkarzh: can be simplified

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index >= size || index < 0);
+        T removedElement = (T) values[index];
+        Object[] temp = Arrays.copyOfRange(values, index + 1, size);
+        size = index;
+        addArray(temp);
```
dSkarzh, line 67: `System.arraycopy()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        checkIndex(index > size || index < 0);
+        if (index == size) {
+            add(value);
+        } else {
+            Object[] temp = Arrays.copyOfRange(values, index, size);
+            values[index] = value;
+            size = index + 1;
+            addArray(temp);
+        }
```
dSkarzh, line 39: `System.arraycopy()`

dSkarzh (changes requested):

dSkarzh (approved):

PR #567, opened 2022-08-08
relize ArrayList class

Toltbcm:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] values;
```
justpopova, line 7: i think here is better to define as Object
```suggestion
    private Object[] values;
```
i think here is better to define as Object, so in constructor we won't need to cast an array.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+
```
justpopova, line 8: remove empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+            T[] valuesTemp = (T[]) new Object[size - index + 1];
+            for (int i = index; i <= size; i++) {
+                valuesTemp[i - index] = values[i];
+            }
+            values[index] = value;
+            for (int i = index + 1; i <= size; i++) {
+                values[i] = valuesTemp[i - index - 1];
+            }
+            size++;
+        }
```
justpopova, line 38:
```suggestion
        } else {
            increase();
            System.arraycopy(values, index, values, index + 1, size - index);
            values[index] = value;
            size++;
        }
```
better use System.arrayCopy, code will be smaller and no need to use two loops for that , i think this solution needs to work with your code(i'm still new so maybe did some mistake:( )

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void increase() {
+        if (size >= values.length) {
+            T[] valuesTemp = (T[]) new Object[size * 3 / 2];
```
justpopova, line 101:
```suggestion
            Object[] valuesTemp = new Object[size * 3 / 2];
```
if we change type to Object, than here also need change, so don't need to cast

justpopova (changes requested):
few changes

src/main/java/core/basesyntax/ArrayList.java
Toltbcm: I'm not sure about it, but I'll do it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+            for (int i = 0; i < size; i++) {
+                valuesTemp[i] = values[i];
+            }
+            values = valuesTemp;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("To big index");
```
justpopova, line 111: add index number to exception message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (index != 0 && (index > size || index < 0)) {
+            throw new ArrayListIndexOutOfBoundsException("To big index");
```
justpopova, line 117: add index number to exception message

src/main/java/core/basesyntax/ArrayList.java
Toltbcm: I knew about this method, but as usual, I forget to read "Try to avoid these common mistakes while solving task". I thought it was better. )

justpopova (changes requested):
add index

src/main/java/core/basesyntax/ArrayList.java
justpopova: also maybe better if in message written "index " + index + " out of bounds"

justpopova (approved):

justpopova (approved):

justpopova (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Too big index " + index);
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (index != 0 && (index > size || index < 0)) {
```
Sofasmile, line 106:
```suggestion
        if (index > size || index < 0) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        checkAddIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
+            increase();
+            size++;
```
Sofasmile, line 28: better move `size++` after `values[index] = value;` operation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+            Object[] valuesTemp = new Object[size - index];
+            System.arraycopy(values, index, valuesTemp, 0, valuesTemp.length);
+            values[index] = value;
+            System.arraycopy(valuesTemp, 0, values, index + 1, valuesTemp.length - 1);
```
Sofasmile, line 32: you can copy elements in the same array
```suggestion
            System.arraycopy(values, index, values, index + 1, size - index);
            values[index] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Object oldValue = values[index];
+        size--;
+        System.arraycopy(values, index + 1, values, index, size - index);
```
Sofasmile, line 60:
```suggestion
        System.arraycopy(values, index + 1, values, index, size - index - 1);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return (T) oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (values[i] == element || values[i] != null && values[i].equals(element)) {
+                index = i;
```
Sofasmile, line 69: let's call method `remove(int index)` in `if` statement

Sofasmile (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("There is no such element " + element);
+        }
+        return remove(index);
```
Sofasmile, line 74: same but fewer lines :)
```suggestion
        for (int i = 0; i < size; i++) {
            if (values[i] == element || values[i] != null && values[i].equals(element)) {
                return remove(index);
            }
        }
        throw new NoSuchElementException("There is no such element " + element);
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #579, opened 2022-08-08
implemented own ArrayList class methods

ArturZhadan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
```
justpopova, line 7:
```suggestion
    private Object[] arrayList;
```
better to do Object[], so then you dont need to cast

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private int size;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[DEFAULT_CAPACITY];
```
justpopova, line 11:
```suggestion
        arrayList = new Object[DEFAULT_CAPACITY];
```
if change type to Object than no need to cast it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        arrayList[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index low than 0 or bigger than array size");
```
justpopova, line 27:
```suggestion
                    "Index " + index + " low than 0 or bigger than array size");
```
add index to exception message, so can understand which index caused problem.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index low than 0 or bigger than array size");
```
justpopova, line 48:
```suggestion
                    "Index " + index + " low than 0 or bigger than array size");
```
also add index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
+        return (T) arrayList[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index low than 0 or bigger than array size");
```
justpopova, line 57:
```suggestion
                    "Index " + index + " low than 0 or bigger than array size");
```
also add index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index low than 0 or bigger than array size");
```
justpopova, line 66:
```suggestion
                    "Index " + index + " low than 0 or bigger than array size");
```
also add index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] arrayListGrow() {
+        T[] newArrayList = (T[]) new Object[(int)((arrayList.length + (arrayList.length * 2)) / 2)];
```
justpopova, line 96:
```suggestion
        Object[] newArrayList = new Object[(int)(size + size * 1.5)];
        OR
        Object[] newArrayList = new Object[(int)(arrayList.length + arrayList.length * 1.5)];
```
Alos aftrer changing the type of array, no need to cast, and maybe you can change capacity increase to this kind of solution, up to you ^_^

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        System.arraycopy(arrayList, 0, newArrayList, 0, arrayList.length);
+        arrayList = (T[]) newArrayList;
```
justpopova, line 98:
```suggestion
        arrayList =  newArrayList;
```
here also won't need to cast anymore i think.

justpopova (changes requested):
good job!

src/main/java/core/basesyntax/ArrayList.java
ArturZhadan: if we use 1.5, then we need to add this to private fields as constant, that's why I choose just formula)))

justpopova: Okay, great! :)

justpopova (approved):
Great job!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] arrayList;
```
Sofasmile, line 7:
```suggestion
    private Object[] data/elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Object[] arrayListGrow() {
```
Sofasmile, line 95: method should begin with verb

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index " + index + " low than 0 or bigger than array size");
+        }
```
Sofasmile, line 49: this check is the same as in the method set() or remove(), so let's create a separate private method for it and reuse

Sofasmile (changes requested):

Sofasmile (changes requested):

Sofasmile (approved):

PR #588, opened 2022-08-09
It's working! Still don`t believe in God?

Aptypio13:
I try to use simple methods to implement it. Magic number at 99:90 when i resize a array, I really don`t know have to replace

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] dataArray;
+    private int size;
+
+    public ArrayList() {
+        dataArray = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growUp();
+        dataArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        growUp();
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index < size && index >= 0) {
+            System.arraycopy(dataArray, index, dataArray, index + 1, size - index);
+            dataArray[index] = value;
+            size++;
+            return;
+        }
+        throw new ArrayListIndexOutOfBoundsException(
+                "Index " + index + " is wrong for size " + size);
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list.size() > size) {
+            growUp();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            dataArray[size] = list.get(i);
+            size++;
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        validationForIndex(index);
+        return (T)dataArray[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        validationForIndex(index);
+        growUp();
+        if (index == size) {
+            add(value);
+        }
+        dataArray[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validationForIndex(index);
+        Object elemetRemove = dataArray[index];
+        System.arraycopy(dataArray, index + 1, dataArray, index, size - index - 1);
+        size--;
+        return (T) elemetRemove;
     }
 
     @Override
     public T remove(T element) {
-        return null;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] dataArray;
```
justpopova, line 7:
```suggestion
    private Object[] data/elements;
```
maybe better change the name? all up to you ^_^

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        if (list.size() > size) {
+            growUp();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            dataArray[size] = list.get(i);
+            size++;
+        }
```
justpopova, line 46: here maybe better create just for each loop with adding elements? and than create in add() your condition that :
if (list.size() > size) {
            growUp();
        }
 other away i think that every time, after each adding of element to your list, you are widening your capacity(maybe i wrong, but i think that in this method you can remove this condition and move it to add(T value))^_^

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
-        return null;
+        validationForIndex(index);
+        return (T)dataArray[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        validationForIndex(index);
+        growUp();
```
justpopova, line 58: not sure if here need that method, cause here we are just setting new value to existing, and if our index is out of bounds than this method won't work anyway. nut if it fits our size, we don't need to grow it more.(maybe this code works with your solutions, check it, or maybe you don't need it here) ^_^

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            add(value);
+        }
+        dataArray[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validationForIndex(index);
+        Object elemetRemove = dataArray[index];
```
justpopova, line 68:
```suggestion
        Object removedElement = dataArray[index];
```
better change to this name

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
-        return null;
+        for (int i = 0; i < dataArray.length; i++) {
+            if (element != null
+                    && element.equals(dataArray[i])
+                    || element == dataArray[i]) {
+                return remove(i);
+
+            }
+        }
+        throw new NoSuchElementException();
```
justpopova, line 84:
```suggestion
        throw new NoSuchElementException("No such element in the list.");
```
add message

justpopova (changes requested):
Good job!

justpopova (approved):
🔥

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementsArray;
```
Sofasmile, line 7:
```suggestion
    private Object[] elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        growUp();
+        if (index == size) {
+            add(value);
+            return;
+        }
```
Sofasmile, line 27: firstly we should check index, and then call growUp() and add element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void growUp() {
```
Sofasmile, line 92: should this method be public?

Sofasmile (changes requested):

Sofasmile (approved):

PR #607, opened 2022-08-31
java.lang.ArrayIndexOutOfBoundsException: Index 10 out of bounds for …

oleksandr-kravchenko:
…length 10

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_ITEMS_NUMBER = 10;
+    private static final double INCREASE_RATE = 1.5;
+    private T[] item;
+    private static int size;
+
+    public ArrayList() {
+        item = (T[]) new Object[START_ITEMS_NUMBER];
+    }
+
     @Override
     public void add(T value) {
+        if (item.length == size) {
+            grow();
+        }
+        item[size] = value;
+        size++;
 
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
+            if (item.length == size) {
+                grow();
+            }
+            System.arraycopy(item, index, item, index + 1, size - index);
+            item[index] = value;
+            size++;
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    private int size;
+
+    public ArrayList() {
+        item = (T[]) new Object[START_ITEMS_NUMBER];
+    }
+
     @Override
     public void add(T value) {
-
+        grow();
```
```suggestion
        if (size == items.length) {
            grow();
        }
```
Better check if you actually need to increase the capacity, because the name of the method suggests that it will be increased without any checks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+        } else {
+            checkIndex(index);
+            grow();
```
aaananas, line 28: check here too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || (index >= size)) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
+
+    private int findValue(T element) {
+        for (int i = 0; i < size; i++) {
+            if (item[i] != null && item[i].equals(element) || item[i] == element) {
```
aaananas, line 95:
```suggestion
            if (item[i] == element || item[i] != null && item[i].equals(element)) {
```

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
oleksandr-kravchenko: I moved the check to the method itself - so as not to duplicate the code!

src/main/java/core/basesyntax/ArrayList.java
oleksandr-kravchenko: I moved the check to the method itself - so as not to duplicate the code!

src/main/java/core/basesyntax/ArrayList.java
oleksandr-kravchenko: yes, that's my mistake

aaananas (approved):

PR #627, opened 2022-09-26
class ArrayList was implemented, required methods were overridden.

kateryna-mykh:


aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+            add(value, index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (list.size() > (array.length - currentSize)) {
+            resizeArray();
+        }
```
kateryna-deshchinska, line 43: we don't need that, because we have resizeArray inside of add(T element) method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        checkIndex(index);
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedElement = array[index];
+        int newSize = currentSize - 1;
```
kateryna-deshchinska, line 65: try to avoid redudant variable, you may do it inside System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double GROWTH_INDEX = 1.5;
+    private static final int DEFAULT_CAPACITY = 10;
+    private int currentSize;
```
kateryna-deshchinska, line 8: we can name it "size", it will be easy to read)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= currentSize) {
+            throw new ArrayListIndexOutOfBoundsException("index [" + index + "] is invalid");
+        }
+    }
+
+    private boolean checkSize() {
+        return currentSize < array.length;
+    }
```
kateryna-deshchinska, line 99: lets avoid creating redundant methods, we use it only twice, and it is short

zachar-chorny (approved):
Good job. But for the future, don't change anything in the tests, even if it's just line breaks, it's hard to tell right away that you haven't changed any of the checks

kateryna-mykh:
> Good job. But for the future, don't change anything in the tests, even if it's just line breaks, it's hard to tell right away that you haven't changed any of the checks

I committed the wrong file, it should be ArrayList.java

zachar-chorny (approved):
yes)

PR #628, opened 2022-09-28
realized methods in arraylist

Nikostrik:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private T[] values;
+    private int size;
+
+    public ArrayList(int initCapacity) {
+        if (initCapacity < 0) {
+            throw new ArrayListIndexOutOfBoundsException("this index is invalid");
+        }
+        values = (T[]) new Object[MAX_ARRAY_LENGTH];
+    }
```
kateryna-deshchinska, line 15: why are you doing that?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+
+    public ArrayList() {
+        this(MAX_ARRAY_LENGTH);
```
kateryna-deshchinska, line 18: You need to initialize variable "values" inside constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+            System.arraycopy(values, 0, newArray, 0, size);
+            values = newArray;
+        }
+    }
+
+    private void checkIndex(int index, int size) throws ArrayListIndexOutOfBoundsException {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index is incorrect: " + index);
+        }
+    }
```
kateryna-deshchinska, line 34: private method need to be after public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    private void resizeArray() {
+        if (size == values.length) {
+            int sizeNewArray = (int) (values.length * 1.5);
+            T[] newArray = (T[]) new Object[(int) (sizeNewArray)];
+            System.arraycopy(values, 0, newArray, 0, size);
+            values = newArray;
+        }
+    }
+
+    private void checkIndex(int index, int size) throws ArrayListIndexOutOfBoundsException {
```
kateryna-deshchinska, line 30: you don't need to add parameter size in that method, we've already know size for our list

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index is incorrect: " + index);
+        }
+    }
+
+    private void removeValue(int index) {
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        size--;
+        values[size] = null;
+    }
```
kateryna-deshchinska, line 40: We don't need to create separate method for it, because we use it only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+            if (values[i] == element || values[i] != null && values[i].equals(element)) {
+                index = i;
+            }
+        }
+        if (index < 0) {
+            throw new NoSuchElementException("Element " + element + "is not find");
+        }
+        remove(index);
+        return element;
     }
```
kateryna-deshchinska, line 99: you may "return remove(i)" inside for loop, and throw exception, if there no such element after loop, and your code will be more readable and clear

kateryna-deshchinska (changes requested):

PR #631, opened 2022-09-29
v1 own ArrayList implement

EvgenyTyshchenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+            resize();
+        }
+        data[this.size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
```
ChumakAn, line 27: it's better to increment the size variable at the end of the method, after adding the element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        checkIndex(index);
+        if (data.length == this.size) {
+            resize();
+        }
+        for (int i = this.size - 2; i >= index; i--) {
```
ChumakAn, line 32: 2 - is magic number

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
-
+        checkIndex(index);
+        data[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        this.size--;
```
ChumakAn, line 60: the same for size here, you should decrement the size after removing the element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        T removedData = (T) data[index];
+        for (int i = index; i < this.size; i++) {
+            this.data[i] = this.data[i + 1];
+        }
```
ChumakAn, line 64: it's better to use System.arraycopy, read the common mistakes

ChumakAn (changes requested):
Good job, let's make this solution better

ChumakAn (approved):

PR #632, opened 2022-10-01
made my own ArrayList and implemented methods of List interface

igorgalenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+                    String.format("Out of index: trying to work with %s position but size is %s",
+                            index, size));
+        }
+    }
+
+    private void resizeStorage() {
+        T[] storageCopy = (T[]) new Object[size];
+        System.arraycopy(storage, 0, storageCopy, 0, size);
+        storage = (T[]) new Object[storage.length + (storage.length >> 1)];
+        System.arraycopy(storageCopy, 0, storage, 0, size);
```
KostyaNikiforov, line 100: It would be enough to create a new list only once and copy old values there.

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #636, opened 2022-10-04
implemented methods of ArrayList

alex-anofriev:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_INDEX = 0;
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] defaultArray = new Object[INITIAL_CAPACITY];
+    private int sizeCounter = 0;
+    private int newCapacity = 10;
```
KostyaNikiforov, line 10: Can you just use an array length?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+        for (int i = 0; i <= sizeCounter; i++) {
+            if (element == null) {
+                sizeCounter--;
+                return null;
+            }
+            if (element.equals(defaultArray[i])) {
+                Object[] temp = defaultArray.clone();
+                defaultArray = new Object[newCapacity];
+                System.arraycopy(temp, DEFAULT_INDEX, defaultArray, DEFAULT_INDEX, i);
+                System.arraycopy(temp, i + 1, defaultArray, i, sizeCounter);
```
KostyaNikiforov, line 125: Can we just create a new empty array and copy the value from the previous one there?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+    }
+
+    @Override
+    public ArrayList<T> clone() {
+        try {
+            return (ArrayList<T>) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException("Can not clone", e);
+        }
     }
```
KostyaNikiforov, line 150: It can be implemented without cloning. Let's try to simplify your solution

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+
+    private void increaseArray(List<T> list) {
+        Object[] tempList = new Object[list.size()];
+        for (int i = 0; i < list.size(); i++) {
+            tempList[i] = list.get(i);
+        }
+        Object[] temp = defaultArray.clone();
+        defaultArray = new Object[newCapacity];
+        System.arraycopy(temp, DEFAULT_INDEX, defaultArray, DEFAULT_INDEX, sizeCounter);
+        System.arraycopy(tempList, DEFAULT_INDEX, defaultArray, sizeCounter, tempList.length);
```
KostyaNikiforov, line 53: Same. It is possible to use arraycopy method only once

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+                return null;
+            }
+            if (element.equals(defaultArray[i])) {
+                T removedElement = (T)defaultArray[i];
+                System.arraycopy(defaultArray, i + 1, defaultArray, i,sizeCounter - i);
+                sizeCounter--;
+                return removedElement;
+            }
+        }
+        throw new NoSuchElementException("No such element");
```
KostyaNikiforov, line 125: Make the exception message more informative. Add an information about the element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        return temp;
+    }
+
+    public int grow() {
+        return arrayLength + (arrayLength / 2);
+    }
+
+    private void rangeChecker(int index) {
+        if ((index < 0 || index >= sizeCounter) && index != DEFAULT_INDEX) {
+            throw new ArrayListIndexOutOfBoundsException("Out of bounds");
```
KostyaNikiforov, line 71: Same

KostyaNikiforov (changes requested):

alex-anofriev:
Is this code okay for a beginner? I am not sure about the way I am trying to find solutions for my tasks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        System.arraycopy(array, DEFAULT_INDEX, temp, DEFAULT_INDEX, sizeCounter);
+        return temp;
+    }
+
+    private Object[] increaseArray(Object[] array, T value, int index) {
+        Object[] temp = new Object[arrayLength];
+        System.arraycopy(array,DEFAULT_INDEX, temp, DEFAULT_INDEX, index);
+        System.arraycopy(array, index, temp, index + 1, sizeCounter - index);
+        temp[index] = value;
+        return temp;
```
KostyaNikiforov, line 50: You can simplify your solution. Let's copy the array only one time. You can use next algorithm:
````
Objects[] data = values;
values = new Object[size * 3 / 2];
````

**Review by KostyaNikiforov — CHANGES_REQUESTED** [review]

File: src/main/java/core/basesyntax/ArrayList.java [hunk 8]
```diff
@@ -1,48 +1,133 @@
...
+        return (T)defaultArray[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        rangeChecker(index);
+        if (index == sizeCounter) {
+            sizeCounter++;
+        }
```
hronograf, line 87: seems like this condition is always `false`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i <= sizeCounter; i++) {
+            if (element == null) {
+                sizeCounter--;
+                return null;
+            }
```
hronograf, line 112: is this logic right?
check this
```
arr.add("a");
arr.add(null);
arr.add("b");
arr.remove(null);
print(arr.get(1));
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
+        }
+        System.arraycopy(defaultArray, index + 1, defaultArray, index, sizeCounter - index);
+        sizeCounter--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i <= sizeCounter; i++) {
```
hronograf, line 108:
```suggestion
        for (int i = 0; i < sizeCounter; i++) {
```

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        rangeChecker(index);
+        T removedElement = (T) defaultArray[index];
+        if (index == sizeCounter - 1) {
+            System.arraycopy(defaultArray, DEFAULT_INDEX, defaultArray,
+                    DEFAULT_INDEX, sizeCounter - 1);
```
hronograf, line 85: I think we don't need this `System.arraycopy`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < sizeCounter; i++) {
+            if ((element != null && element.equals(defaultArray[i]))
+                    || element == defaultArray[i]) {
+                T removedElement = (T)defaultArray[i];
+                System.arraycopy(defaultArray, i + 1, defaultArray, i,sizeCounter - i);
+                sizeCounter--;
+                return removedElement;
```
hronograf, line 102: looks like something similar with `remove(int index)` 
can we use it here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
-        return null;
+        rangeChecker(index);
+        T removedElement = (T) defaultArray[index];
+        if (index == sizeCounter - 1) {
+            System.arraycopy(defaultArray, DEFAULT_INDEX, defaultArray,
+                    DEFAULT_INDEX, sizeCounter - 1);
+            sizeCounter--;
+            return removedElement;
+        }
+        System.arraycopy(defaultArray, index + 1, defaultArray, index, sizeCounter - index);
```
hronograf, line 89:
```suggestion
        System.arraycopy(defaultArray, index + 1, defaultArray, index, sizeCounter - index - 1);
```
with this you can remove if that above this line
try to figure out why

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
alex-anofriev: because it was "index + 1"  and so length to add can be additionally minus 1 if  i counted right

hronograf (approved):
Great!

PR #638, opened 2022-10-05
implemented task

GrishaSeniv:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+        if (size >= capacity) {
+            extendCapacity();
+        }
+        Object[] tempArr = new Object[++size];
+        System.arraycopy(listArr, 0, tempArr, 0, index);
+        tempArr[index] = value;
+        System.arraycopy(listArr, index, tempArr, index + 1, size - (index + 1));
+        System.arraycopy(tempArr, 0, listArr, 0, size);
```
ChumakAn, line 37: why do you copy the array 3 times?

ChumakAn: you should increment the size after the addition

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            add(list.get(i));
+        }
+    }
+
+    private void extendCapacity() {
+        capacity = capacity + 1 + (capacity >> 1);
+        Object[] tempArray = new Object[capacity];
+        System.arraycopy(listArr, 0, tempArray, 0, size);
+        listArr = new Object[capacity];
+        System.arraycopy(tempArray, 0, listArr, 0, size);
```
ChumakAn, line 55: same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        checkRange(index);
+        for (int i = 0; i < size; i++) {
+            if (index == i) {
+                listArr[i] = value;
+            }
+        }
```
ChumakAn, line 83: we use this lopp a couple of times? can we create a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        throw new NoSuchElementException(ELEMENT_ERROR_MESSAGE);
+    }
+
+    private T getRemovedElementAndRemove(int i) {
+        Object[] temp = new Object[--size];
+        final T foundElement = (T) listArr[i];
+        System.arraycopy(listArr, 0, temp, 0, i);
+        System.arraycopy(listArr, i + 1, temp, i, size - i);
+        System.arraycopy(temp, 0, listArr, 0, size);
```
ChumakAn, line 114: same

ChumakAn (changes requested):
try to avoid code duplication and calling the System.arraycopy a lot of times in the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     @Override
     public T get(int index) {
-        return null;
+        checkRange(index);
+        T element = null;
+        int i = getIndexIfFind(size, index);
+        if (i != -1) {
+            element = (T) listArr[i];
+        }
+        return element;
```
ChumakAn, line 69:
```suggestion
        checkRange(index);
        int i = getIndexIfFind(size, index);
        return (i!= -1) ? (T) listArr[I] :  null;
```

ChumakAn (changes requested):
Good! Let's make it better. Are you sure you need to use the loop to check if for example  index 4 is in range of size 8?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+        if (size >= capacity) {
+            extendCapacity();
+        }
```
ChumakAn, line 32: code duplication with 18-20 lines, can me a separate method and called extendIfNecessary or like that

ChumakAn (changes requested):
You're on the right track, small comments

liliia-ponomarenko (approved):
good job!🔥

PR #639, opened 2022-10-05
Created ArrayList so that it works like an original one

YuriiHryniv:


OleksandrKorniychuk (changes requested):

OleksandrKorniychuk (approved):
Very nice! 🚀

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int capacity = 10;
```
hronograf, line 6: it would be better if `capacity` was a constant (`private static int INITIAL_CAPACITY`)
and after that we can use `elementData.length` to get info about current capacity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    ArrayList() {
+        elementData = new Object[capacity];
+    }
+
     @Override
     public void add(T value) {
-
+        if (currentNumberOfMembers >= capacity) {
+            grow();
+        }
```
hronograf, line 18: like here, we can use `elementData.length` instead of `capacity`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
-        return null;
+        int index;
+        boolean isFound = false;
+        for (int i = 0; i < currentNumberOfMembers; i++) {
+            if (elementData[i] == element || elementData[i] != null
+                    && elementData[i].equals(element)) {
+                index = i;
+                isFound = true;
+                remove(index);
+                break;
```
hronograf, line 77: can we just write `return remove(index);`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+            }
+        }
+        if (!isFound) {
+            throw new NoSuchElementException();
+        }
+        return element;
     }
```
hronograf, line 84: so here we will be able to just throw the exception

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static int INITIAL_CAPACITY = 10;
+    private int currentNumberOfMembers = 0;
+    private Object [] elementData;
```
hronograf, line 8:
```suggestion
    private Object[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index exception");
+        }
+        if (currentNumberOfMembers >= INITIAL_CAPACITY) {
+            grow();
+        }
```
hronograf, line 29: let's do something like in `add(T value)` method, 
because now we increase capacity each time after we added 10 elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        int index;
+        boolean isFound = false;
+        for (int i = 0; i < currentNumberOfMembers; i++) {
+            if (elementData[i] == element || elementData[i] != null
+                    && elementData[i].equals(element)) {
+                index = i;
+                return remove(index);
```
hronograf, line 75:
```suggestion
                return remove(i);
```
also let's remove `index` variable

hronograf (changes requested):

PR #643, opened 2022-10-07
implemented ArrayList class

nazariipertsak:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        checkForIndexRemove(index);
+        oldArray = array;
+        if (index == 0) {
+            System.arraycopy(oldArray, index + 1, array, index, size);
+            return;
+        }
+        if (size == array.length) {
+            index = 0;
+        }
+        System.arraycopy(oldArray, index + 1, array, index, size - 1);
```
liliia-ponomarenko, line 120: you can simplify  it if use ```System.arraycopy(array, index + 1, array, index, size - index - 1);```

liliia-ponomarenko (approved):
Great 😊

PR #645, opened 2022-10-08
Solution

OleksiiMasaltsev:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
     public void add(T value, int index) {
-
+        ensureCapacity();
+        ensurePosition(index - 1);
+        Object[] temp = new Object[storage.length];
+        System.arraycopy(storage, 0, temp, 0, index);
+        temp[index] = value;
+        System.arraycopy(storage, index, temp, index + 1, size - index);
+        storage = temp;
+        size++;
```
GHSpeedWagon, line 41: maybe in we can use add(value) method if index one more then size

i mean 
if (index - 1 == size) {
add(value);
return:
} else {
your code

src/test/java/core/basesyntax/ArrayListTest.java
```diff
@@ -7,6 +7,43 @@
 public class ArrayListTest {
     private static final int ELEMENTS_COUNT = 1000;
 
+    // My test
+//    @Test
+//    public void customAddElementToArrayList() {
+//        ArrayList<String> arrayList = new ArrayList<>();
+//        arrayList.add("Test");
+//        arrayList.add(null);
```
GHSpeedWagon, line 15: and we can use /* */ if we want to comment some lines

GHSpeedWagon (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
+        storage = new Object[INITIAL_CAPACITY];
+    }
+
+    public ArrayList(int capacity) {
+        if (capacity >= 0) {
+            storage = new Object[capacity];
+        } else {
+            throw new ArrayListIndexOutOfBoundsException(OUT_OF_BOUNDS_MESSAGE);
+        }
+    }
```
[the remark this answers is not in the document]
OleksiyMasaltsev: This is my implementation of original constructor in java.util.ArrayList

src/test/java/core/basesyntax/ArrayListTest.java
OleksiyMasaltsev: I will simply delete this )

GHSpeedWagon (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+
+    private void reduceCapacity(int index) {
+        Object[] temp = new Object[storage.length - 1];
+        System.arraycopy(storage, 0, temp, 0, index);
+        System.arraycopy(storage, index + 1, temp, index, size - index - 1);
+        storage = temp;
+        size--;
+    }
```
hronograf, line 120: it is a cool idea that we should reduce capacity!
and it would be better to have some check like 
```
if (storage.length > size + size / 2) { 
  Object[] temp = new Object[someSize];
  ...
}
```
so we don't need to create new array for each removing, 
also let's separate `removing` and `reducing capacity`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        if (size == storage.length) {
+            increaseCapacity(storage.length);
+        }
+    }
+
+    private void ensurePosition(int index) {
+        if (index + 1 < 0 || index + 1 > size) {
+            throw new ArrayListIndexOutOfBoundsException(OUT_OF_BOUNDS_MESSAGE);
+        }
+    }
```
hronograf, line 112: what if `index=-1`
try this `yourArray.get(-1);`
and can we change `index + 1 > size` to `index > size`?

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/ArrayList.java
OleksiyMasaltsev: The thing is that I use reduceCapacity for deleting any object in the List and we cannot have situation when internal array  becomes larger than 1.5 of size. So there is no overusing of System.arraycopy.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index - 1 == size) {
+            add(value);
+        } else {
+            ensureCapacity();
+            ensureAddition(index);
+            Object[] temp = new Object[storage.length];
+            System.arraycopy(storage, 0, temp, 0, index);
```
hronograf, line 33: we don't need this lines

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+            temp[index] = value;
+            System.arraycopy(storage, index, temp, index + 1, size - index);
+            storage = temp;
```
hronograf, line 36: we don't need `temp` variable, we can swap 34 and 35 lines and after that use `storage` instead of `temp`

src/main/java/core/basesyntax/ArrayList.java
hronograf: we have methods `grow` and `add`, they do different functions and `grow` doesn't add element 
same thing here, let's reduce capacity and remove element in separate methods

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
hronograf: previous comment wasn't applied
we can remove 32 and 33 lines and use `storage` instead of `temp`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+        throw new ArrayListIndexOutOfBoundsException(OUT_OF_BOUNDS_MESSAGE);
+    }
+
+    private void removeBy(int index) {
+        Object[] temp = new Object[storage.length];
+        System.arraycopy(storage, 0, temp, 0, index);
+        System.arraycopy(storage, index + 1, temp, index, size - index - 1);
+        storage = temp;
+        size--;
     }
```
hronograf, line 135:
```suggestion
    private void removeBy(int index) {
        System.arraycopy(storage, index + 1, storage, index, size - index - 1);
        size--;
    }
```
this should work

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
OleksiyMasaltsev: Finally I understand how System.arraycopy works! 😀

hronograf (approved):
Good job! 🔥

PR #649, opened 2022-10-10
homework has been done

AndriiMatveev:


NSay0 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        indexCheck(index);
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
+        indexCheck(index);
+        System.arraycopy(elementData, index,
+                elementData, index + 1,
+                size);
```
NSay0, line 94: i don't really understand why we need sys.arraycopy here.

NSay0 (changes requested):

NSay0 (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] DEFAULTCAPACITY_EMPTY_ELEMENTDATA = {};
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = DEFAULTCAPACITY_EMPTY_ELEMENTDATA;
+    }
```
hronograf, line 13: if we have `DEFAULT_CAPACITY` let's write `elementData = new Object[DEFAULT_CAPACITY]`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+
+    private Object[] grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        Object[] oldElement = elementData;
+        if (oldCapacity > 0 || elementData != DEFAULTCAPACITY_EMPTY_ELEMENTDATA) {
+            int prefLength = oldCapacity + Math.max(minCapacity - oldCapacity, oldCapacity >> 1);
```
hronograf, line 19: line 19 👍

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
     public void add(T value, int index) {
-
+        rangeCheckForAdd(index);
+        if (size == elementData.length) {
+            elementData = grow(size + 1);
+        }
+        System.arraycopy(elementData, index,
+                elementData, index + 1,
+                size - index);
+        elementData[index] = (T) value;
```
hronograf, line 59:
```suggestion
        elementData[index] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        size = size + 1;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        Object[] a = new Object[list.size()];
```
hronograf, line 65: `a` - isn't good name for variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        for (int i = 0; i < a.length; i++) {
+            a[i] = list.get(i);
+        }
+        int numNew = a.length;
```
hronograf, line 69: looks like `numNew` is a redundant variable, we can get this info from `a.length`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        if (numNew == 0) {
+            return;
+        }
+        Object[] elementData = this.elementData;
```
hronograf, line 73: `Object[] elementData` redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        final int s = size;
```
hronograf, line 74: `s` redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        elementData[index] = value;
+    }
 
+    private void fastRemove(Object[] es, int i) {
+        final int newSize;
+        if ((newSize = size - 1) > i) {
+            System.arraycopy(es, i + 1, es, i, newSize - i);
+        }
+        es[size = newSize] = null;
     }
```
hronograf, line 101: `es` and `i` isn't obvious names, let's rename these variables
`if ((newSize = size - 1) > i)` - let's separate assigning and check inside if
`es[size = newSize]` - same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        T oldValue = element;
+        for (int j = 0; j < size; j++) {
+            if (element == null && elementData[j] == null
+                    || (element != null && element.equals(elementData[j]))) {
+                fastRemove(elementData, j);
+                return oldValue;
```
hronograf, line 119: it would be better to user `remove(int index)` here

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
AndriiMatveev: Math.max(minCapacity - oldCapacity, oldCapacity >> 1) I've actualy looked it in default ArrayList. But I thought it's a good logic for addAll. If adding array is bigger than half of our elementData, we'll need to grov twice. And it's will last too long, so better to grow once to the right size.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        Object[] addingObject = new Object[list.size()];
+        for (int i = 0; i < addingObject.length; i++) {
+            addingObject[i] = list.get(i);
+        }
+        if (addingObject.length == 0) {
+            return;
+        }
```
hronograf, line 70:
```suggestion
        if (list.size() == 0) {
            return;
        }
```
and it would be better to check this before loop

hronograf (approved):
Nice!

PR #651, opened 2022-10-10
Implement add(value), add(index, value), addAll(List) , get(index), s…

marchTooMuch:
…et(value, index),  remove(index), remove(value). Made isEmpty(), checkIndex(index0 methods

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,119 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIALIZATION_LENGTH = 10;
+    private T[] array;
+    private int size = 0;
+
+    ArrayList() {
+        array = (T[])new Object[INITIALIZATION_LENGTH];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            makeBiggerArray();
+            array[size] = value;
+            size++;
+            return;
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant add element"
+                    + "for such index");
+        }
+        if (array.length == size) {
+            makeBiggerArray();
+        }
+        System.arraycopy(array,index,array,index + 1, size - index);
+        array[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant get element"
+                    + "for such index");
+        }
+        return (T)array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant set element"
+                    + "for such index");
+        }
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant remove element"
+                    + "for such index to add");
+        }
+        size--;
+        T returnValue = array[index];
+        System.arraycopy(array, index + 1, array, index, size - index);
+        return returnValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if (array[i] == null && element != null) {
+                continue;
+            }
+            if (array[i] == null && element == null) {
+                remove(i);
+                return null;
+            }
+            if (array[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Can't remove element for such"
+                + " value " + element);
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant add element"
```
GrishaSeniv, line 30: Replace all hardcoded messages with constants

GrishaSeniv: And maybe it would be better to add throw new ArrayListIndexOutOfBoundsException to the checkInIndex method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant get element"
+                    + "for such index");
```
GrishaSeniv, line 51: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        }
+        return (T)array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant set element"
+                    + "for such index");
```
GrishaSeniv, line 60: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        }
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (checkIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Cant remove element"
+                    + "for such index to add");
```
GrishaSeniv, line 69: same

GrishaSeniv (changes requested):

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution😉

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+    public void makeBiggerArrayIfNeeded() {
+        if (size == array.length) {
+            int newSize = (int) (size * GROW_INDEX);
+            T[] newArray = (T[]) new Object[newSize];
+            System.arraycopy(array, 0, newArray, 0, array.length);
+            array = newArray;
+        }
     }
 }
+
```
liliia-ponomarenko, line 99: Redundant  empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void checkIndex(int index) {
```
liliia-ponomarenko, line 84: this method can be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(ERROR_MESSAGE_INDEX);
+        }
+    }
+
+    public void makeBiggerArrayIfNeeded() {
```
liliia-ponomarenko, line 90: the same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if (array[i] == null && element != null) {
+                continue;
+            }
+            if ((array[i] == null && element == null) || array[i].equals(element)) {
+                return remove(i);
+            }
```
liliia-ponomarenko, line 69: I guess  you can simplify  it more 
```suggestion
         
            if (array[i] ==  element ||  element != null && element.equals(array[i])) {
                return remove(i);
            }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIALIZATION_LENGTH = 10;
+    private static final String ERROR_MESSAGE_INDEX = "Your index is incorrect";
+    private static final String ERROR_MESSAGE_ELEMENT = "Can't remove element for such";
+    private static final double GROW_INDEX = 1.5;
+    protected T[] array;
+    protected int size = 0;
```
liliia-ponomarenko, line 11: why protected?

liliia-ponomarenko (changes requested):
Not all my previous  comments were fixed

src/main/java/core/basesyntax/ArrayList.java
marchTooMuch: When it was private you said "Let’s add the correct access modifiers". I don`t understand so which it must be.

src/main/java/core/basesyntax/ArrayList.java
marchTooMuch: I think I cant simplify it more, this code raising error, I need to check if array[i] is null and if it is and element is not I need to continue cycle but not to return.

liliia-ponomarenko (approved):
good job! 👍🏻

PR #653, opened 2022-10-10
added solution to arrayList task

Bohdan20b:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_ARRAY_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= elementData.length) {
+            elementData = grow();
```
GrishaSeniv, line 20: You have code duplication with 33-34, maybe it would be better to separate it to another method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        }
+        if (checkIndex(index)) {
+            System.arraycopy(elementData, index,
+                    elementData, index + 1,
+                    size - index);
+            elementData[index] = value;
+            size++;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
```
GrishaSeniv, line 43: Replace all hardcoded messages to constants

GrishaSeniv (changes requested):

GrishaSeniv (approved):

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: usually, error messages are not made constants

liliia-ponomarenko (changes requested):
You have a lot of if-else constructions in your solution, I recommend you think about how to simplify it. It makes your code clear.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+        } else {
+            throw new ArrayListIndexOutOfBoundsException(OUT_OF_BOUNDS_MESSAGE);
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T removedElement;
+        if ((size == elementData.length) && index + 1 == size) {
```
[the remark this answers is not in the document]
Bohdan20b: without it there is an OutOfBound exception for last index of array [10], index = 11, so i separated it into this conditional

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(OUT_OF_BOUNDS_MESSAGE);
+        } else {
+            System.arraycopy(elementData, index,
+                    elementData, index + 1,
+                    size - index);
+            elementData[index] = value;
+            size++;
+            elementDataGrow(size);
```
liliia-ponomarenko, line 34: better do this before adding the element

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
liliia-ponomarenko: You just need to modify array copying ```System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);``` and it will work.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == null && element == null) {
+                size--;
+                return null;
+            }
+            if (elementData[i] != null && elementData[i].equals(element)) {
+                return remove(i);
+            }
```
liliia-ponomarenko, line 82: if (array[i] ==  element ||  element != null && element.equals(array[i])) {
                return remove(i);
            }
you can simplify it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        }
+    }
+
+    private void elementDataGrow(int size) {
+        if (size == elementData.length) {
+            elementData = grow();
+        }
     }
 }
+
```
liliia-ponomarenko, line 115: Redundant empty line

liliia-ponomarenko (changes requested):
Let’s fix a few mistakes.

liliia-ponomarenko (approved):
great 😊

PR #655, opened 2022-10-10
realised ArrayList class with overrided methods

apc10bbe:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double RESIZE_COUNT = 1.5;
+    private Object[] elementData;
+    private int defaultSize;
+    private int count;
+
+    public ArrayList() {
+        defaultSize = 10;
```
PavloPolovyi, line 13: Make this as constant field.

PavloPolovyi, line 13: Make this as constant field in class (private static final).

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        elementData = new Object[defaultSize];
+        count = 0;
+    }
+
+    private boolean rangeCheckForCount(int index) {
+        if (index >= count || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound of array.");
```
PavloPolovyi, line 20: You can make message more informative by adding index value to it, like it made in standard java library (e.g. "Index 2 out of bounds for length 2").

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+            Object[] localeData = elementData;
+            elementData = new Object[defaultSize *= RESIZE_COUNT];
+            System.arraycopy(localeData, 0, elementData, 0, localeData.length);
+        }
+    }
+
+    public boolean checkValue(T element, int i) {
+        if (element != null && element.equals(elementData[i])) {
+            return true;
+        }
```
PavloPolovyi, line 36: Unnecessary return true statement. You can make this code shorter. If I were you I will rewrite this method as: if (element == null) {return element == elementData[i]} else { return element.equals(elementData[i])}.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound of array.");
+        }
+        return true;
+    }
+
+    private void checkArrayLength(int count) {
+        if (count >= defaultSize) {
+            Object[] localeData = elementData;
+            elementData = new Object[defaultSize *= RESIZE_COUNT];
+            System.arraycopy(localeData, 0, elementData, 0, localeData.length);
```
PavloPolovyi, line 29: As i understood you use here default size to compare if there enough space, but you never update it. I mean, this if statement always will be true, when count reach default size in value. And it's mean that you always create new element data. My advice - as i mentioned early - make DEFAULT_SIZE as constant, and change logic of this method by taking in consideration actual size of your elementData.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        defaultSize = 10;
+        elementData = new Object[defaultSize];
+        count = 0;
+    }
+
+    private boolean rangeCheckForCount(int index) {
+        if (index >= count || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound of array.");
+        }
+        return true;
```
PavloPolovyi, line 22: You actually don't need to return anything from this method. Just make it void and use in all your methods as: rangeCheckForCount(index) without if. If something wrong with index - ok, exception will be throwed. With this approach you can delete unnecessary <return null;> statement from many methods in your code.

PavloPolovyi (changes requested):

src/main/java/core/basesyntax/ArrayList.java
PavloPolovyi: Sorry, everything ok, i haven't noticed it : defaultSize *= RESIZE_COUNT. But again, it's better to make default size constant and don't change it in code.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double RESIZE_COUNT = 1.5;
+    private Object[] elementData;
+    private int startSize;
+    private int count;
+
+    public ArrayList() {
+        startSize = 10;
```
liliia-ponomarenko, line 13: this value  should  be constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-
+        if (index == count) {
+            add(value);
+        } else {
+            rangeCheckForCount(index);
+            checkArrayLength(count);
+            Object[] localeOne = Arrays.copyOfRange(elementData, 0, index);
+            Object[] localeTwo = Arrays.copyOfRange(elementData, index, count);
+            System.arraycopy(localeOne, 0, elementData, 0, localeOne.length);
+            System.arraycopy(localeTwo, 0, elementData, localeOne.length + 1, count - index);
```
liliia-ponomarenko, line 34: Can you simplify this code?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        rangeCheckForCount(index);
+        final Object deletedElement = elementData[index];
+        Object[] localeOne = Arrays.copyOfRange(elementData, 0, index);
+        Object[] localeTwo = Arrays.copyOfRange(elementData, index + 1, count);
+        System.arraycopy(localeOne, 0, elementData, 0, localeOne.length);
+        System.arraycopy(localeTwo, 0, elementData, localeOne.length, localeTwo.length);
```
liliia-ponomarenko, line 66: the same here. Do you really need separate variables for this action?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        count--;
+        return (T) deletedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        boolean isDelete = false;
```
liliia-ponomarenko, line 73: you can avoid using  this flag

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    }
+
+    private void rangeCheckForCount(int index) {
+        if (index >= count || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bound for length " + count);
+        }
+    }
+
+    private void checkArrayLength(int count) {
```
liliia-ponomarenko, line 104: better name this method grow() or resize()

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution 😉

liliia-ponomarenko (approved):
LGTM! 🔥

PR #661, opened 2022-10-11
solved jv array list task

HryhorashPavlo:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+            resize();
+        }
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        checkIndex(index);
```
hronograf, line 38: so if we will add value at index `-1` our size will be increased?

hronograf (changes requested):
Good!

hronograf (approved):
Great! 🔥

PR #663, opened 2022-10-11
jv-array-list - completed task

vazonchik:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == THIS_IS_ZERO;
+    }
+
+    public void checkGrow() {
+        if (size == currentArray.length) {
```
GrishaSeniv, line 82: maybe it would be better to use utils methods with private access modifier?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+            currentArray = Arrays.copyOf(currentArray, currentArray.length
+                    + currentArray.length / THIS_IS_TWO);
+        }
+    }
+
+    public void chekIndex(int index) {
```
GrishaSeniv, line 88: same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     public void add(T value) {
-
+        checkGrow();
+        currentArray[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
```
GrishaSeniv, line 22: i think you should increase size in the end

GrishaSeniv (changes requested):

src/main/java/core/basesyntax/ArrayList.java
GrishaSeniv: hmmm, I wrote this because I got the same comment))

GrishaSeniv (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int THIS_IS_ZERO = 0;
+    private static final int THIS_IS_ONE = 1;
+    private static final int THIS_IS_TWO = 2;
```
liliia-ponomarenko, line 10: so informative constants  names, let's think about how to rename this value.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+    private T[] currentArray = (T[]) new Object[DEFAULT_CAPACITY];
```
liliia-ponomarenko, line 11: Let’s  initialize this variable in constructor

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: agree, it must work correctly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == THIS_IS_ZERO;
+    }
+
+    private void checkGrow() {
+        if (size == currentArray.length) {
+            currentArray = Arrays.copyOf(currentArray, currentArray.length
+                    + currentArray.length / THIS_IS_TWO);
```
liliia-ponomarenko, line 84: Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution 😉

vazonchik:
when i put "size++" to end of method add by index, tests are not performed correctly and i can't figure it out why

liliia-ponomarenko (approved):
Good job! 👏

PR #664, opened 2022-10-11
Complete the task;

Mar4uk21:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_INDEX];
+    }
+
+    public void exception() {
+        throw new ArrayListIndexOutOfBoundsException("Index out of bounds for length");
+    }
```
luka-lozovyk, line 16: It is better to use verbs in method naming, as long as method does something.

luka-lozovyk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size == 0;
+    }
+
+    public void grow() {
+        T[] newData = (T[]) array;
+        array = new Object[size + (size >> 1)];
+        System.arraycopy(newData, 0, array, 0, size);
+
     }
 }
```
luka-lozovyk, line 103: Good job!

luka-lozovyk (approved):

src/main/java/core/basesyntax/ArrayList.java
hronograf: yes, we should use verbs, and I don't see the reason why we need this method
also if we use it, it should be `private`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void grow() {
```
hronograf, line 97:
```suggestion
    private void grow() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        T[] newData = (T[]) array;
```
hronograf, line 98:
```suggestion
        Object[] oldData = array;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        array = new Object[size + (size >> 1)];
```
hronograf, line 99: `size >> 1` 😎

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throwException();
+        }
+        return (T) array[index];
     }
```
hronograf, line 53: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     public T remove(T element) {
-        return null;
+        T findElement;
+        for (int i = 0; i <= size; i++) {
+            if (element == array[i] || element != null && element.equals(array[i])) {
+                findElement = (T) array[i];
+                remove(i);
+                return findElement;
+            }
+        }
```
hronograf, line 83:
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == array[i] || element != null && element.equals(array[i])) {
                return remove(i);
            }
        }
```

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_INDEX = 10;
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_INDEX];
+    }
+
+    private void throwException(int index) {
```
hronograf, line 14:
```suggestion
    private void checkIndex(int index) {
```

hronograf (approved):
Great! 
@Mar4uk21 check the comment

PR #672, opened 2022-10-12
ArrayList implementation

AlinaOryshak:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        Object[] newElements = new Object[list.size()];
+        checkCapacity(list.size());
+        for (int i = 0; i < newElements.length; i++) {
+            newElements[i] = list.get(i);
+        }
+        System.arraycopy(newElements, 0, elementData, size, newElements.length);
+        size += newElements.length;
```
Andew-Miroshnikov, line 39: do we need Arraycopy.here ?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    }
+
+    private void indexCheck(int index) {
+        if (index < 0 || index > elementData.length) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bounds");
+        }
+        if (index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
```
Andew-Miroshnikov, line 95: if we separate this methods < it will be better

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        int index = getIndexByValue(element);
+        return remove(index);
+    }
+
+    private int getIndexByValue(T value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(value, elementData[i])) {
```
Andew-Miroshnikov, line 71: reed once more common mistakes ,

src/main/java/core/basesyntax/ArrayList.java
AlinaOryshak: but in this way I can check two conditions using one method
and if I separate them I will call two methods instead of one every time

Andew-Miroshnikov (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
     }
```
hronograf, line 20: nice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index + 1);
+        return (T) elementData[index];
     }
```
hronograf, line 43: what if we run `ourArray.get(-1);`?
I think we will get exception, but not from our `indexCheck` method, you can check it in debugger

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        indexCheck(index + 1);
+        elementData[index] = value;
     }
```
hronograf, line 49: same, because of `indexCheck` implementation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+
+    private int getIndexByValue(T value) {
+        for (int i = 0; i < size; i++) {
+            if (value == elementData[i]
+                    || (value != null && Objects.equals(value, elementData[i]))) {
+                return i;
+            }
+        }
+        throw new NoSuchElementException("No such element");
     }
```
hronograf, line 74: looks good
and also line 69
image

src/main/java/core/basesyntax/ArrayList.java
hronograf: I think we don't need `index > elementData.length` check
also, be careful, we have different constraints for `get(int index)` and `add(T value, int index)` methods

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/ArrayList.java
hronograf: what about my previous comment?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        checkCapacity(1);
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        checkCapacity(list.size());
```
hronograf, line 33: we will check capacity inside `add` method, so this check is redundant

hronograf (changes requested):

src/main/java/core/basesyntax/ArrayList.java
AlinaOryshak: ohh sorry, I've understood this after Andrew comment but forgot to delete delete Object

hronograf (approved):

PR #677, opened 2022-10-13
Implement ArrayList

Serhii-Prykhodko:
ArrayList implementation

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -15,7 +15,7 @@ public class ArrayList<T> implements List<T> {
 
     @Override
     public void add(T value) {
-        sizeCheck();
+        resize();
         arrList[size] = value;
         size++;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index:" + index + " out of size " + size);
+        }
+        if (arrList.length == size) {
+            grow();
```
TymurProkhorov, line 37: It will be better to move this 'if' conditional in resize() method to avoid code duplication. (duplicates in this method and in add(T value) as well).

TymurProkhorov (changes requested):

TymurProkhorov (approved):

liliia-ponomarenko (approved):
LGTM 🔥

PR #678, opened 2022-10-15
Arraylist task done

IvanDrebot:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int NEGATIVE_INDEX = -1;
+    private static final int DEFAULT_SIZE = 10;
+    private int size = 0;
```
anastasiavasylyk, line 9: it`s will be better if you initialize this field in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("No element with index: " + index);
+        }
```
anastasiavasylyk, line 49: do this check in separate method. It`s better when one method do one action

anastasiavasylyk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size == values.length) {
+            Object[] draft = values;
+            values = new Object[(int) (size * 1.5)];
```
ChumakAn, line 89: you should not use magic numbers, make 1.5. a constant

ChumakAn (changes requested):
Good, but read the common mistakes and don't use magic numbers

ChumakAn (approved):

PR #686, opened 2022-10-21
ArrayList implementation is done

Vitalina-V:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        for (int i = 0; i < size; i++) {
+            if (i == index) {
+                array[i] = value;
+            }
+        }
```
aaananas, line 50: you have direct access to that value, do you need for-loop?

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void checkIndex(int index) {
```
zachar-chorny, line 84:
```suggestion
    private void checkIndex(int index) {
```

zachar-chorny (approved):
Good job

PR #701, opened 2022-11-05
task solved

olei23:


aaananas (changes requested):
image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double SCALE_FACTOR = 1.5;
+    private int size;
+    private Object[] arrList;
```
Rommelua, line 9:
```suggestion
    private Object[] elements;
```
or **elementData**

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == arrList[i] || element != null && element.equals(arrList[i])) {
+                T value = (T) arrList[i];
+                remove(i);
+                return value;
```
Rommelua, line 61:
```suggestion
                return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+            }
+        }
+        throw new NoSuchElementException("this element does not exist");
```
Rommelua, line 64:
```suggestion
        throw new NoSuchElementException("this element does not exist " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
+    }
+
+    private void checkIndex(int index) {
```
Rommelua, line 72: **private** methods should be after **public** ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("index is not valid");
+        }
+    }
+
+    private void resize() {
```
Rommelua, line 78: What is the difference between resize and grow? I mean word meaning.
Better remove resize() and use IF before call grow() where you need it

Rommelua (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("index is not valid");
+        }
+    }
+
+    private void grow() {
```
Rommelua, line 88:
```suggestion
    private void growIfNeeded() {
```
The name should reflect inner logic. We do not grow every time here.

Rommelua (approved):

PR #718, opened 2022-11-20
First check

MaksymTseniukh:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (elements.length == size) {
+            Object[] newArray = new Object[elements.length + (elements.length >> 1)];
```
aleksandr-bevza, line 84: Please, don't use bit operations. It reduces readability of code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+    private Object[] elements;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        grow();
```
aleksandr-bevza, line 16: Remove length checking from grow() method. Grow() should be responsible for growing but not for checking size.
Otherwise it should be named like "checkLengthAndGrow()"
```suggestion
if (elements.length == size){
    grow();
}        
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
+        grow();
```
aleksandr-bevza, line 28: Remove length checking from grow() method. Grow() should be responsible for growing but not for checking size.
Otherwise it should be named like "checkLengthAndGrow()"
```suggestion
if (elements.length == size){
    grow();
}        
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #720, opened 2022-11-25
Java: jv-arraylist

AlenaVyrodova:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double MULTIPLIER = 1.5;
+    private Object[] elementsData;
+    private int size;
+
+    public ArrayList() {
+        this.elementsData = (T[]) new Object[DEFAULT_SIZE];
+        this.size = 0;
```
Rommelua, line 13:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        resize();
+        System.arraycopy(elementsData, index, elementsData, index + 1, size - index);
+        elementsData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        resize();
```
Rommelua, line 34:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        if (res) {
+            for (int i = 0; i < size; i++) {
+                if ((element == null && elementsData[i] == null)
+                        || (element != null && element.equals(elementsData[i]))) {
+                    index = i;
+                }
+            }
+        }
+        return index;
+    }
```
Rommelua, line 133: delete these methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        boolean res = containsElement(element);
+        T removeElement = null;
+        if (!res) {
+            throw new NoSuchElementException("This element is absent in elementsData " + element);
+        }
+        if (res) {
+            int indexOfElement = findIndexForDelete(element);
```
Rommelua, line 70: containsElement(element); - loop inside
findIndexForDelete(element) -  loop inside

use one loop in remove(T element) method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    }
+
+    private void checkIndexOutOfBound(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid " + index);
+        }
+    }
+
+    private void resize() {
+        if (size == elementsData.length) {
```
Rommelua, line 100: Put this IF out of `resize()` OR rename `resize() `
When I read **resize** I expect resizing without IF

Rommelua (changes requested):

src/main/java/core/basesyntax/ArrayList.java
AlenaVyrodova: Delete resize from addAll

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elementsData = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        createNewArray();
```
aleksandr-bevza, line 17: The problem here is in method name. It's still not a good name for this method. It's confusing to understand by method's name: what does this method do. We have already created new array in constructor, why do we need to create it again?

You can fix it in two ways:
1) You can name this method like: "checkSizeAndResize()". 
2) But better way is to take "IF" out of resize(): 
if (size == elementsData.length) {
    resize();
}

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        elementsData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexOutOfBoundForAdd(index);
+        createNewArray();
```
aleksandr-bevza, line 25: The problem here is in method name. It's still not a good name for this method. It's confusing to understand by method's name: what does this method do. We have already created new array in constructor, why do we need to create it again?

You can fix it in two ways:
1) You can name this method like: "checkSizeAndResize()". 
2) But better way is to take "IF" out of resize(): 
if (size == elementsData.length) {
    resize();
}

aleksandr-bevza (changes requested):

kozhukhovsky (approved):

PR #722, opened 2022-11-27
implemented ArrayList v. 1

ItsUnbidden:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+
+    private int size;
+    private Object[] elements;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_SIZE];
+        size = 0;
+    }
+
+    private Object[] grow() {
```
```suggestion
    private void grow() {
```
let this method change the array
also lets move grow to the end - private methods should be after public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
+        }
+    }
+
+    private int getElementIndex(Object element) {
+        for (int i = 0; i < size; i++) {
+            if ((elements[i] == null && element == null)
+                    || (elements[i] != null && element != null 
+                    && elements[i].equals(element))) {
```
aaananas, line 115:
```suggestion
            if ((elements[i] ==  element)
                    || (elements[i] != null && elements[i].equals(element))) {
```

aaananas (changes requested):

aaananas (approved):

PR #723, opened 2022-12-04
Added constants

metelska-anastasiia:
all methods were overridden
method increaseArray was created

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private static final double INCREASE_RATE = 1.5;
+    private T[] array;
+    private int size;
+
+    public ArrayList() {
+        array = (T[]) new Object[CAPACITY];
+    }
+
+    public void increaseArray() {
+        T[] arrayCopy = (T[]) new Object[(int) (array.length * INCREASE_RATE)];
+        System.arraycopy(array, 0, arrayCopy, 0, array.length);
+        array = arrayCopy;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size() >= array.length) {
+            increaseArray();
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
+        if (size() >= array.length) {
+            increaseArray();
+        }
+        if (index < size()) {
+            System.arraycopy(array, index, array, index + 1, size() - index);
+            array[index] = value;
+            size++;
+        }
+        if (index == size()) {
+            add(value);
+        }
 
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        T value = null;
+        if (index < 0 || index >= size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        } else {
+            value = array[index];
+        }
+        return value;
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        } else {
+            array[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Object removeElement;
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        } else {
+            removeElement = get(index);
+            System.arraycopy(array, index + 1, array, index, size - index - 1);
+            size--;
+        }
+        return (T) removeElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        if (size() >= array.length) {
+            increaseArray();
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size()) {
```
aleksandr-bevza, line 32: Please change 'size()' to 'size' everywhere. It's easier to read the code when we see call to the variable, but not to method
```suggestion
        if (index < 0 || index > size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+            increaseArray();
+        }
+        if (index < size()) {
+            System.arraycopy(array, index, array, index + 1, size() - index);
+            array[index] = value;
+            size++;
+        }
+        if (index == size()) {
+            add(value);
+        }
```
aleksandr-bevza, line 46: Better to move this part to the begin of method
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        }
```
aleksandr-bevza, line 34: Better to move this part to the begin of method
```suggestion
        }
        if (index == size) {
            add(value);
            return;
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        if (size() >= array.length) {
```
aleksandr-bevza, line 35:
```suggestion
        if (size == array.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+            increaseArray();
+        }
+        if (index < size()) {
+            System.arraycopy(array, index, array, index + 1, size() - index);
+            array[index] = value;
+            size++;
+        }
```
aleksandr-bevza, line 42:
```suggestion
        System.arraycopy(array, index, array, index + 1, size - index);
        array[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     @Override
     public T get(int index) {
-        return null;
+        T value = null;
+        if (index < 0 || index >= size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
+        } else {
+            value = array[index];
+        }
+        return value;
```
aleksandr-bevza, line 64:
```suggestion
        }
        return array[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        T value = null;
```
aleksandr-bevza, line 58:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size(); i++) {
+            if (array[i] != null && array[i].equals(element) || array[i] == element) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
aleksandr-bevza, line 96:
```suggestion
        throw new NoSuchElementException("Element not found " + element);
```

aleksandr-bevza (commented):
One more thing: you have this check four times:

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
```
aleksandr-bevza, line 5: You have this check 4 times:
        if (index < 0 || index > size()) {
            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
        } else {
It's better to create separate method for this check. After it your methods will look like this:
    
    public T get(int index) {
        checkIndex(index);
        return elements[index];
    }

        checkIndex(index);
        elements[index] = value;
    }

    public T remove(int index) {
        checkIndex(index);
        Object removeElement = elements[index];
        ...
    }

aleksandr-bevza (changes requested):

src/main/java/core/basesyntax/ArrayList.java
metelska-anastasiia: Tried to move it to the begin of the method, but method is failed

java.lang.AssertionError: Test failed! Size of array should be 1but it is 2 
Expected :1
Actual   :2

src/main/java/core/basesyntax/ArrayList.java
metelska-anastasiia: Try to get how to create this method

 private ArrayListIndexOutOfBoundsException checkIndex(int index) {
            throw new ArrayListIndexOutOfBoundsException("Index passed to the method is invalid");
        }
       return null;
    }

aaananas (approved):

PR #731, opened 2022-12-09
added solution

MaxMaksym:


Nick97-git (changes requested):
fix this first
Screenshot 2022-12-09 at 17 35 08

PR #732, opened 2022-12-09
added solution

MaxMaksym:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while ((list.size() + size) >= values.length) {
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
```
KostyaNikiforov, line 43: Can you use for-each?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while ((list.size() + size) >= values.length) {
+            grow();
+        }
```
KostyaNikiforov, line 42: Do you really need this logic? Don't you check it in the `add` method?

KostyaNikiforov, line 42: Do you really need this logic? Don't you check it in the `add` method?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        if (index == size - 1) {
+            grow();
+        }
```
KostyaNikiforov, line 65: How is it possible?)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return (T) value;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < values.length; i++) {
+            if ((element == null && values[i] == null)
+                    || element != null && (element.equals(values[i]))) {
```
KostyaNikiforov, line 78:
```suggestion
            if (element == values[i] || element != null && element.equals(values[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("Element does not exist");
+        }
+        return remove(index);
```
KostyaNikiforov, line 86:
```suggestion
              return remove(index);
            }
        }
        throw new NoSuchElementException("Element does not exist");
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #735, opened 2022-12-12
hw-array-list-solution

Bogdan-Pokhilko:
Created constant values. Created an empty array and int size, for future checks for size of the array. Created new method Arraylist to give an empty array default size. Created arrayCopy to increase the length of the array in case it's size will reach the limit. Created method checkIndex to check if given index is exists, otherwise throw the exception. Override the add method. Checked for array size, if it is equals to 10, increased it to 15. Otherwise, added to created array value and increased the size by 1. Override add method with index. Checked if index exists, if not, throw the exception. Then if size equals to the length of the values array (10) - increase array length to 15, (called arrayCopy method). After that coppy the value and index to created array. Increased size by 1. Override addAll method. with for loop added every single element of the given List of elements/values. Override get method. Checked if the index exists with checkIndex method, then returned requested index in created array. Override set method. Checked if the index exists with checkIndex method, then replaced/added given value with value in created array with the same index. Override remove(index) method. Checked if the index exists with checkIndex method, then found value by given index, and removed it plus decreased size by 1. Override remove(element) method. With for loop, checked if there is such element in our array, if so, removed element and decreased size by 1. Otherwise, throw an exception. Override size method. It should return int size;
Override isEmpty method. It's boolean, if it is true, made it return size 0.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double INCREASE_RATE = 1.5;
+    private T[] values;
+    private int size;
+
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public void arrayCopy() {
+        values = Arrays.copyOf(values, (int) (size * INCREASE_RATE));
+    }
+
+    public void checkIndex(int index) throws ArrayListIndexOutOfBoundsException {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index does not exist");
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == values.length) {
+            arrayCopy();
+        }
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index does not exist");
+        }
 
+        if (size == values.length) {
+            arrayCopy();
+        }
+
+        System.arraycopy(values, index, values, index + 1, size - index);
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return values[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double INCREASE_RATE = 1.5;
+    private T[] values;
+    private int size;
+
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public void arrayCopy() {
```
Rommelua, line 17: only interface methods should be public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        values = Arrays.copyOf(values, (int) (size * INCREASE_RATE));
+    }
+
+    public void checkIndex(int index) throws ArrayListIndexOutOfBoundsException {
```
Rommelua, line 21:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private static final double INCREASE_RATE = 1.5;
+    private T[] values;
+    private int size;
+
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public void arrayCopy() {
+        values = Arrays.copyOf(values, (int) (size * INCREASE_RATE));
```
Rommelua, line 18: [Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements)
Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        T[] valuesCopy = (T[]) new Object[(int) (values.length * INCREASE_RATE)];
+        System.arraycopy(values,0,valuesCopy, 0, values.length);
+        values = valuesCopy;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index does not exist");
+        }
+    }
```
aaananas, line 26: private methods should go after public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        size--;
+        return removedValue;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        for (int i = 0; i <= size; i++) {
+            if (Objects.equals(values[i], element)) {
```
aaananas, line 83: image

aaananas (changes requested):
great job

kozhukhovsky (approved):

PR #739, opened 2022-12-17
Jv array list solution

dreeeamcatcher:


karma-o (approved):
Looks good, but I have a feeling that you have been looking at the actual `ArrayList` implementation for too much(

PR #760, opened 2022-12-21
implement array list methodes

ar4ibaldd:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index != size) {
+            checkIndex(index);
+        }
```
Dmytro-Mk, line 26: if index == size you should just call add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] != null && elements[i].equals(element)
+                    || elements[i] == null && element == null) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Can`t find element");
```
Pandoryms: it is better to show which element it cannot find
```suggestion
        throw new NoSuchElementException("Element not found: " + element);
```

Dmytro-Mk, line 70:
```suggestion
        throw new NoSuchElementException("Can`t find element " + element);
```

Dmytro-Mk (changes requested):

Dmytro-Mk (approved):

Nick97-git (approved):

PR #766, opened 2022-12-24
arraylist task completed

OleksiiZaloznyi:


Uka0001 (approved):
Change for constant is optional but, I think, it will be better, the choice is on you. Good work!:rocket:

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        elements[size] = null;
+        return (T) element;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && elements[i] == null)
+                    || element != null && element.equals(elements[i])) {
```
KostyaNikiforov, line 70:
```suggestion
            if (element ==  elements[i] 
                    || element != null && element.equals(elements[i])) {
```

KostyaNikiforov (approved):

PR #768, opened 2022-12-26
Arraylist class was created and implemented

Andrew-Yakhimets:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        elementData[index] = value;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
```
Dmytro-Mk, line 58: not needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        checkCapacity();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index, true);
+        checkCapacity();
```
Dmytro-Mk, line 27:
```suggestion
        if (index == size) {
        add(value)
        }
        checkIndex(index);
        checkCapacity();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    private void checkIndex(int index, boolean add) {
+        boolean checkIndex = index >= size;
+        if (add) {
+            checkIndex = index > size;
+        }
+        if (index < 0 || checkIndex) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " is not valid for size " + size + ".");
+        }
     }
```
Dmytro-Mk, line 104:
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index
                    + " is not valid for size " + size + ".");
        }
    }
```

Dmytro-Mk (changes requested):

PR #770, opened 2022-12-27
complete(all): complete task beta

urlo0k:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return (size == 0);
```
okuzan, line 137: unnecessary brackets

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private int maxSize = 10;
+    private int size;
+    private T[] array;
+    private T[] arrayGrow;
```
okuzan, line 9: we don't need this field

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+
+    public ArrayList() {
+        array = (T[]) new Object[maxSize];
+    }
 
+    public void resize() {
+        if (size == maxSize) {
+            maxSize = (int) (maxSize * 1.5);
```
okuzan, line 17: make it `GROW_FACTOR` constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+            size++;
+        }
     }
 
     @Override
-    public T get(int index) {
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("No such index exception");
```
okuzan, line 74: You can write a more informative message. Include the current size and index that caused trouble. It will make debugging easier :)
Also you can make it's message a string constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        }
+        return array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("No such index exception");
+        }
```
okuzan, line 83: It's code duplication. Move this logic to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private int maxSize = 10;
```
okuzan, line 6: this should be constant. DEFAULT_SIZE would be more appropriate

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+    private T[] arrayGrow;
+
+    public ArrayList() {
+        array = (T[]) new Object[maxSize];
+    }
 
+    public void resize() {
+        if (size == maxSize) {
+            maxSize = (int) (maxSize * 1.5);
+            arrayGrow = (T[]) new Object[maxSize];
```
okuzan, line 18: you can replace `arrayGrow` with a temporary array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private int maxSize = 10;
+    private int size;
+    private T[] array;
```
okuzan, line 8: `values` would be more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        while (array.length < size + list.size()) {
+            maxSize = (int) (maxSize * 1.5);
+            arrayGrow = (T[]) new Object[maxSize];
+            System.arraycopy(array, 0, arrayGrow, 0, size);
+            array = arrayGrow;
+        }
+        for (int i = 0; i < list.size(); i++) {
+            array[size] = list.get(i);
+            size++;
+        }
```
okuzan, line 68: with refactored `add()` this could be one loop iterating over the list, calling `add()` on each of its member

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        resize();
+        for (int i = 0; i < size; i++) {
+            if ((array[i] != null && array[i] == value)
+                    || (array[i] != null && array[i].equals(value))) {
+                array[i] = value;
+                return;
+            }
+        }
+        array[size] = value;
+        size++;
```
okuzan, line 35: this is not java's ArrayList behavior. You need only the last two lines here, which can be combined into one BTW

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+    }
 
+    @Override
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size + 1 || index < 0) {
```
okuzan, line 40: index >= size + 1 can be rewritten as index > size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        resize();
+        if (index < size) {
+            T[] arrayGrow = (T[]) new Object[maxSize];
+            System.arraycopy(array, 0, arrayGrow, 0, index);
+            arrayGrow[index] = value;
+            System.arraycopy(array, index, arrayGrow, index + 1, size - index);
+            array = arrayGrow;
+            size++;
+            return;
+        }
```
okuzan, line 52: instead of all of this you can use System.arraycopy once (copy array into array)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        T[] arrayNew = (T[]) new Object[maxSize];
+        final T result;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("No such index exception");
+        }
```
okuzan, line 93: checking index should be the first thing you do in this method, also extract this logic to a dedicated method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        array[index] = value;
     }
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        T[] arrayNew = (T[]) new Object[maxSize];
```
okuzan, line 89: you don't need a new array, the thing you're probably missing is in System.arraycopy you can pass your array twice, as source and as a destination

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        array = arrayNew;
+        size--;
+        return result;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        T[] arrayNew = (T[]) new Object[maxSize];
```
okuzan, line 104: same here, do everything in-place, no need for a new array, it unjustifiably increases complexity, which is crucial to a Collection impl

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+                resultElement = true;
+                array = arrayNew;
+                size--;
+                return element;
+            }
+        }
+        if (resultElement == false) {
+            throw new NoSuchElementException("No such element exception");
+        }
+        return element;
```
okuzan, line 127: Let's do it simpler: you iterate your array (just like you do know) and when you find it (using check on line 115) you just call the earlier defined `remove(i)` (removing by index). Much less code and more readable

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    }
+
+    public void errorCheck(int index, int size) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("No such index exception: "
+                    + index + " for size" + size);
+        }
+    }
 
+    public void resize() {
```
aaananas, line 24: this method should be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    private int maxSize = DEFAULT_INDEX;
+    private int sizeOfArray;
+    private T[] arrayDefault;
+    private T[] arrayTemporary;
+
+    public ArrayList() {
+        arrayDefault = (T[]) new Object[maxSize];
+    }
+
+    public void errorCheck(int index, int size) throws ArrayListIndexOutOfBoundsException {
```
```suggestion
     private void rangeCheck(int index) {
```
no need to put size here - you have it in the class
move private methods after public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_INDEX = 10;
+    private static final double GROW_INDEX = 1.5;
+    private int maxSize = DEFAULT_INDEX;
+    private int sizeOfArray;
+    private T[] arrayDefault;
+    private T[] arrayTemporary;
```
aaananas, line 11:
```suggestion
    private T[] array;
```
temporary array should be local variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-    public void add(T value, int index) {
+    public void add(T value) {
+        resize();
+        for (int i = 0; i < sizeOfArray; i++) {
+            if ((arrayDefault[i] != null && arrayDefault[i] == value)
+                    || (arrayDefault[i] != null && arrayDefault[i].equals(value))) {
+                arrayDefault[i] = value;
+                return;
+            }
+        }
```
aaananas, line 42:
```suggestion
```
you should just add a new value to the end

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        arrayDefault[sizeOfArray] = value;
+        sizeOfArray++;
+    }
 
+    @Override
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
```
aaananas, line 48:
```suggestion
    public void add(T value, int index) {
```
it is a runtime exception no need to put it in method signature

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        sizeOfArray++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            T value = list.get(i);
+            add(value);
+            resize();
```
aaananas, line 69:
```suggestion
```
resize is called inside add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        boolean resultElement = false;
+        for (int i = 0; i < sizeOfArray; i++) {
+            if ((element != null && element.equals(arrayDefault[i]))
+                    || element == arrayDefault[i]) {
+                remove(i);
+                resultElement = true;
+                return element;
```
aaananas, line 102:
```suggestion
                return element;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        System.arraycopy(arrayDefault, index + 1, arrayDefault, index, sizeOfArray - index - 1);
+        sizeOfArray--;
+        return result;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        boolean resultElement = false;
```
aaananas, line 96:
```suggestion
    public T remove(T element) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+                    || element == arrayDefault[i]) {
+                remove(i);
+                resultElement = true;
+                return element;
+            }
+        }
+        if (resultElement != true) {
+            throw new NoSuchElementException("No such index exception: " + element + " is absent");
+        }
+        return element;
```
aaananas, line 108:
```suggestion
        throw new NoSuchElementException("No such index exception: " + element + " is absent");
```
if the element is found - it will be removed and method will be finished
thats why there is no need to add additianal check

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
okuzan: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private void resize() {
+        if (sizeOfArray == maxSize) {
+            maxSize = (int) (maxSize * GROW_INDEX);
+            T[] arrayTemporary = (T[]) new Object[maxSize];
+            System.arraycopy(array, 0, arrayTemporary, 0, sizeOfArray);
+            array = arrayTemporary;
+        }
+    }
+
+    private void errorCheck(int index) {
```
okuzan, line 100: method name doesn't make it clear what it checks, think of a better one, previous reviewer proposed `rangeCheck`  for example

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return sizeOfArray;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return sizeOfArray == 0;
+    }
+
+    private void resize() {
```
nacenik, line 91: better segregate logic from this method into two methods, cause size doesn't change every time that method is called: 
checkSize() - check if it needs to increase the size of the array 
grow() - increase array capacity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_INDEX = 10;
+    private static final double GROW_INDEX = 1.5;
+    private int maxSize = DEFAULT_INDEX;
+    private int sizeOfArray;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_INDEX = 10;
```
nacenik, line 6: It isn't the default index, it's the default capacity. Rename it

nacenik (changes requested):

Ivan95kos (approved):

PR #772, opened 2022-12-27
jv-array-list first solution

kcasPilot:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        if (size == dataArray.length) {
+            T[] temporaryArray = (T[]) new Object[size + (size >> 1)];
+            System.arraycopy(dataArray, 0, temporaryArray, 0, size);
+            this.dataArray = temporaryArray;
+        }
+    }
+
+    private void checkIndex(int index, int range) {
+        if (index < 0 || index > range) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of available range");
```
aleksandr-bevza, line 100: Include index and size to this message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+
+    public ArrayList() {
+        this.dataArray = (T[]) new Object[DEFAULT_ARRAY_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        checkAndGrowArraySize();
+        this.dataArray[size] = value;
```
aleksandr-bevza, line 17: "this" is no needed here. It will be added by default, no need to write it. Fix in other places
```suggestion
        dataArray[size] = value;
```

aleksandr-bevza (changes requested):
Good job! Just small fixes needed

zachar-chorny (approved):

PR #775, opened 2023-01-01
added proper solution for this task

polikarpov13:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_SIZE = 10;
+    private Object[] values;
+    private int size;
+    private int arraySize = DEFAULT_SIZE;
+    private T removedObject;
```
okuzan, line 10: we don't need these fields

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_SIZE = 10;
+    private Object[] values;
```
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+    private int size;
+    private int arraySize = DEFAULT_SIZE;
+    private T removedObject;
 
+    public ArrayList() {
+        values = new Object[DEFAULT_SIZE];
```
```suggestion
        values = (T[]) new Object[DEFAULT_SIZE];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+    public int size() {
+        return size;
     }
 
     @Override
-    public void addAll(List<T> list) {
+    public void add(T object) {
+        resize();
+        values[size] = object;
+        size++;
```
```suggestion
        values[size++] = object;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        }
+        while (list.size() > values.length) {
+            resize();
+        }
+        int j = 0;
+        for (int i = size;j < list.size();i++) {
+            values[i] = list.get(j);
+            j++;
+            size++;
+        }
```
okuzan, line 56:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

okuzan: isEmpty() and size()==0 are the same thing
we don't need to handle this specifically, it it's the case the loop won't run even once. Resize is called inside add()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
     }
 
     @Override
-    public void set(T value, int index) {
+    public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
+        return false;
```
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+    }
+
+    public T get(int index) {
+        if (index < size && index >= 0) {
+            return (T) values[index];
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bound!");
+        }
```
okuzan, line 72: move the check to a separate method which will throw exception, after the check just return `values[index]`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        return -1;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index > size - 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bound!");
+        }
```
okuzan, line 88: use the same checker method as in get

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        removedObject = (T) values[index];
```
okuzan, line 89:
```suggestion
        T removed = values[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        for (int i = index; i < size - 1; i++) {
+            values[i] = values[i + 1];
+        }
```
okuzan, line 92: use `System.arraycopy` instead, as you do on line 102

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
-    public T remove(T element) {
-        return null;
+    public T remove(T value) {
+        int index = indexOf(value);
+        if (index >= 0) {
+            System.arraycopy(values, index + 1, values, index, size - index - 1);
+            size--;
+            values[size] = null;
+            return value;
+        }
```
okuzan, line 106: use the logic of `indexOf` method (iteration + equality check) right here and call remove(i) when the T value is found

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        throw new NoSuchElementException("There is no element like this!");
     }
 
-    @Override
-    public int size() {
-        return 0;
+    public void set(T value, int index) {
+        if (index < 0 || index > size - 1) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bound!");
+        }
```
okuzan, line 113: use the aforementioned checkIndex method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        if (size >= arraySize) {
+            Object[] newValues = new Object[size * 3 / 2 + 1];
+            System.arraycopy(values, 0, newValues, 0, size);
+            values = newValues;
+        }
+        if (size >= DEFAULT_SIZE && size < arraySize / 4) {
+            Object[] newValues = new Object[size * 3 / 2 + 1];
+            System.arraycopy(values, 0, newValues, 0, size);
+            values = newValues;
+        }
```
okuzan, line 127:
```suggestion
       if (size == values.length) {
          T[] data = values;
          values = (T[]) new Object[size * 3 / 2];
          System.arraycopy(data, 0, values, 0, size);
      }
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
polikarpov13: We need removedObject field because we need to return removed(index) object in remove method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        return -1;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isOutOfBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bound!");
+        }
+        removedObject = values[index];
```
okuzan, line 82:
```suggestion
        T removedObject = values[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_SIZE = 10;
+    private T[] values;
+    private int size;
+    private T removedObject;
```
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                return i;
+            }
+        }
+        return -1;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isOutOfBound(index)) {
```
okuzan, line 79: make `checkInBounds` void and throw exception there

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
polikarpov13: Distance between variable 'removedObject' declaration and its first usage is 4, but allowed 3.  
Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value). [VariableDeclarationUsageD
istance]. This is the reason why I can not do this

okuzan: Got it, but making if a class field is not an option, here I missed that we don't need `resize()` at all (java.util.ArrayList doesn't shrink), so that will fix your checkstyle issue, in other cases make it `final` if needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkInBounds(index);
+        removedObject = values[index];
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        values[--size] = null;
+        resize();
```
okuzan, line 79:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
okuzan: make it local variable

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        values = (T[]) new Object[DEFAULT_SIZE];
     }
 
-    @Override
-    public void add(T value, int index) {
+    public ArrayList(int initialCapacity) {
+        values = (T[]) new Object[initialCapacity];
+    }
 
+    public void checkInBounds(int index) {
```
okuzan, line 18: this should be private, put it next to `resize()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        if (index > size - 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bound!");
```
okuzan, line 20: here and in ohter invocations include index and bound in exception to make it more informative

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T value) {
+        int index = indexOf(value);
+        if (index == -1) {
+            throw new NoSuchElementException("There is no element like this!");
+        }
+        checkInBounds(index);
```
nacenik, line 81: When you find the index - use the method `remove(int index)`. This method contains a check index mechanism

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        remove(index);
+        return value;
```
nacenik, line 83:
```suggestion
        return remove(index);
```

nacenik (changes requested):

aleksandr-bevza (approved):

PR #777, opened 2023-01-03
Created my own ArrayList class which implements the List interface. O…

Andrii-First:
…verrided the required methods.

Happy New Year))And safe sky)

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,107 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        arrayCapacityCheck();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        indexCheck(index);
+        arrayCapacityCheck();
+        System.arraycopy(elementData, index, elementData,
+                (index + 1), elementData.length - index - 1);
+        elementData[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size();i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index);
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        indexCheck(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexCheck(index);
+        T temp = elementData[index];
+        System.arraycopy(elementData,(index + 1),
+                elementData, index, (elementData.length - index - 1));
+        size--;
+        return temp;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T temp = null;
+        int count = 0;
+        for (int i = 0; i < size;i++) {
+            if ((element == null && elementData[i] == null)
+                    || (elementData[i] != null && (elementData[i].equals(element)))) {
+                temp = elementData[i];
+                remove(i);
+                count++;
+                break;
+            }
+        }
+        noSuchElementCheck(count);
+        return temp;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void arrayCapacityCheck() {
+        if (size >= elementData.length) {
+            T[] temp = elementData;
+            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
+            System.arraycopy(temp, 0, elementData, 0, temp.length);
+        }
+    }
+
+    private void indexCheck(int index) {
+        if (index < 0 || (index != 0 && index >= size)) {
+            throw new ArrayListIndexOutOfBoundsException();
+        }
+    }
+
+    private void noSuchElementCheck(int count) {
+        if (count == 0) {
+            throw new NoSuchElementException("No such element in current array");
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
+++ b/src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
@@ -1,7 +1,7 @@
 package core.basesyntax;
 
 public class ArrayListIndexOutOfBoundsException extends RuntimeException {
-    public ArrayListIndexOutOfBoundsException(String message) {
-        super(message);
+    public ArrayListIndexOutOfBoundsException() {
+        super("Wrong index value(Out of bound)");
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        arrayCapacityCheck();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        indexCheck(index);
```
karma-o, line 25: what if you increment the size and throw exception after this?
all of the logic should be after the checks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+            if ((element == null && elementData[i] == null)
+                    || (elementData[i] != null && (elementData[i].equals(element)))) {
+                temp = elementData[i];
+                remove(i);
+                count++;
+                break;
+            }
+        }
+        noSuchElementCheck(count);
+        return temp;
```
karma-o, line 75: we don't need this `count` and `temp` vars and `noSuchElementCheck()` method, we can just return the values right in the loop if we found it, and throw an exception after the loop if we did not 
+ `remove(index)` gets you the element, so you can use it
+ we can simplify if conditon
```suggestion
        for (int i = 0; i < size;i++) {
            if ((element == elementData[i] || element != null && element.equals(elementData[i]) {
                return remove(i);
            }
        }
        throw new NoSuchElementException(...);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+
+    private void arrayCapacityCheck() {
+        if (size >= elementData.length) {
+            T[] temp = elementData;
+            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
+            System.arraycopy(temp, 0, elementData, 0, temp.length);
+        }
+    }
+
+    private void indexCheck(int index) {
```
karma-o, line 96: method names should be verbs
```suggestion
    private void ckeckArrayCapacity() {
        if (size >= elementData.length) {
            T[] temp = elementData;
            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
            System.arraycopy(temp, 0, elementData, 0, temp.length);
        }
    }

    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
```diff
@@ -1,7 +1,7 @@
 package core.basesyntax;
 
 public class ArrayListIndexOutOfBoundsException extends RuntimeException {
-    public ArrayListIndexOutOfBoundsException(String message) {
-        super(message);
+    public ArrayListIndexOutOfBoundsException() {
+        super("Wrong index value(Out of bound)");
```
karma-o, line 5: do not hardcode a message, pass it as an argument instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        arrayCapacityCheck();
+        System.arraycopy(elementData, index, elementData,
+                (index + 1), elementData.length - index - 1);
+        elementData[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size();i++) {
```
karma-o, line 34: can be for each here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private void arrayCapacityCheck() {
+        if (size >= elementData.length) {
+            T[] temp = elementData;
+            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
+            System.arraycopy(temp, 0, elementData, 0, temp.length);
+        }
+    }
+
+    private void indexCheck(int index) {
+        if (index < 0 || (index != 0 && index >= size)) {
```
karma-o, line 97:
```suggestion
        if (index < 0 || index >= size) {
```

karma-o (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.stream.IntStream;
+
 public class ArrayList<T> implements List<T> {
+
```
Rommelua, line 7: line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        checkArrayCapacity();
+        elementData[size] = value;
```
Rommelua, line 19: I see you only **check** Capacity here. But checking is not enough - array must grow if it is full.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+        System.arraycopy(elementData, index, elementData,
+                (index + 1), elementData.length - index - 1);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        IntStream.range(0, list.size()).mapToObj(list::get).forEach(this::add);
```
Rommelua, line 35:
```suggestion
        IntStream.range(0, list.size())
        .mapToObj(list::get)
        .forEach(this::add);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex((index + 1));
+        T temp = elementData[index];
+        System.arraycopy(elementData,(index + 1),
+                elementData, index, (elementData.length - index - 1));
```
Rommelua, line 55:
```suggestion
        System.arraycopy(elementData, index + 1, elementData, index, size - index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+    private void checkArrayCapacity() {
+        if (size >= elementData.length) {
+            T[] temp = elementData;
+            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
+            System.arraycopy(temp, 0, elementData, 0, temp.length);
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size) {
```
Rommelua, line 89:
```suggestion
        if (index < 0 || index >= size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     public void set(T value, int index) {
-
+        checkIndex((index + 1));
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex((index + 1));
```
Rommelua, line 52:
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     public T get(int index) {
-        return null;
+        checkIndex((index + 1));
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex((index + 1));
```
Rommelua, line 46: checkIndex(index);

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        IntStream.range(0, list.size()).mapToObj(list::get).forEach(this::add);
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex((index + 1));
```
Rommelua, line 40: checkIndex(index);

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+        if (size >= elementData.length) {
+            T[] temp = elementData;
+            elementData = (T[]) new Object[elementData.length + (elementData.length / 2)];
+            System.arraycopy(temp, 0, elementData, 0, temp.length);
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index value(Out of bound)");
```
Rommelua, line 90: Make your exceptions informative.
If you throw an Exception, you should always add an informative message so that it would be clear from the stack trace what exactly went wrong.

Rommelua (changes requested):

aaananas (approved):

PR #779, opened 2023-01-07
Jv array list solution

isakov-github:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        grow(INITIAL_ARRAY_SIZE / 2);
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("The array index is out of bounds");
+        }
+    }
+
+    private void checkAddIndex(int index) {
```
nacenik, line 34: You don't need this method, cause it is called only one time. Put logic from this method in method where it is called.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("The array index is out of bounds");
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
```
nacenik, line 42: Create an additional method for this code.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int listSize = list.size();
+        if (array.length - size < listSize) {
```
nacenik, line 62: unnecessary if statement

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            grow(listSize);
+        }
+        for (int i = 0; i < listSize; i++) {
+            array[size + i] = list.get(i);
```
nacenik, line 66: just call method `add(T value)` for each element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        if (size + value > MAX_ARRAY_SIZE) {
+            throw new OutOfMemoryError();
+        }
+        Object[] temp = new Object[size + value];
+        System.arraycopy(array, 0, temp, 0, size);
+        array = temp;
+    }
+
+    private void grow() {
+        grow(INITIAL_ARRAY_SIZE / 2);
```
nacenik, line 25: you resize list each time only on 5 element, but you should do it with 1.5 multipluer 
image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private void grow(int value) {
+        if (size + value > MAX_ARRAY_SIZE) {
+            throw new OutOfMemoryError();
+        }
+        Object[] temp = new Object[size + value];
+        System.arraycopy(array, 0, temp, 0, size);
+        array = temp;
+    }
+
+    private void grow() {
```
nacenik, line 24: the private method should be after the public. Fix it for all methods.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            array[size + i] = list.get(i);
+        }
+        size += listSize;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T)array[index];
```
nacenik, line 74:
```suggestion
        return (T) array[index];
```

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private int size;
+    private Object[] array;
+
+    ArrayList() {
+        array = new Object[INITIAL_ARRAY_SIZE];
+    }
+
+    private void grow(int value) {
+        if (size + value > MAX_ARRAY_SIZE) {
+            throw new OutOfMemoryError();
```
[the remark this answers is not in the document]
isakov-github: Ok, but how about java source code?
    private static int hugeCapacity(int minCapacity) {
        if (minCapacity < 0) // overflow
            throw new OutOfMemoryError();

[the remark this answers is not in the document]
isakov-github: :/ I confused an error with an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_ARRAY_SIZE = 10;
+    private int size;
+    private Object[] array;
```
```suggestion
    private T[] array;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
+    public ArrayList() {
+        array = new Object[INITIAL_ARRAY_SIZE];
```
okuzan, line 11:
```suggestion
        array = (T[]) new Object[INITIAL_ARRAY_SIZE];
```

src/main/java/core/basesyntax/ArrayList.java
okuzan, line 8: values would be more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
-
+        checkIndex(index);
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T item = (T) array[index];
```
okuzan, line 53:
```suggestion
        T item = array[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        if (size != index + 1) {
```
okuzan, line 54: check not needed, arraycopy won't work

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null ? array[i] == null : element.equals(array[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Such element does not exist in array");
```
okuzan, line 68: add element to message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size + (size >> 1) > Integer.MAX_VALUE) {
+            throw new ArrayListIndexOutOfBoundsException("The maximum size of the array reached");
```
okuzan, line 83: add max size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        }
+        Object[] newArray = new Object[size + (size >> 1)];
+        System.arraycopy(array, 0, newArray, 0, size);
+        array = newArray;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("The array index is out of bounds");
```
okuzan, line 92: add index + bound

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size + (size >> 1) > Integer.MAX_VALUE) {
+            throw new ArrayListIndexOutOfBoundsException("The maximum size of the array reached");
+        }
+        Object[] newArray = new Object[size + (size >> 1)];
```
```suggestion
        T[] newArray = (T[]) new Object[size + (size >> 1)];
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
isakov-github: Sorry, don't understand. What values?

src/main/java/core/basesyntax/ArrayList.java
isakov-github: this logic don't work if we delete the last element, so check is needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T item = array[index];
+        if (size != index + 1) {
+            System.arraycopy(array, index + 1, array, index, size - index);
+        }
```
aleksandr-bevza, line 56: You can do it like this:
```suggestion
        System.arraycopy(array, index + 1, array, index, size - index - 1);
```

aleksandr-bevza (approved):

PR #785, opened 2023-01-10
jv-array-list implementation

Dmytro-Byrka:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        System.arraycopy(data,0,arrayCopy,0,data.length);
+        data = arrayCopy;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Passed index is invalid");
+        }
+    }
+
```
aaananas, line 34: grow() should be private 
private methods should go after public

aaananas (approved):

src/main/java/core/basesyntax/ArrayList.java
Dmytro-Byrka: Thanks, my bad

PR #786, opened 2023-01-13
first version ArrayList

OksanaC:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        arrayList[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null) {
+            return;
+        }
```
karma-o, line 43: I think this check is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
-
+        checkIndex(index);
+        arrayList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedArrayList = arrayList[index];
```
karma-o, line 64: did we remove the whole list? :)
```suggestion
        final T removedElement = arrayList[index];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY_DEFAULT = 10;
+    private static final double CAPACITY_GROW = 1.5;
+    private int size;
+    private T[] arrayList;
```
karma-o, line 9: names are a bit off
```suggestion
    private static final int CAPACITY_DEFAULT = 10;
    private static final double CAPACITY_GROW_MULTIPLIER = 1.5;
    private int size;
    private T[] data;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < this.size; i++) {
+            if (arrayList[i] == element || arrayList[i] != null && arrayList[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException(" The list is empty");
```
karma-o, line 77: That does not mean our list is empty, only that we don't have such an element
```suggestion
        throw new NoSuchElementException("Element: " + element + "is not present in the list");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
```
karma-o, line 90:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
     }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index is invalid");
```
karma-o, line 95: what exactly index?
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index: " + index + " is invalid");
```

karma-o (changes requested):
good job on the task :1st_place_medal:

karma-o (approved):

PR #790, opened 2023-01-14
Task completed. Class ArrayList and its methods implemented

Oleh8:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        }
         return null;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < actualSize; i++) {
+            if (arrayOfElements[i] == element || arrayOfElements[i] != null
+                    && arrayOfElements[i].equals(element)) {
```
karma-o, line 78: just to be sure :)
```suggestion
            if (arrayOfElements[i] == element || element != null
                    && element.equals(arrayOfElements[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        return true;
+    }
+
+    private void increaseCapacity() {
+        T[] temporaryArray = (T[]) new Object[(int) (arrayOfElements.length * INCREASE_RATE)];
+        System.arraycopy(arrayOfElements, 0, temporaryArray, 0, actualSize);
+        arrayOfElements = temporaryArray;
+    }
+
+    private T removeElementAndResize(int index) {
```
karma-o, line 108: as said earlier, you are not actually changing the size of the array, so the naming is a bit off.
Also, this method is an exact duplicate of `remove(index)`, and is not used anywhere, so remove it

karma-o (approved):
Approved assuming you will fix the comments

PR #793, opened 2023-01-18
Overrode all methods

OleksiiKlymenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
+    public static final int DEFAULT_SIZE = 10;
+    public static final double INCREASE = 1.5;
```
```suggestion
    public static final double GROW_FACTOR = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
```
okuzan, line 6: delete line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            increaseArray();
+        }
```
okuzan, line 20: move this check inside increaseArray() if it's only ever called on this condition

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            return;
+        }
```
okuzan, line 43: it's better to let it crash with NPE than silently return, addAll(null) isn't some normal expected behavior to handle

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
```
okuzan, line 30: here we need a different check,   with this condition: if (index < 0 || index > size) {

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void increaseArray() {
```
```suggestion
    private void growArray() {
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
OleksiiKlymenko: If honestly I do not understand why we need it. 
checkIndex checks index and then works.

if I change "(index < 0 || index > size)" tests don't pass

In checkIndex I have "index < 0 || index >= size" and it works
Why is it wrong?

andriy-shulepa: `        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException(YOUR_MESSAGE);
        }
// other logic
`

It should work in such way

OleksiiKlymenko: Oh, thank you, got it
I've thought I need to add it in another block of code))

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is invalid");
+        }
+        if (elementData.length == size) {
```
okuzan, line 31: you already check this inside the method, delete this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+            growArray();
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
```
okuzan, line 41: it will crash with NPE inside for loop when .size() method will be called on null, so this doesn't add much tbh, delete the check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        if (size == elementData.length) {
+            T[] newArray = (T[]) new Object[(int) (elementData.length * GROW_FACTOR)];
+            System.arraycopy(elementData, 0, newArray, 0, elementData.length);
+            elementData = newArray;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is invalid");
```
okuzan, line 102: add index and size

okuzan (changes requested):

okuzan (approved):
Good job!

PR #796, opened 2023-01-21
Realized class ArrayList and implemented interface List. Also realize…

Vlad1507:
…d methods checkIndex(), checkRangeIndexForAdd, overloaded method grow() and implemented equals().

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,126 @@
 package core.basesyntax;
 
-public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+import java.util.Arrays;
+import java.util.NoSuchElementException;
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
+                counter++;
+            }
+        }
+        if (counter > 0) {
+            int index = Arrays.asList(elementData).indexOf(element);
+            remove(index);
+        } else {
+            throw new NoSuchElementException("There is no such element in list!");
+        }
+        return element;
```
aaananas, line 86:
```suggestion
        for (Object el : elementData) {
            if (equals(el, element)) {
                return remove(index);
            }
        }
        throw new NoSuchElementException("There is no such element in list!");
```

aaananas (changes requested):

aaananas (approved):

PR #798, opened 2023-01-24
Override methods of ArrayList

katerynapoli:


Ivan95kos (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        return removedElement;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(EXCEPTION_MESSAGE);
+        }
+    }
+
+    private void checkSizeEquality() {
```
okuzan, line 110:
```suggestion
    private void growIfNeeded() {
```

okuzan: may be other name, but it should reflect the fact that you do something here, not just check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        for (int i = index; i < size - 1; i++) {
+            array[i] = array [i + 1];
+        }
+        size--;
+        return removedElement;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(EXCEPTION_MESSAGE);
```
okuzan, line 106: include index, size in the method. If you opt to use constant for message, you can create template and use String.format()

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-
+        growIfSizeEquals();
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(EXCEPTION_MESSAGE);
```
andriy-shulepa, line 31: format message template with actual values

andriy-shulepa (changes requested):

andriy-shulepa (approved):

PR #799, opened 2023-01-25
Added solution to the task

ivbudenkov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double DEFAULT_GROW_RATE = 1.5;
```
```suggestion
    public static final double GROW_RATE = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    private Object[] values;
```
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    private int size;
+
+    public ArrayList() {
+        values = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
okuzan, line 14: it's 0 by default, this isn't necessary

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double DEFAULT_GROW_RATE = 1.5;
+    private Object[] values;
+    private int size;
+
+    public ArrayList() {
+        values = new Object[DEFAULT_CAPACITY];
```
```suggestion
        values =  (T[])  new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
okuzan: capacities can be different - that's why we have a default one, but there ever will be only one grow_rate

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    }
+
+    private void checkSizeEnlargeCapacity(int sizeIncrement) {
+        if (size + sizeIncrement > values.length) {
+            int newSize = values.length;
+            while (newSize < size + sizeIncrement) {
+                newSize = (int) (newSize * DEFAULT_GROW_RATE);
+            }
+            values = Arrays.copyOf(values,
+                    (int) (this.values.length * DEFAULT_GROW_RATE));
```
okuzan, line 30: there is a mess with array copying here. It's only needed in these methods: add(T value, int index), remove(index), and method responsible for resizing. Try to use only one mechanism everywhere, for example, one System.arraycopy call will be enough in each case

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        size = 0;
+    }
+
+    private void copyValues(int start, int end, int newStart) {
+        Object[] copy = new Object[end - start + 1];
+        System.arraycopy(values, start, copy, 0, copy.length);
+        System.arraycopy(copy, 0, values, newStart, copy.length);
+    }
+
+    private void checkSizeEnlargeCapacity(int sizeIncrement) {
```
okuzan, line 23: it's weird that this checker method has an argument, size == values.length condition is enough and can be calculated by accessing fields directly

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double GROW_RATE = 1.5;
+    private T[] values;
+    private int size;
+
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    private void copyValues(int start, int end, int newStart) {
```
nacenik, line 15: private methods should be after public. Fix for all

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        return values[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    String.format("Index %d out of bounds for length %d", index, size));
+        }
```
nacenik, line 82: `checkIndex`

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
-
+        checkIndex(index, size);
+        if (index == size) {
+            add(value);
+        } else {
+            checkSizeEnlargeCapacity();
+            copyValues(index, size - 1, index + 1);
+            values[index] = value;
+            size++;
+        }
```
okuzan, line 31:
```suggestion
        checkIndex(index, size);
        copyValues(index, size - 1, index + 1);
        values[index] = value;
        size++;
```

okuzan (changes requested):

okuzan (approved):

PR #800, opened 2023-01-25
realisation class ArrayList

bigvini:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int DEFAULT_SIZE = 0;
+    private static final int START_INDEX = 0;
+    private static final int NEGATIVE_INDEX = -1;
+    private static final int ADD_INDEX_NUMBER = 1;
+    private int capacity;
+    private T[] listT;
```
okuzan, line 12: bad naming, go with `values`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int DEFAULT_SIZE = 0;
+    private static final int START_INDEX = 0;
+    private static final int NEGATIVE_INDEX = -1;
+    private static final int ADD_INDEX_NUMBER = 1;
+    private int capacity;
```
okuzan, line 11: from all of these leave only DEFAULT_CAPACITY = 10;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    private T[] listT;
+    private int size;
+
+    public ArrayList() {
+        this.capacity = DEFAULT_CAPACITY;
+        this.listT = (T[]) new Object[capacity];
+        size = DEFAULT_SIZE;
```
```suggestion
        this.values = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == DEFAULT_SIZE) {
+            return true;
+        }
```
okuzan, line 94: check if size == 0, that's enough

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+
     @Override
     public void add(T value) {
-
+        if (capacity > size) {
+            listT[size] = value;
+            size++;
+            return;
+        }
+        increaseCapacity();
```
okuzan, line 28: Everywhere: instead of having if-else based on comparing capacity to size, use this:
        if (size == values.length) {
            increaseCapactiy();
        }
        --rest of the logic here--

```suggestion
        if (size == values.length) { 
             increaseCapacity();
        }
        values[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        listT[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (capacity > size) {
```
okuzan, line 35: here check index first: 
        `if (index < 0 || index > size) { throw new ArrayListIndexOutOfBoundsException(message)`
        then just paste logic of addToIndex() here. We won't need resizing handling, because new array will be created by System.arraycopy anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        }
+        fullCopyArray(newList);
+        newArrayListIndexOutOfBoundsException(index);
+    }
+
+    private boolean isEquals(T firstKey, T secondKey) {
+        return firstKey == secondKey || (firstKey != null && firstKey.equals(secondKey));
+    }
+
+    private void newArrayListIndexOutOfBoundsException(int index) {
```
okuzan, line 122: methods should be mostly verbs. This sounds like the name of the exception class. Call this method like `checkIndexInBounds`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        if (index >= size || index < DEFAULT_SIZE) {
+            throw new ArrayListIndexOutOfBoundsException("There is no such index");
+        }
+    }
+
+    private void fullCopyArray(T[] newList) {
+        System.arraycopy(newList, START_INDEX, listT, START_INDEX, size);
```
okuzan, line 129: it's okay to use zeros here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        return listT[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < size && index >= DEFAULT_SIZE) {
+            listT[index] = value;
+        }
+        newArrayListIndexOutOfBoundsException(index);
```
okuzan, line 61: call checker methods first thing in the method, then goes other logic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
-        return null;
+        newArrayListIndexOutOfBoundsException(index);
+        return listT[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < size && index >= DEFAULT_SIZE) {
+            listT[index] = value;
```
okuzan, line 59: your method will check index, no need for this check, extract body of if and use it directly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        }
         return false;
     }
+
+    private void increaseCapacity() {
+        capacity = capacity + (capacity >> ADD_INDEX_NUMBER);
+        T[] newList = (T[]) new Object[capacity];
+        System.arraycopy(listT, START_INDEX, newList, START_INDEX, listT.length);
+        this.listT = (T[]) new Object[capacity];
+        fullCopyArray(newList);
```
okuzan, line 103: you call system.arraycopy twice here, its costly operation and it's not necessary

okuzan, line 103: T[] data = values;
        values = (T[]) new Object[size * GROW_FACTOR];

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    private void addToIndex(T value, int index) {
+        T[] newList = (T[]) new Object[listT.length];
+        if (index <= size && index > NEGATIVE_INDEX) {
+            System.arraycopy(listT, START_INDEX, newList, START_INDEX, index + ADD_INDEX_NUMBER);
+            newList[index] = value;
+            System.arraycopy(listT, index, newList, index + ADD_INDEX_NUMBER, size - index);
+            size++;
+        }
+        fullCopyArray(newList);
+        newArrayListIndexOutOfBoundsException(index);
```
okuzan, line 115: if `if` is true you call arraycopy thrice here. You need to do it once. Don't check the index twice - you already checked it where the method was called

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    }
+
+    private boolean isEquals(T firstKey, T secondKey) {
+        return firstKey == secondKey || (firstKey != null && firstKey.equals(secondKey));
+    }
+
+    private void newArrayListIndexOutOfBoundsException(int index) {
+        if (index >= size || index < DEFAULT_SIZE) {
+            throw new ArrayListIndexOutOfBoundsException("There is no such index");
```
okuzan, line 124: include index and size in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+            if (isEquals(listT[i], element)) {
+                size--;
+                System.arraycopy(listT, i + ADD_INDEX_NUMBER, newList, i, size - i);
+                fullCopyArray(newList);
+                return element;
+            } else {
+                newList[i] = listT[i];
+            }
+        }
+        throw new NoSuchElementException("There is no such element");
```
okuzan, line 82: include the element in the message

okuzan (changes requested):

andriy-shulepa (approved):

PR #806, opened 2023-02-01
ArrayList is realized

YaroslavKostenko89:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    public ArrayList() {
+        this.arrayOfElements = (T[]) new Object[maxSize];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == maxSize) {
+            grow();
+        }
```
Ivan95kos, line 21: move this check to a method grow()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            if (size == maxSize) {
+                grow();
+            }
+            this.arrayOfElements[size] = list.get(i);
+            size++;
+        }
```
Ivan95kos, line 45: Here it will be enough to use the already existing method
```suggestion
        for (int i = 0; i < list.size(); i++) {
              add(list.get(i));
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        indexExist(index);
+        size--;
+        T oldValue = arrayOfElements[index];
+        if (size > index) {
+            System.arraycopy(arrayOfElements, index + 1, arrayOfElements, index, size - index);
+        }
+        arrayOfElements[size] = null;
```
Ivan95kos, line 68:
```suggestion
        T oldValue = arrayOfElements[index];
        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
        size--;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int i = 0;
+        for (; i < size; i++) {
```
Ivan95kos, line 75:
```suggestion
        for (int i = 0; i < size; i++) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int GROW = 2;
+    private T[] arrayOfElements;
```
Ivan95kos, line 9:
```suggestion
    private T[] elements;
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int GROW = 2;
```
nacenik, line 8: image
```suggestion
    private static final int GROW_FACTOR = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private T[] elements;
+    private int maxSize = 10;
```
nacenik, line 10: [Make it constant](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#any-magic-numbers-should-be-constants)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size == maxSize) {
+            maxSize = maxSize + maxSize / GROW;
+            elements = Arrays.copyOf(elements, maxSize);
+        }
+    }
```
nacenik, line 87: `if` should be in another method cause grow name mean that size should change every time 
```suggestion
    private void checkSize() {
        if (size == maxSize) {
            grow();
        }
    }
    
    private void grow() {
        maxSize = maxSize + maxSize / GROW;
        elements = Arrays.copyOf(elements, maxSize);
    }
```

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
-
+        checkSize();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        rangeCheckForAdd(index);
```
okuzan, line 27: not really necessary to extract logic from the method if it's used only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    }
+
+    private void checkSize() {
+        if (size == maxSize) {
+            grow();
+        }
+    }
+
+    private void indexExist(int index) {
+        if (index >= size || index < 0) {
```
okuzan, line 95:
```suggestion
        if (index < 0  || index >= size) {
```

okuzan: just to make two methods look similar :)

okuzan (approved):
Good job!

PR #812, opened 2023-02-07
hw-jv-array-list

ihor-sydorenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resizeArray() {
```
Rommelua, line 89: It should be clear from the name that you do not resize always.
```suggestion
    private void resizeIfNeeded() {
```

Rommelua (approved):

PR #813, opened 2023-02-08
solved task

slaybrute:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private int indexOfLastAddedElement = 0;
+    private Object[] array = new Object[10];
```
nacenik, line 9: [See common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private int indexOfLastAddedElement = 0;
```
nacenik, line 8: 0 is default value for int 
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    private Object[] array = new Object[10];
+
     @Override
     public void add(T value) {
-
+        if (indexOfLastAddedElement >= array.length) {
+            array = Arrays.copyOf(array, (int) (array.length * 1.5));
```
nacenik, line 14: [You need to use `System.arraycopy()`](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements). Fix it in all places
extract logic that inside if to `grow()` method. Reuse it everywhere

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private int indexOfLastAddedElement = 0;
+    private Object[] array = new Object[10];
+
     @Override
     public void add(T value) {
-
+        if (indexOfLastAddedElement >= array.length) {
```
nacenik, line 13: if should be extracted to the method `checkSIze()` or kinda this and reuse it in all places

src/main/java/core/basesyntax/ArrayList.java
```suggestion
    private T[] array;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public int size() {
-        return 0;
+        return indexOfLastAddedElement;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return indexOfLastAddedElement <= 0;
```
nacenik, line 107: size cannot be less then 0

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+                oldValue = (T) array[i];
+                remove(i);
+                counter++;
+                break;
+            }
+        }
+        if (counter == 0) {
+            throw new NoSuchElementException("no such element in List");
+        }
+        return oldValue;
```
nacenik, line 97: in this method you need to find an index of the element and call the method `remove(int index)`, otherwise - throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (indexOfLastAddedElement <= index || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: "
+                    + index + ", Size: " + indexOfLastAddedElement);
+        }
```
nacenik, line 72: extract this logic to the method and reuse it where you need

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T oldValue = null;
+        int counter = 0;
+        for (int i = 0; i < array.length; i++) {
+            if (Objects.equals(array[i], element)) {
```
nacenik, line 87: don't use Objects class in your solution

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
slaybrute: In this method I've found the the index and called the method remove(int index)

src/main/java/core/basesyntax/ArrayList.java
slaybrute: It's impossible to initialize an array with a given type T

okuzan: You initialize array of objects, then cast it to T

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void callArrayListIndexOutOfBoundsExceptionIfNeeded() {
```
okuzan, line 112:
```suggestion
    private void checkIndexInBounds() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_ARRAY_SIZE = 10;
```
```suggestion
    private static final int DEFAULT_SIZE = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+    private int size = 0;
+    private int index;
+    private Object[] array = new Object[START_ARRAY_SIZE];
```
okuzan, line 10:
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+
```
okuzan, line 11: create constructor and initialize `values` with an array of objects of DEFAULT_SIZE, cast it to T[]

src/main/java/core/basesyntax/ArrayList.java
okuzan: not fixed, `Arrays.copyOf` is not allowed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     @Override
     public void add(T value) {
-
+        if (size >= array.length) {
+            array = Arrays.copyOf(array, (int) (array.length * 1.5));
```
okuzan, line 15: 1.5 should be GROW_FACTOR constant, for example

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+        }
+        array[size] = value;
+        size++;
```
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: "
+                    + index + ", Size: " + size);
+        } else {
```
okuzan, line 26: we don't need `else` clause, throwing means exiting the method, unwrap the logic inside of it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() >= array.length) {
+            array = Arrays.copyOf(array, (int) (array.length * 1.5));
+        }
+        for (int i = size, q = 0; q < list.size(); i++, q++) {
+            array[i] = list.get(q);
+            size++;
+        }
```
okuzan, line 45:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }

```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_ARRAY_SIZE = 10;
+    private int size = 0;
```
okuzan, line 8: it's useless to initialize int with 0, it's 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+    private int index;
```
okuzan, line 9: we don't need this field

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
-        return null;
+        this.index = index;
+        callArrayListIndexOutOfBoundsExceptionIfNeeded();
+        return (T) array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        this.index = index;
```
okuzan, line 57: delete this line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        this.index = index;
+        callArrayListIndexOutOfBoundsExceptionIfNeeded();
+        T oldValue = (T) array[index];
+        Object[] secondPart = Arrays.copyOfRange(array, index + 1, size);
+        System.arraycopy(secondPart, 0, array, index, secondPart.length);
+        size--;
+        return oldValue;
```
okuzan, line 70:
```suggestion
        checkIndexInBounds();
        T oldValue = values[index];
        System.arraycopy(values, index + 1, values, index, size - index - 1);
        values[--size] = null;
        return oldValue;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+                    remove(i);
+                    counter++;
+                    break;
+                }
+            }
+        }
+        if (counter == 0) {
+            throw new NoSuchElementException("no such element in List");
+        }
+        return oldValue;
```
okuzan, line 99: I understand why you created `counter`, but we can do it in a more elegant way. instead of all of the logic inside the for-loop, you need one if-check, that will take care of edge cases like the element being null; inside of the if, return `remove(i)`; After the loop, throw an exception right away, without any checks - we wouldn't reach that part if the value was found, because we would `return`.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index: "
+                    + index + ", Size: " + size);
+        } else {
+            if (size == array.length) {
+                array = Arrays.copyOf(array, (int) (array.length * 1.5));
+            }
+            Object[] secondPart = Arrays.copyOfRange(array, index, size);
+            array[index] = value;
+            System.arraycopy(secondPart, 0, array, index + 1, secondPart.length);
+            size++;
```
okuzan, line 33: instead of all of this, you just need one System.arraycopy call, then lines 31,33

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
slaybrute: Why Arrays.copyOf is not allowed. I reckon It's the best solution. Because if we here apply System.arraycopy, initially we should create new array with new size, then copy old to new one, then again copy to main array. It's too much extra code, don't you think ?

slaybrute: I reckon that in some cases System.arraycopy is more logically, I have applied it when it reduce code, but in this case it add extra lines of code

src/main/java/core/basesyntax/ArrayList.java
slaybrute: if you mean field index, it's important, because we use it in this separate method   "" private void checkIndexInBounds() ""
otherwise it's impossible to use it in method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    private int index;
+    private final Object[] array1 = new Object[DEFAULT_SIZE];
+    private T[] values = (T[]) array1;
+
     @Override
     public void add(T value) {
-
+        if (size >= values.length) {
+            values = Arrays.copyOf(values, (int) (values.length * 1.5));
+        }
```
Ivan95kos, line 17: create method grow()
Instead of 1.5, use a constant
[You need to use System.arraycopy()](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements). Fix it in all places
extract logic that inside if to grow() method. Reuse it everywhere

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: If you need to pass it as a parameter checkIndexInBounds(int index)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexInBounds() {
+        if (size <= index || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: "
+                    + index + ", Size: " + size);
+        }
```
Ivan95kos, line 102:
```suggestion
    private void checkIndexInBounds(int index) {
        if (size <= index || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Index: "
                    + index + ", Size: " + size);
        }
```

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+                if (values[i] != null) {
+                    if (values[i].equals(element)) {
+                        oldValue = values[i];
+                        remove(i);
+                        return oldValue;
+                    }
+                }
+            }
+        }
+        throw new NoSuchElementException("no such element in List");
```
Ivan95kos, line 85:
```suggestion
        for (int i = 0; i < size; i++) {
            if (values[i] == element
                    || values[i] != null && values[i].equals(element)) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("There is no such element " + element);
```

Ivan95kos (changes requested):
Please pay attention to [common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html)
And fix all previous comments

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+                    + index + ", Size: " + size);
+        }
+    }
+
+    private void grow() {
+        T[] tmp = (T[]) new Object[values.length];
+        System.arraycopy(values, 0, tmp, 0, tmp.length);
+        values = (T[]) new Object[(int) (values.length * 1.5)];
+        System.arraycopy(tmp, 0, values, 0, tmp.length);
     }
```
Ivan95kos, line 97:
```suggestion
    private void grow() {
        if (values.length == size) {
            T[] newArray = (T[]) new Object[(int) (values.length * GROW_FACTOR)];
            System.arraycopy(values, 0, newArray, 0, size);
            values = newArray;
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+    public ArrayList() {
+        this.values = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= values.length) {
+            grow();
+        }
```
Ivan95kos, line 18:
```suggestion
            grow();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: "
+                    + index + ", Size: " + size);
+        }
+        if (size == values.length) {
+            grow();
+        }
```
Ivan95kos, line 30:
```suggestion
            grow();
```

Ivan95kos (approved):
Great job

PR #826, opened 2023-02-16
implemented ArrayList class

outref:


Dominskyi93 (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.list = new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
```
sarakhmen, line 17: You are coping arrays each time this method is called. Let's find another approach so that you're just adding a new value at the end of the array instead.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private void grow() {
+        int newCapacity = (list.length + list.length / 2);
+        Object[] newArray = new Object[newCapacity];
+        System.arraycopy(list, 0, newArray, 0, list.length);
+        list = newArray;
+    }
+
+    public int indexOf(Object obj) {
+        for (int i = 0; i < list.length; i++) {
+            if (Objects.equals(obj, list[i])) {
```
sarakhmen, line 88: As described in Common Mistakes, let's not use Objects class.

sarakhmen (changes requested):
Good job, but let's improve your solution :) Also let's change the ordering of your public and private methods at the end of a class (public ones should be preceded)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int INITIAL_CAPACITY = 10;
+    private Object[] list;
```
nacenik, line 7: Create generic array instead of Object array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public int indexOf(Object obj) {
```
nacenik, line 74:
```suggestion
    public int indexOf(T obj) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
+    private void checkSize() {
+        if (size == list.length) {
+            grow();
+        }
+    }
+
+    private void grow() {
+        int newCapacity = (list.length + list.length / 2);
+        Object[] newArray = new Object[newCapacity];
```
nacenik, line 91: create generic array

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) list[index];
```
Andriy-dz, line 42: you don't need to cast directly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    }
+
+    public int indexOf(T obj) {
+        for (int i = 0; i < list.length; i++) {
+            if (list[i] == null ? obj == null : list[i].equals(obj)) {
+                return i;
+            }
+        }
+        throw new NoSuchElementException("Element no found");
+    }
```
Andriy-dz, line 81: you don't need this method because you only use it once
+ is not a valid access modifier

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+            grow();
+        }
+    }
+
+    private void grow() {
+        int newCapacity = (list.length + list.length / 2);
+        T[] newArray = (T[]) new Object[newCapacity];
+        System.arraycopy(list, 0, newArray, 0, list.length);
+        list = newArray;
+    }
```
Andriy-dz, line 94: you can combine these methods and call it grow

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
```
Andriy-dz, line 98: not informative message.
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is invalid);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        return size == 0;
+    }
+
+    public int indexOf(T obj) {
+        for (int i = 0; i < list.length; i++) {
+            if (list[i] == null ? obj == null : list[i].equals(obj)) {
+                return i;
+            }
+        }
+        throw new NoSuchElementException("Element no found");
```
Andriy-dz, line 80:
```suggestion
        throw new NoSuchElementException("Element " + element + " no found");
```

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #832, opened 2023-02-17
create class ArrayList<T>

MaksymHorielikov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    @SuppressWarnings("unchecked")
     public void add(T value, int index) {
-
+        if (index >= 0 && index <= size) {
+            if (size == currentCapacity) {
+                currentCapacity = currentCapacity + (currentCapacity >> 1);
+                T[] temp = (T[]) new Object[currentCapacity];
+                System.arraycopy(listArray, 0, temp, 0, index);
+                System.arraycopy(listArray, index, temp, index + 1, size - index);
+                listArray = temp;
```
IlkoGood, line 31: move lines 27, 28, 29, 31 to separate method
change line 30 so it will copy array in itself :
System.arraycopy(listArray, index, listArray, index + 1, size - index);

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= 0 && index < size) {
+            return listArray[index];
+        }
+        throw new ArrayListIndexOutOfBoundsException("Can't get to index " + index);
```
IlkoGood, line 59: You need to create separate method for checking indexes and throwing an exception and not repeat code in get() and set().
Something like checkIndex() in first line that will throw exception and rest of the code in next line.

Also it would be nice to generalize all index checks in all methods if possible (this, however is completely up to you).

IlkoGood (changes requested):
Good job averall, but ome improvements needed.

src/main/java/core/basesyntax/ArrayList.java
MaksymHorielikov: 1. Lines 27 - 31 are unique, they are not repeated anywhere - therefore they do not need to be moved to a separate method.
My opinion: an additional method always requires copying its parameters, and when used in bulk, this will lead to a performance drop.

2. The proposal to change the 30th line is not correct at all, the program will not work this way. The third parameter is the array to which the data is copied, in our partner it should be a temporary array.

IlkoGood (approved):
Well done!

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+            resizeListArray(index);
+        } else {
+            System.arraycopy(listArray, index, listArray, index + 1, size - index);
+        }
+        listArray[index] = value;
+        size++;
+    }
 
+    private void checkIndex(int index, String message) {
+        if (!(index >= 0 && index < size)) {
```
karma-o, line 39: can we simplify the condition?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    }
+
+    private void checkIndexInclusiveSize(int index, String message) {
+        if (!(index >= 0 && index <= size)) {
+            throw new ArrayListIndexOutOfBoundsException(message);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private void resizeListArray(int index) {
```
karma-o, line 51: private after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        size++;
+    }
 
+    private void checkIndex(int index, String message) {
+        if (!(index >= 0 && index < size)) {
+            throw new ArrayListIndexOutOfBoundsException(message);
+        }
+    }
+
+    private void checkIndexInclusiveSize(int index, String message) {
```
karma-o, line 44: Petro was talking about not passing the messages as method parameters, this is a bad practice, instead, we should hardcode them, and instead of creating several methods for each type of message that you have got, I'd just suggest changing it to a single more general thing like `invalid index: + index`

karma-o (changes requested):

sarakhmen (approved):
Good job :)

PR #837, opened 2023-02-18
Hw-list

KosMarch:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
```
IlkoGood, line 7: Clear naming is important. This is default value and not representing current capacity, so DEFAULT_CAPACITY suits better.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        }
+        list[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
```
IlkoGood, line 35: Naming again, despite our code works well, names shoudln't be ambiguous. Please rename either ArrayList field or incoming parameter.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        size--;
+        System.arraycopy(list, index + 1, list, index, size - index);
+        return element;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, list[i])) {
```
IlkoGood, line 63: We are not allowed to use util.Objects. Provide your custom check.

IlkoGood (changes requested):
Good job, but need some minor improvements.

IlkoGood (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
+    private int size;
+
+    ArrayList() {
```
nacenik, line 10: Add access modifier

src/main/java/core/basesyntax/List.java
```diff
@@ -18,4 +18,6 @@
     int size();
 
     boolean isEmpty();
+
+    int indexOf(T element);
```
nacenik, line 22: don't change interface

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    @Override
+    public int indexOf(T element) {
```
nacenik, line 80: it should be private method

src/main/java/core/basesyntax/ArrayList.java
nacenik: This is absolutely necessary. In java it is the best practice to use interface instead of concrete implementation.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
+
+    private void grow() {
+        if (size == arrayList.length) {
+            int newCapacity = arrayList.length + (arrayList.length >> 1);
+            Object[] newList = new Object[newCapacity];
+            System.arraycopy(arrayList, 0, newList, 0, size);
+            arrayList = (T[]) newList;
```
nacenik, line 112: This part should be in the method that's called `grow()`, cause name grow means that it should increase size every time it is called

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        return -1;
+    }
+
+    private void indexCheck(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
+
+    private void grow() {
```
nacenik, line 107: this method better to rename as _checkSize_ or kind of this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            }
+        } else {
+            for (int i = 0; i < end; i++) {
+                if (element.equals(arrayList[i])) {
+                    return i;
+                }
+            }
+        }
+        return -1;
+    }
```
nacenik, line 99: you don't need to put `size` as parameter.
```suggestion
    private int indexOf(T element) {
        if (element == null) {
            for (int i = 0; i < size; i++) {
                if (arrayList[i] == null) {
                    return i;
                }
            }
        } else {
            for (int i = 0; i < size; i++) {
                if (element.equals(arrayList[i])) {
                    return i;
                }
            }
        }
        return -1;
    }
```

nacenik (changes requested):

Andriy-dz (approved):

PR #838, opened 2023-02-18
initial commit

imhatep:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int ONE = 1;
+    private static final int ZERO = 0;
```
KosMarch, line 9: Do we really need those constants?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == values.length) {
+            grow();
+        }
```
KosMarch, line 21: this code is repeated, it may be better to add it to the grow() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+        values[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        if (size == values.length) {
+            grow();
+        }
```
KosMarch, line 30: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(values[i], element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found in ArrayList");
+    }
+
+    public void grow() {
```
KosMarch, line 86: I think it is better to make this method private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+    private void checkIndex(int index) {
+        if (index >= size || index < ZERO) {
+            throw new ArrayListIndexOutOfBoundsException("Can't find element by index");
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index > size || index < ZERO) {
+            throw new ArrayListIndexOutOfBoundsException("Index Out Of Bounds");
+        }
```
KosMarch, line 112: these two methods do almost the same thing, it may be better to use one with the message ArrayListIndexOutOfBoundsException("Incorrect index: " + index)

KosMarch (changes requested):
Great job!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
+
+    private void checkIndexWithSameSize(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
```
KosMarch, line 106: since you're using the checkIndex() method 
method once, you can use checkIndexWithSameSize() with a certain condition and then you don't need two methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
+        grow();
+        System.arraycopy(values, index,
+                values, index + 1,
+                size - index);
```
KosMarch, line 26:
```suggestion
       grow();
       if (index != size) {
            checkIndexWithSameSize(index);
            System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
        }
```

KosMarch (approved):
I would recommend changing two little things and I think the code is great

src/main/java/core/basesyntax/ArrayList.java
imhatep: thx

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        }
+        throw new NoSuchElementException("Element not found in ArrayList");
+    }
+
+    private void grow() {
+        if (size == values.length) {
+            T[] temp = (T[]) new Object[values.length];
+            System.arraycopy(values, 0, temp, 0, size);
+            values = (T[]) new Object[values.length << 1];
+            System.arraycopy(temp, 0, values, 0, size);
```
sarakhmen, line 82: Could be simplified to only one arraycopy call. Also pay attention, that your array should grow 1.5 times (not 2 times) when it is full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return size == 0;
+    }
+
+    private void checkIndexWithSameSize(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+    }
+
+    private boolean customEquals(Object a, Object b) {
```
sarakhmen, line 102:
```suggestion
    private boolean customEquals(T a, T b) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexWithSameSize(index);
+        final T removedElement = values[index];
+        System.arraycopy(values, index + 1,
+                values, index,
+                values.length - index - 1);
+        values[values.length - 1] = null;
```
sarakhmen, line 62: I think you wanted to do the next
```suggestion
        values[--size] = null;
```
However, it's not necessary

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     public void addAll(List<T> list) {
-
+        T[] a = (T[]) new Object[list.size()];
+        for (int i = 0; i < list.size(); i++) {
+            a[i] = list.get(i);
+        }
+        int varLength = a.length;
+        grow();
+        System.arraycopy(a,0,values,size, varLength);
+        size += varLength;
```
sarakhmen, line 40: A lot of redundant code. Let's simplify it. I'd recommend iterating throughout a list and add new elements using your add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (customEquals(element, values[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found in ArrayList");
+    }
+
+    private void grow() {
```
sarakhmen, line 77: This method has unclear responsibility: it grows only if size == values.length. Let's create one more method, which checks size value and calls grow() if needed

sarakhmen (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        isNeadToResize();
+        values[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        isNeadToResize();
+        if (index != size) {
+            checkIndexWithSameSize(index);
```
sarakhmen, line 24: This check is better to move above resizing method, so that storage is not grown, if index is incorrect

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (isEquals(element, values[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found in ArrayList");
+    }
+
+    private void grow() {
```
sarakhmen, line 71: Let's place private methods after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void isNeadToResize() {
```
sarakhmen, line 88: I'd recommend renaming it to 'ensureCapacity'

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        if (size == values.length || (size - 1) == values.length) {
+            grow();
+        }
+    }
+
+    private void checkIndexWithSameSize(int index) {
```
sarakhmen, line 94: I'd recommend renaming it to simple 'checkIndex' or 'validateIndex'

sarakhmen (approved):
Good job :)

src/main/java/core/basesyntax/ArrayList.java
imhatep: thanks for all recommendation

PR #840, opened 2023-02-19
implemented required functionality in ArrayList class

Skrudj:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        }
+        for (int i = 0; i < elementIndex; i++) {
+            newValues[i] = values[i];
+        }
+        for (int i = elementIndex + 1; i < size; i++) {
+            newValues[i - 1] = values[i];
+        }
+        values = newValues;
+        size--;
+        return element;
```
VitaliiPotseluiko, line 116: There are too much extra lines))))
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == values[i] || (values[i] != null && values[i].equals(element))) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("There is no such element: " + element);
```
It wil be enough)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        } else if (index == size) {
+            add(value);
+        } else {
+            if (values.length == size) {
+                grow();
+            }
+            T[] newValues = (T[]) new Object[values.length];
+            for (int i = 0; i < index; i++) {
+                newValues[i] = values[i];
+            }
```
VitaliiPotseluiko, line 38: Use System.arraycopy method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("No such element exists");
+        } else if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get element by negative index");
+        }
```
VitaliiPotseluiko, line 61: Make checks shorter and put these to separate mothod. for example:
```suggestion
private void checkIndex(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("index " + index
                    + " is out of bound for length " + size);
        } 
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("No such index exists");
+        } else if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can't set to negative position");
+        }
```
VitaliiPotseluiko, line 71: double code. See previus comment.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can't remove by negative index");
+        } else if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("No such element exists");
+        }
```
VitaliiPotseluiko, line 81: double code. See previus comment.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        T[] newValues = (T[]) new Object[values.length];
+        for (int i = 0; i < index; i++) {
+            newValues[i] = values[i];
+        }
```
VitaliiPotseluiko, line 85: Use System.arraycopy() method.

VitaliiPotseluiko (changes requested):

src/main/java/core/basesyntax/ArrayList.java
VitaliiPotseluiko: What about this "Use System.arraycopy() method"?

src/main/java/core/basesyntax/ArrayList.java
VitaliiPotseluiko: What about this :Use System.arraycopy() method?

VitaliiPotseluiko (approved):

VitaliiPotseluiko (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    private int size;
+
+    public ArrayList() {
+        values = (T[]) new Object[START_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        addFreeSpace(values.length);
```
KseniiaMakarova, line 17: we don't need to pass values.length, it will be available anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+        } else {
+            checkIndex(index);
+            addFreeSpace(values.length);
+            T[] newValues = (T[]) new Object[values.length];
```
KseniiaMakarova, line 29: Use System.arraycopy method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
-
+        checkIndex(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T[] newValues = (T[]) new Object[values.length];
```
KseniiaMakarova, line 64: let's use the same array in System.arraycopy method
just shift values in the same array

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-
+        if (index == size) {
+            add(value);
+        } else {
+            checkIndex(index);
+            addFreeSpace();
+            T[] newValues = (T[]) new Object[values.length];
+            System.arraycopy(values, 0, newValues, 0, index);
+            newValues[index] = value;
+            System.arraycopy(values, index, newValues, index + 1, size - index);
```
KseniiaMakarova, line 32:
```suggestion
            System.arraycopy(values, index, values, index + 1, size - index);
```
let's use the same array and just shift values

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T[] newValues = (T[]) new Object[values.length];
+        System.arraycopy(values, 0, newValues, 0, index);
+        System.arraycopy(values, index + 1, newValues, index + 1 - 1, values.length - (index + 1));
```
KseniiaMakarova, line 62: same
just copy original value in a variable first

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        values = newValues;
+        size--;
+        return result;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(values[i], element)) {
```
KseniiaMakarova, line 72: please don't use Objects class - see 1st point in checklist

KseniiaMakarova (changes requested):

PR #842, opened 2023-02-19
added solution hw-10-jv-array-list

HelenMatus:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = NON_EXISTENT_INDEX;
+        if (element == null) {
+            lengthCounter--;
+            return null;
+        }
```
denys-mg, line 75: An ArrayList elements can be an object reference or the value null, therefore is no need for this condition

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
+        for (int i = 0; i < values.length; i++) {
+            if (Objects.equals(values[i], element)) {
+                index = i;
+            }
+        }
+        checkElement(element, index);
+        checkIndex(index);
+        T value = (T) values[index];
+        remove(index);
+        return value;
```
denys-mg, line 85:
```suggestion
        throw new NoSuchElementException("The element " + element + " is not exist");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
+        if (index >= lengthCounter || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " out of bounds");
+        }
+    }
+
+    public void checkElement(T element, int index) {
+        if (index < 0) {
+            throw new NoSuchElementException("The element " + element + " is not exist");
+        }
+    }
```
denys-mg, line 121: You probably don't need this checkElement method as you used it only once, and it seems better to write it directly in 'remove(T element)' method as there we do the same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
-        return 0;
+        return lengthCounter;
     }
 
     @Override
     public boolean isEmpty() {
+        if (lengthCounter == 0) {
+            return true;
+        }
         return false;
```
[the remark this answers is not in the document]
denys-mg: Don’t complicate it

denys-mg (changes requested):
Good work, left you my remarks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        checkIndex(index);
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T value = (T) values[index];
+        int newLengthCounter = size - 1;
```
Andriy-dz, line 56: you don't need this variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) values[index];
```
Andriy-dz, line 43: you don't need to cast directly

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #843, opened 2023-02-20
realize methods for ArrayList

yurihhhT:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        elements[--size] = null;
+        return elementToRemove;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element
+                    || elements[i] != null && (elements[i].equals(element))) {
```
IvanZuravlev228, line 68: would be better if you moved it to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size() == elements.length) {
+            oldArrayCopy();
```
IvanZuravlev228, line 17: would be better if you moved it to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        System.arraycopy(elements, index + 1, elements, index, (size - index) - 1);
+        elements[--size] = null;
+        return elementToRemove;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (equalElement(elements[i],element)) {
```
IvanZuravlev228, line 63: add space after coma

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        elements = resizedElements;
+    }
+
+    private void checkIndexRange(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
+    }
+
+    private boolean equalElement(T elementFromArray, T element) {
```
IvanZuravlev228, line 93: rename parameter names for this method. Because you are comparing two parameters to each other, whatever they are, and not specifically "elementFromArray" and "element". In the future, you can pass other parameters there as well.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return elementFromArray == element
+                || elementFromArray != null && (elementFromArray.equals(element));
+    }
+
+    private void fullArray() {
+        if (size() == elements.length) {
```
IvanZuravlev228, line 99: I think it would be better if you use a property instead of a method

IvanZuravlev228 (changes requested):

IvanZuravlev228 (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
+    }
+
+    private boolean equalsForTtypeElements(T element1, T element2) {
+        return element1 == element2
+                || element1 != null && (element1.equals(element2));
+    }
+
+    private void ifArrayIsFull() {
```
nacenik, line 98: Method name should be verbs. `checkArraySize` - is better

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void oldArrayCopy() {
```
nacenik, line 80: that's method should called `grow`, cause you increase size of array and copy it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        elements = resizedElements;
+    }
+
+    private void checkIndexRange(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
+    }
+
+    private boolean equalsForTtypeElements(T element1, T element2) {
```
nacenik, line 93: `elementsEqual` or `areElementsEqual` is more suitable name for method. Also, using numbers for variable names are bad practice. Rename method variables too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
-
+        ifArrayIsFull();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        ifArrayIsFull();
```
nacenik, line 23: make this check after if. First - check bounds, next - all other logic. Why do you need to increase capacity, if you index is more then size.

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        checkArraySize();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
```
Andriy-dz, line 24:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index " + index + " out of bounds");
```

Andriy-dz: need to make more informative posts. The same with the following

Andriy-dz (approved):

PR #845, opened 2023-02-20
jv-array-list-solution

MelnychukMyhailo:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] array = new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
+        resizeArray();
+        array[size] = value;
+        size++;
```
olekskov, line 14: commands in lines 13-14 can be combined

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+
+    private void resizeArray() {
+        if (size == array.length) {
+            Object[] booferArray = new Object[array.length + (array.length >> 1)];
+            System.arraycopy(array, 0, booferArray, 0, size);
+            array = booferArray;
+        }
     }
+
+    private void isIndexValid(int index) {
```
olekskov, line 92: such naming suits for boolean method, for void it's better to use action in name (e.g. validateIndex())

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list != null) {
+            for (int i = 0; i < list.size(); i++) {
```
olekskov, line 33: for-each cycle will look more tidy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        size--;
+        return result;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == element || array[i] != null
+                    && array[i].equals(element)) {
```
olekskov, line 67: I'd recommend to use Objects.equals() method in such cases

olekskov (changes requested):
Nice job! Just need method name correction

src/main/java/core/basesyntax/ArrayList.java
MelnychukMyhailo: Do we need to make some extra import? Because if we use for-each loop with "import core.basesyntax.ArrayList;", it create some mistake. And "import java.util.ArrayList;" we can`t make.

kozhukhovsky (approved):

PR #846, opened 2023-02-20
HomeWork JavaCore/ArrayList

NickolayVoloshchuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of array");
+        }
+    }
+
+    private boolean compareT(T first, T second) {
```
Deschna, line 95: better to use name that starts with "is", like "isEquals"

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
```
Deschna, line 80: better to add "IfNeeded" to the name because method grows your array only if  condition is true

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double FACTOR_SIZE = 1.5;
+    private static final int DEFAULT_SIZE = 10;
+    private T[] data = (T[]) new Object[DEFAULT_SIZE];
```
Deschna: you should initialize this inside the constructor
`private T[] data = (T[]) new Object[DEFAULT_SIZE];`

Deschna (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growFromNeed() {
```
nacenik, line 80: better to rename `checkSize`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        if (size == data.length) {
+            int newLength = (int) (data.length * FACTOR_SIZE);
+            T[] newData = (T[]) new Object[newLength];
+            System.arraycopy(data, 0, newData, 0, data.length);
+            data = newData;
```
nacenik, line 85: create an additional method `grow` for this code and call it here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of array");
+        }
+    }
+
+    private boolean isCompare(T first, T second) {
```
nacenik, line 95: `isCompare` - not a suitable name here. `isEquals` is a more appropriate name for the validation method

nacenik (changes requested):

nacenik (approved):
Perfect! 🔥

PR #848, opened 2023-02-20
Created basic ArrayList logic

ArtemGrunin:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    public ArrayList() {
+        this(DEFAULT_CAPACITY);
+    }
+
     @Override
     public void add(T value) {
-
+        resizeIfNeeded();
+        elements[size] = value;
+        size++;
```
petro-ushchuk, line 26: You can inline it 
```suggestion
        elements[size++] = value;
```

petro-ushchuk (approved):
Good work! 🔥

PR #852, opened 2023-02-20
Implemented solution for HW.

RandomEastEcho:


NadiaHuryk (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private void increaseLength() {
+        int newLength = (int) (currentArray.length * MULTIPLIER);
+        T[] tempArray = (T[]) new Object[newLength];
+        System.arraycopy(currentArray, 0, tempArray, 0, size);
+        currentArray = tempArray;
+    }
+
+    private void indexLengthCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index");
```
karma-o, line 93: add index param to the message, check other exceptions too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < currentArray.length; i++) {
+            if (compare(currentArray[i], element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("There`s no such element");
```
karma-o, line 71: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
+        System.arraycopy(currentArray, index, currentArray, index + 1, size - index);
+        currentArray[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
karma-o, line 38: list null check would be nice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    public ArrayList() {
+        currentArray = (T[]) new Object[DEFAULT_LENGTH];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == currentArray.length) {
+            increaseLength();
+        }
```
karma-o, line 19: I think it is redundant here, since we have it other `add()` method

karma-o (changes requested):

src/main/java/core/basesyntax/ArrayList.java
RandomEastEcho: If i remove this size check - there is possible out of bounds exception - in case if currentArray is already have 10 elements.

sarakhmen: @RandomEastEcho I think @karma-o meant that you can reuse your add(T value, int index) method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    private void increaseLength() {
+        int newLength = (int) (currentArray.length * MULTIPLIER);
+        T[] tempArray = (T[]) new Object[newLength];
+        System.arraycopy(currentArray, 0, tempArray, 0, size);
+        currentArray = tempArray;
+    }
+
+    private void indexLengthCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " don`t exist!");
```
sarakhmen, line 96: Pay attention that this error message format is similar to other ones. So, think about how to create a constant with such a format.

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #857, opened 2023-02-20
Array list

AleksandrLts:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkCapacity();
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
```
ArtemPonomar, line 26: Preferable to move these lines to separate method or use modified method checkIndex(int index).
Also error message should be more informative, what was the incorrect index, what are the bounds.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (data[i] == element || (data[i] != null && data[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
ArtemPonomar, line 67: You can add which element was not found to make error message more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+    private void growDataLength() {
+        int newCapacity = data.length + (data.length >> 1);
+        T[] newData = (T[]) new Object[newCapacity];
+        System.arraycopy(data, 0, newData, 0, data.length);
+        data = newData;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
```
ArtemPonomar, line 95: Error message can be more informative. What was the incorrect index, what are the bounds?

ArtemPonomar (commented):
Great work

ArtemPonomar (changes requested):
forgot to request changes

ArtemPonomar (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        checkCapacity();
+        data[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkCapacity();
+        if (index > size || index < 0) {
```
Andriy-dz, line 24: here you can use the checkIndex(Index) method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        }
+    }
+
+    private String arrayListIndexOutOfBoundsMessage(int index) {
+        return String.format("Index %d out of bounds. Size of list: %d", index, size);
+    }
+
+    private String noSuchElementMessage(T element) {
+        return "Element " + element + " not found!";
     }
```
Andriy-dz, line 105: in fact, you don't need these methods, because you only use them in one place

Andriy-dz (changes requested):

src/main/java/core/basesyntax/ArrayList.java
AleksandrLts: I cant, because it will runtime exception,  Index 0 out of bounds. Size of list: 0. in the checkIndex(Index) method, i have another implementetion

Andriy-dz (approved):

PR #862, opened 2023-02-20
solution for arraylist

nikolaichuk-artem:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int EMPTY_SIZE = 0;
+    private static final int CAPACITY_DIVISOR = 2;
+    private Object[] dataElement = new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
 
+        if (size == dataElement.length) {
```
marianapolishchuk2169, line 17: You can add a size check in the method grow()
(so you don't have to add it every time)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get from index " + index);
+        }
```
marianapolishchuk2169, line 48: make a separate method for checking IndexOutOfBoundsException

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't remove from index "
+                    + index);
+        }
+        final T element = (T) dataElement[index];
+        System.arraycopy(dataElement, index + 1, dataElement, index, size-- - index - 1);
```
marianapolishchuk2169, line 68:
```suggestion
        System.arraycopy(dataElement, index + 1, dataElement, index, size-- - index - 1);
        dataElement[size] = null;
```
you should remove the last element after the shift

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        return element;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, dataElement[i])) {
```
marianapolishchuk2169, line 75: In the common mistakes it says "Don’t use class Objects."

marianapolishchuk2169 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
nikolaichuk-artem: added size check to method grow

src/main/java/core/basesyntax/ArrayList.java
nikolaichuk-artem: implemented method checkIndexBounds

src/main/java/core/basesyntax/ArrayList.java
nikolaichuk-artem: ty

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int EMPTY_SIZE = 0;
+    private static final int CAPACITY_DIVISOR = 2;
```
marianapolishchuk2169, line 9:
```suggestion
    private static final double CAPACITY_DIVISOR = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+    private Object[] dataElement = new Object[DEFAULT_CAPACITY];
```
marianapolishchuk2169, line 10:
```suggestion
    private T[] elements;
```

it is not necessary to create an Object immediately. And fix the relevant errors related to this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexBounds(index);
+        return (T) dataElement[index];
```
marianapolishchuk2169, line 41:
```suggestion
        return  dataElement[index];
```
You don't need to cast directly.
Fix the relevant errors related to this

marianapolishchuk2169 (changes requested):

marianapolishchuk2169 (approved):
LGTM!)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            if (newCapacity < 0) {
+                newCapacity = Integer.MAX_VALUE;
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
+        }
+    }
+
+    private void checkIndexBounds(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get from index " + index);
```
nacenik, line 97: It is not a suitable message, cause you use this check not only in `get` method. I'd recommend you rewrite the message on `throw new ArrayListIndexOutOfBoundsException("Index  " + index + " does not exist");` or something like this.

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+
     @Override
     public void add(T value) {
-
+        grow();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
```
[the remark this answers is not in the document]
nacenik: Not fixed 
that's a photo from your file. 
image

src/main/java/core/basesyntax/ArrayList.java
nacenik: not fixed message for the exception. `"Can't get from index " + index` isn't suitable, cause you use the method `checkIndexBounds` not only in `get` method.

nacenik (changes requested):

nacenik (approved):
Good job!

PR #868, opened 2023-02-20
jv-arr-list

s-hrynus:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        System.arraycopy(items, index + 1, items, index, size - 1 - index);
+        size--;
+        return item;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == items[i] || element != null && element.equals(items[i])) {
```
KyslyiViacheslav, line 66: you twice do one thing, element == items[i], element.equals(items[i]). We can add null to ArrayList and we must remote null from ArrayList

KyslyiViacheslav (changes requested):

src/main/java/core/basesyntax/ArrayList.java
KyslyiViacheslav, line 66: all good it's my mistake) you realized equal

KyslyiViacheslav (approved):

sarakhmen (approved):
LGTM!

PR #872, opened 2023-02-21
Maksym Yashyn Home Work

Nimakel:


PlagueEbola (approved):
everything is correct, but I would not name the variable MAX_SIZE but DEFAULT-SIZE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int MAX_SIZE = 10;
```
Andriy-dz, line 6: why the maximum size? better to call it DEFAULT_SIZE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
+    private static final double MULTIPLIER = 1.5;
+    private int size;
+    private Object[] array;
```
Andriy-dz, line 9:
```suggestion
    private T[] array;
```

Andriy-dz: and don't forget to pick up casting where it will be redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == element
+                    || (array[i] != null) && array[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("No such element");
```
Andriy-dz, line 66: you need to write more informative messages - indicate with which index/element the error occurred. Corrections here and in the following cases
```suggestion
        throw new NoSuchElementException("No such element " + element);
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resize() {
```
nacenik, line 79: The name `resize` means that the size has to be changed every time this method is called. Better to rename this method on `checkSize` or something like that

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
+        if (array.length == size) {
+            T[] newArray = (T[]) new Object[(int) (array.length * MULTIPLIER)];
+            System.arraycopy(array, 0, newArray, 0, size);
+            array = newArray;
+        }
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
```
nacenik, line 89: The exception massage should be informative. Which index?

nacenik (changes requested):

nacenik (approved):
Good job! 🔥

PR #873, opened 2023-02-21
Added hw solution

KyrylH:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    }
+
+    private void growCheck() {
+        if (size == array.length) {
+            grow();
+        }
+    }
+
+    private void grow() {
+        int newCapacity = size + (int)(DEFAULT_CAPACITY * GROW_MULTIPLY);
```
sarakhmen, line 95: This is not how the array length is expected to be increased. You should multiply an array length should be multiplied by 1.5 each time. But here, your array length is increased by 5 units only.

sarakhmen (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (isNotValidValue(index)) {
+            throw new ArrayListIndexOutOfBoundsException(ERROR_ELEMENT_NOT_EXIST);
+        }
```
sarakhmen, line 44: This check is very similar to the next ones, let's create a separate method for it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == element
+                    || array[i] != null && array[i].equals(element)) {
+                return remove(i);
+            }
+        }
+
```
sarakhmen, line 75: redundant empty line

sarakhmen (changes requested):
Let's improve your solution one more time ;)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        T[] arrayCopy = (T[]) new Object[newCapacity];
+        System.arraycopy(this.array, 0, arrayCopy, 0, size);
+        this.array = arrayCopy;
+    }
+
+    private boolean isNotValidValue(int index) {
+        return index >= size || index < 0;
+    }
+
+    private void checkedException(int index) {
```
sarakhmen, line 101: It's better to rename it to validateIndex or checkIndex instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isNotValidValue(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Element does not exist in position"
+                    + index);
+        }
```
sarakhmen, line 57: could be replaced with your new validation method

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #876, opened 2023-02-21
implemented ArrayList<T> class

Deschna:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int capacity;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        capacity = DEFAULT_CAPACITY;
+        elementData = (T[]) new Object[capacity];
+        size = 0;
```
NickolayVoloshchuk, line 15: In principle, the int primitive cannot be equal to null when declared, so in my opinion its initialization is redundant (if it is initialized anyway by zero)
```suggestion
        size = 0;
```

NickolayVoloshchuk (changes requested):

petro-ushchuk (approved):
Good work!

PR #878, opened 2023-02-21
Override the methods from List

yuriy49:
add grow method
add checkInvalidIndex with custom massage
BUILD SUCCESS

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,92 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] data;
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkInvalidIndex(index);
+        if (size == data.length) {
+            grow();
+        }
+        System.arraycopy(data, index, data, index + 1, size - index);
+        data[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkInvalidIndex(index);
+        return data[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkInvalidIndex(index);
+        data[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkInvalidIndex(index);
+        T item = data[index];
+        System.arraycopy(data, index + 1, data, index, size - 1 - index);
+        size--;
+        return item;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, data[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("No such element");
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        T[] newItems = (T[]) new Object[(int) (data.length * 1.5)];
+        System.arraycopy(data, 0, newItems, 0, size);
+        data = newItems;
+    }
+
+    private void checkInvalidIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index of the bounds "
+                    + size + " Check your input index!"
+                    + index);
+        }
     }
 }


IvanZuravlev228 (changes requested):

IvanZuravlev228 (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        }
+        System.arraycopy(data, index, data, index + 1, size - index);
+        data[index] = value;
+        size++;
+    }
 
+    public void checkIndexFoAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index of the bounds "
+                    + size + " Check your input index!"
```
nacenik, line 34: add space 
```suggestion
                    + size + " Check your input index! "
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private void grow() {
+        T[] newItems = (T[]) new Object[(int) (data.length * 1.5)];
+        System.arraycopy(data, 0, newItems, 0, size);
+        data = newItems;
+    }
+
+    private void checkInvalidIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index of the bounds "
+                    + size + " Check your input index!"
```
nacenik, line 96: add space

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        return size == 0;
+    }
+
+    private void grow() {
+        T[] newItems = (T[]) new Object[(int) (data.length * 1.5)];
+        System.arraycopy(data, 0, newItems, 0, size);
+        data = newItems;
+    }
+
+    private void checkInvalidIndex(int index) {
```
nacenik, line 93: `checkIndex` is enough for the name.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private T[] data;
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
```
nacenik, line 17: better to add separate logic here. That will check the size and grow, and after that just put a value on the `size` position and change this `size`.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        checkIndexFoAdd(index);
+        if (size == data.length) {
+            grow();
+        }
+        System.arraycopy(data, index, data, index + 1, size - index);
+        data[index] = value;
+        size++;
+    }
 
+    public void checkIndexFoAdd(int index) {
```
nacenik, line 31: It should be a private method. Also, according to the clean code: all private methods should be after all public methods. FIx typo mistake
```suggestion
    public void checkIndexToAdd(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
-
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexFoAdd(index);
+        if (size == data.length) {
+            grow();
+        }
```
nacenik, line 25: Extract this logic to the new method `checkSize` or something like that

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private T[] data;
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        checkSize(size, data);
```
KseniiaMakarova, line 18: we don't need to pass these values - they will be accessible anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        data[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexToAdd(index);
+        checkSize(size, data);
```
KseniiaMakarova, line 26: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        System.arraycopy(data, index + 1, data, index, size - 1 - index);
+        size--;
+        return item;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, data[i])) {
```
KseniiaMakarova, line 63: please see checklist

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("No such element");
```
KseniiaMakarova, line 67: which element?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
yuriy49: I don't quite understand what that means

src/main/java/core/basesyntax/ArrayList.java
yuriy49: I don't quite understand what that means

KseniiaMakarova: https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#dont-use-class-objects

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        data[size] = null;
+        return item;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, data[i])) {
+                data[size] = null;
```
KseniiaMakarova, line 65: why do we need this here? let's remove this line

KseniiaMakarova (changes requested):
see this and previous comment

Andriy-dz (approved):

PR #881, opened 2023-02-21
ArrayList interface List implements

AnonimToster:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_CAPACITY = 10;
+    private static final float GROW_VALUE = 1.5f;
+    private T[] values;
+    private int size;
+
+    public ArrayList() {
+        values = (T[])new Object[START_CAPACITY];
+        size = 0;
```
sarakhmen, line 13: size is initialized with 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_CAPACITY = 10;
+    private static final float GROW_VALUE = 1.5f;
```
sarakhmen, line 7:
```suggestion
    private static final float GROW_MULTIPLIER = 1.5f;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        values = (T[])new Object[START_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        values[size] = value;
+        size++;
+        grow();
```
sarakhmen, line 20: You should grow your storage before adding new items to it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index must be less then size!");
+        } else if (index == size) {
+            add(value);
+        } else {
+            T oldValue = values[index];
+            values[index] = value;
+            add(oldValue, index + 1);
```
sarakhmen, line 32: You should move all items one position to the right, instead of adding oldValue to the end of the array. Consider using System.arraycopy here. Also remember to check storage size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index must be less then size!");
+        }
```
sarakhmen, line 50: This check is duplicated, consider creating a separate method for it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     public T remove(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index must be less then size!");
+        }
+        size--;
+        T removedValue = values[index];
+        for (int i = index; i <= size; i++) {
+            values[i] = values[i + 1];
+        }
```
sarakhmen, line 71: Let's use System.arraycopy instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private void grow() {
+        if (size == values.length) {
+            int newSize = (int)(GROW_VALUE * values.length);
+            T[] newValues = (T[])new Object[newSize];
+            for (int i = 0; i < values.length; i++) {
+                newValues[i] = values[i];
+            }
+            values = newValues;
+        }
     }
```
sarakhmen, line 105: This method is unclear. Actually, it doesn't simply grow storage but also checks if size == values.length. So, consider renaming the method appropriately. I'd recommend naming it 'ensureCapacity' or something like this. Also use System.arraycopy instead of loop

sarakhmen (changes requested):
Let's improve your solution ;)

src/main/java/core/basesyntax/ArrayList.java
AnonimToster: > You should move all items one position to the right, instead of adding oldValue to the end of the array. Consider using System.arraycopy here. Also remember to check storage size

oldvalue will not be added at the and of array. it will be added on poistion index+1. 
recusively all elements will move to the right, and the size check will be performed when adding a value to a new cell.

Andriy-dz (approved):

PR #882, opened 2023-02-21
add decision array list 1.0

vfedychkanych:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element on position " + index);
+        }
+        capEnsures();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
```
petro-ushchuk, line 27: You already made your array bigger in `capEnsures`, no need to call `arraycopy` twice, just put element in the end

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void capEnsures() {
```
petro-ushchuk, line 82: don't make naming short, rename to `capacityEnsures`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+        if (size == elements.length) {
+            int capacity = (int) (elements.length * STANDART_GROW);
+            T[] tempArray = (T[]) new Object[capacity];
+            System.arraycopy(elements, 0, tempArray, 0, size);
+            elements = tempArray;
+        }
+    }
+
+    private void capChecking(int index) {
```
petro-ushchuk, line 91: And this one too

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
karma-o: I think we should rather remove the `capEnsures()` call from here, just so that we do not have douple array copy, in some cases

petro-ushchuk (changes requested):
Very good! But fix this small issue, try to keep your code regarding to DRY principle

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        capacityEnsures();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element on position " + index);
+        }
```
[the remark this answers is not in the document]
vfedychkanych: There the check is a little different, so I created another method, I hope it will be better

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("This index out of bounds " + index);
+        }
+    }
+
+    private void addingChecking(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element on position " + index);
+        }
     }
```
petro-ushchuk, line 101: These methods are the same, read comment to the previous PR

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
vfedychkanych: So in one **index > size** but in another **index >= size**
it is not one and the same, no?

petro-ushchuk: Answered in your channel.

petro-ushchuk (changes requested):

PR #884, opened 2023-02-22
Hw 1 solution

BetelgeuseUA:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            grow();
+        }
+        elementData[size] = value;
+        size++;
```
olekskov, line 20: Command in lines 19-20 may be combined

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+            }
+            value = checkRemoveIndex(index);
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    index = i;
+                    break;
+                }
+            }
+            value = checkRemoveIndex(index);
```
olekskov, line 95: Too heavy construction with 2 cycles with brakes and repeatable logic, please rework using Objects.equals() method which will allow to perform null-safe check and use one cycle for both null and non-null cases

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index not valid");
+        }
+        return true;
+    }
+
+    private void grow() {
+        int newCapacity = size;
+        T[] newElementData;
+        if (size == elementData.length) {
```
olekskov, line 120: Seems we call grow() method only after size check were already performed in parent methods. Do we need additional check here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+            elementData[index] = value;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (elementData.length - size < list.size()) {
+            grow();
```
olekskov, line 42: What if we obtain input list 10 times bigger than current? Our 1 resize won't be enough to place all the data.
I'd recommend to call here add() method in cycle and let it determine if additional resizes are required

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-        return null;
+        T value = null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, elementData[i])) {
+                index = i;
+                break;
+            }
+        }
+        value = checkRemoveIndex(index);
```
olekskov, line 82: If you implement above comment, will we need checkRemoveIndex() method at all or we will definitely know that after cycle is passed we do not have proper object to delete?
And will we need value and index variables?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        T value = null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, elementData[i])) {
+                index = i;
+                break;
```
olekskov, line 79:
```suggestion
                return remove(i);
```

olekskov (changes requested):
Nice improvement, just need to polish remove() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+        } else if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index not valid");
```
okuzan, line 28: throwing exception doesn't need to be in if-else clauses, after throwing we exit immediately

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    }
+
+    private void grow() {
+        int newCapacity = size;
+        T[] newElementData;
+        newCapacity += newCapacity / 2;
+        newElementData = (T[]) new Object[newCapacity];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        elementData = (T[]) new Object[newCapacity];
+        System.arraycopy(newElementData, 0, elementData, 0, size);
```
okuzan, line 106:
```suggestion
        T[] data = elementData;
        values = (T[]) new Object[size * 3 / 2];
        System.arraycopy(data, 0, elementData, 0, size);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean checkIndex(int index) {
```
okuzan, line 92: I don't think we need to return something here. Just call it directly without if-wrapping, and make it void

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index not valid");
+        }
+    }
+
+    private void grow() {
+        T[] data = elementData;
+        elementData = (T[]) new Object[size * 3 / 2];
```
Ivan95kos, line 95: All magic numbers, strings, symbols in your code should be constants.
[this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

Ivan95kos (changes requested):
Please, fix previous comment

Ivan95kos (approved):
Great job

PR #887, opened 2023-02-28
jv-array-list-solution

knyazkyylubomir:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+            System.arraycopy(array, index, copiedArray, INDEX_BY_ZERO, size - index);
+            System.arraycopy(copiedArray, INDEX_BY_ONE, array, index, size--);
+            return copiedArray[INDEX_BY_ZERO];
+        }
+        if (index < INDEX_BY_ZERO) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "You cannot remove element by negative index");
+        }
+        throw new ArrayListIndexOutOfBoundsException(
+                "You cannot remove element by not existent index");
```
okuzan, line 100: let's simplify this
this method can be written like this:
create and call `checkIndexInBounds` method
get T value you are about to remove
call System.arraycopy once
then        array[--size] = null;
and return value

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
     @Override
     public void set(T value, int index) {
-
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "You cannot add element by not existent index");
+        } else if (index < INDEX_BY_ZERO) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "You cannot add element by negative index");
+        }
```
okuzan, line 76: here just call your checker method - `checkIndexInBounds`
The message will be generic but that's not a problem, inlucde size and index to the message and it will be informative enough for the user of your class to figure out what is wrong

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < size && index >= INDEX_BY_ZERO) {
+            return array[index];
+        }
```
okuzan, line 64: use checker method to throw an exception if needed, that just return array[index] without if-wrapper

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+                    "You cannot add element by not existent index");
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
okuzan, line 55: it's okay as is, but a good use case for for-each loop too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+            System.arraycopy(array, INDEX_BY_ZERO, copiedArray, INDEX_BY_ZERO, index);
+            copiedArray[index] = value;
+            System.arraycopy(array, index, copiedArray, index + INDEX_BY_ONE, size + INDEX_BY_ONE);
+            System.arraycopy(copiedArray, INDEX_BY_ZERO, array, INDEX_BY_ZERO, size + INDEX_BY_ONE);
+            size++;
+            return;
+        } else if (index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "You cannot add element by not existent index");
+        }
```
okuzan, line 48: only one system.arraycopy is necessary instead of all of this
use if check to throw an exception, then just goes the rest of your code, it's a common pattern. You don't need to wrap it, because it wouldn't be reached in the first place if those conditions were false (exception throwing means exiting immeadiately)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int DIVIDE_BY_TWO = 2;
+    private static final int INDEX_BY_ZERO = 0;
+    private static final int INDEX_BY_ONE = 1;
+    private static final int EMPTY = 0;
```
okuzan, line 10: we don't need these constants, it's okay to hardcode 0 and 1 as arguments to System.arraycopy, no big deal. You can create one constant such as GROW_FACTOR = 1.5

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+    private T[] array;
```
okuzan, line 11: ### 
```suggestion
    private T[] values;
```

okuzan (changes requested):

kozhukhovsky (approved):

PR #889, opened 2023-03-05
ArrayList-hw ex.1

vojtcekhovski:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private void resizeIfNeeded() {
+        if (size == capacity) {
+            capacity *= DEFAULT_MULTIPLIER;
+            T[] expandedArray = (T[]) new Object[capacity];
+            System.arraycopy(elementData, 0, expandedArray, 0, size);
+            elementData = expandedArray;
+        }
+    }
+
+    public void checkIndex(int index) {
```
uliana-andrushchuk, line 27: make this method private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of ArrayList bound!");
+        }
+    }
+
+    public boolean isEquals(Object first, Object second) {
```
uliana-andrushchuk, line 33:
```suggestion
    private boolean isEquals(Object first, Object second) {
```

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
+    }  
 
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() > capacity) {
+            resizeIfNeeded();
```
olekskov, line 59: This solution will cause infinite cycle in case incoming list really requires resize as you're checking if resize needed also inside method resizeIfNeeded() and it will not know anything about how much elements will come and will still saying that size is fine

Please use your add() method in a loop and allow it determine itself if resize needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private int capacity;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        capacity = DEFAULT_CAPACITY;
+        elementData = (T[]) new Object[capacity];
+    }
+
+    private void resizeIfNeeded() {
```
olekskov, line 18: please move private methods after public because inner logic that comes before public methods that allow to get understanding of a class looks confusing

olekskov (changes requested):

Rommelua (changes requested):
You should fix (or answer to) all mentor's comments before re-request new review. Be attentive and do not hesitate to contact us in Slack if you do not understand something.

image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
```
Ivan95kos, line 5: You should not use this annotation
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    // Constants for setting the size and increasing the size of the list by default.
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double DEFAULT_MULTIPLIER = 1.5;
+    // Class fields
+    private int capacity; // current list capacity
```
Ivan95kos, line 11: useless veriable, better to use `elementData.length`
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: not fixed
All private methods should be below public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    // Constants for setting the size and increasing the size of the list by default.
```
Ivan95kos, line 7: You don't need to comment the code, 
the code should be easy to read without comments

Ivan95kos (changes requested):
do not write changes in the master branch, create a separate branch for this

PR #908, opened 2023-03-24
1.Created private static fields and two necessary parametres 2.Declar…

oooRule:
…e the size of standart Array 3.Made grow method and checkIndex 4.Overrided methods one by one. 5.Importing NoSuchElementException.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_SIZE = 10;
+    public static final double GROW_RATE = 1.5;
+    private int size;
+    private T[] elementData;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfArrayFull();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkI(index);
+        growIfArrayFull();
+        System.arraycopy(elementData,index,elementData,index + 1,size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkI(index);
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkI(index);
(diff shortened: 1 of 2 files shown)


uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkI(index);
+        growIfArrayFull();
+        System.arraycopy(elementData,index,elementData,index + 1,size - index);
```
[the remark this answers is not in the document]
okuzan: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
okuzan, line 39: it's ok, but I would rather do for-each

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        size--;
+        return deletedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == null && element == null
+                    || elementData[i] != null && elementData[i].equals(element)) {
```
okuzan, line 67: this can be simplified
`elementData[i] == element || (element != null && element.equals(elementData[i]))`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growIfArrayFull() {
+        if (size == elementData.length) {
+            int newObjectsSize = (int) (size * GROW_RATE);
+            T[] newSize = (T[]) new Object[newObjectsSize];
+            System.arraycopy(this.elementData,0, newSize, 0,size);
```
okuzan, line 88: spaces here too
use default auto formatter, in IDEA on windows: Ctrl Alt L before committing

okuzan (changes requested):
Minor issues left

src/main/java/core/basesyntax/ArrayList.java
oooRule: > it's ok, but I would rather do for-each

Screenshot 2023-03-26 at 2 06 55 PM

That mistake has been written😱 How to solve it ?

olekskov: please disregard, we used to have for-each loop working on java.util.List class but as here it's our own implementation it's not supported

olekskov (approved):
Good job

PR #913, opened 2023-04-01
implemented my own ArrayList with all needed properties

pgolibroda:


Ivan95kos (changes requested):
Please, fix all checks
Знімок екрана 2023-04-01 о 18 47 59

olekskov (approved):
Good job

PR #914, opened 2023-04-01
Created a custom ArrayList class that implements the List interface. …

TemkaHillfinger:
…Redefined methods.

Made the ArrayList internal array to grow by 1.5 times when it's full, etc. Throws a custom ArrayListIndexOutOfBoundsException if the index passed to any of the methods is invalid. Throw NoSuchElementException in remove(T element) if no such element exists.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int SIZE_ARRAY = 10;
+    private static final double GROW_FACTOR = 1.5;
+    private T[] array;
+    private int size;
+
+    public ArrayList() {
+        array = (T[])new Object[SIZE_ARRAY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            growArray();
+        }
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexOutOfBoundsAdd(index);
+        if (size == array.length) {
+            growArray();
+        }
+        System.arraycopy(array, index, array, index + 1,size - index);
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size();i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexOutOfBounds(index);
+        return array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndexOutOfBounds(index);
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexOutOfBounds(index);
+        T removed = array[index];
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growArray() {
+        T[] newArray = (T[]) new Object[((int) (array.length * GROW_FACTOR))];
+        System.arraycopy(array,0,newArray,0,array.length);
```
uliana-andrushchuk, line 86:
```suggestion
        System.arraycopy(array, 0, newArray, 0, array.length);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    public ArrayList() {
+        array = (T[])new Object[SIZE_ARRAY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            growArray();
+        }
```
uliana-andrushchuk, line 19: you can move this check `if (size == array.length)` inside `growArray()` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexOutOfBoundsAdd(index);
```
uliana-andrushchuk, line 25: Here, you can add a check using `if (index == size)` and call the `add(value)` method in such cases. 
Then, you can call `checkIndexOutOfBounds` method. This way, you can remove `checkIndexOutOfBoundsAdd` method.

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        growArray();
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            size--;
```
andriy-shulepa, line 25:
```suggestion
    return;
```
you added an element to the new index. why decrease size? after you added an element to the end of the array - just finish method execution

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
+        size--;
+        return removed;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && array[i] == null)
```
olekskov, line 67: this statement may be simplified:
```suggestion
            if ((element == array[i])
```

src/main/java/core/basesyntax/ArrayList.java
olekskov: as Uliana said here you can use your checkIndexOutOfBounds method as case when index==size already checked above
and you can remove checkIndexOutOfBoundsAdd method as it's no longer required

olekskov (changes requested):

olekskov (approved):
Good job

PR #919, opened 2023-04-06
all tests passed

sshestaka:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
```
```suggestion
    private T[] elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < elementData.length; i++) {
+            if ((elementData[i] == element)
+                    || (elementData[i] != null && elementData[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Can't remove non existent value!");
```
okuzan, line 122: add element to the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    private T[] grow(int oldSize) {
```
okuzan, line 14: private methods go after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        return newElementData;
+    }
+
+    public void checkNonExistentOrNegativePosition(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element on this position");
+        }
+    }
+
+    public void checkIndexForAddElement(int index) {
```
okuzan, line 27: this should be private
if it's used only once, you can, but don't need to extract the logic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    }
+
+    private T[] grow(int oldSize) {
+        int newSize = oldSize + oldSize / 2;
+        T[] newElementData = (T[]) new Object[newSize];
+        System.arraycopy(elementData, 0, newElementData, 0, elementData.length);
+        return newElementData;
+    }
+
+    public void checkNonExistentOrNegativePosition(int index) {
```
okuzan, line 21:
```suggestion
    public void checkIndexIsValid(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+            elementData = newElementData;
+            elementData[index] = value;
+            size++;
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
```
okuzan, line 55: check if resize needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAddElement(index);
+        if (index == elementData.length) {
+            elementData = grow(size);
+        }
```
okuzan, line 63: you don't need special check for add, you can use the usual one, you just need to place it after the if, it will filter out the case that makes them different (index == size)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+            elementData = newElementData;
+            size--;
+        } else {
+            System.arraycopy(elementData, 0, newElementData, 0, index);
+            System.arraycopy(elementData, index + 1, newElementData, index,
+                    size - 1 - index);
+            elementData = newElementData;
+            size--;
+        }
+        return removedElement;
```
okuzan, line 111: in each method of this class, you only need to use System.arraycopy once. Rewrite them accordingly. FYI you can pass the same array twice to this method (as src array and dest array simultaneously), maybe that's why you feel the need to reuse it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+        checkIndexForAddElement(index);
+        if (index == elementData.length) {
+            elementData = grow(size);
+        }
+        if (index >= 0 && index < size) {
+            addElementInsideArrayByIndex(value, index);
+            return;
+        }
+        elementData[index] = value;
+        size++;
```
okuzan, line 69: do you know that this if condition is always true and last two lines never execute? Use one System.arraycopy instead of if-block

okuzan: here's the line: 
`System.arraycopy(elements, index, elements, index + 1, size - index);`

src/main/java/core/basesyntax/ArrayList.java
okuzan, line 14: should be void

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        if ((size + list.size()) > elementData.length) {
+            elementData = grow(size + list.size());
+        }
+        for (int i = 0; i < list.size(); i++) {
+            elementData[size] = list.get(i);
+            size++;
+        }
```
okuzan, line 80: this should be just for/for-each with `add()`

src/main/java/core/basesyntax/ArrayList.java
okuzan, line 111: 1. check index
2. T value = elementData[index];
3. System.arraycopy
4. elementData[--size] = null, or just size--;
5. return value

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+    }
+
+    public void checkIndexForAddElement(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add element on this position");
+        }
+    }
+
+    public void addElementInsideArrayByIndex(T value, int index) {
```
okuzan, line 33: you don't need this method

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: Unfortunately it doesn't work... This logic works only with special check (index > size || index < 0)...

sshestaka: checkIndexForAddElement

src/main/java/core/basesyntax/ArrayList.java
sshestaka: Yes, you are right! Thank you!

src/main/java/core/basesyntax/ArrayList.java
sshestaka: It isn't true when index = 0 and size = 0, and it isn't true when index = size... If I'll put this line instead of "if" - I have to change other methods... so I don't sure if it matters...

src/main/java/core/basesyntax/ArrayList.java
sshestaka: Yes!

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: Yes, thank you!

src/main/java/core/basesyntax/ArrayList.java
sshestaka: OK

src/main/java/core/basesyntax/ArrayList.java
sshestaka: I understood what you meant. I have changed the logic of this method by adding only one System.arraycopy() and I have changed "if" logic...

kozhukhovsky (approved):

PR #921, opened 2023-04-10
Solution

MTM1963:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = (int) Math.max(elements.length * GROWTH_FACTOR, minCapacity);
+            elements = Arrays.copyOf(elements, newCapacity);
```
okuzan, line 96: to deal with arrays, we can only use systemarraycopy
image

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[--size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && elements[i] == null)
```
olekskov, line 61: this may be simplified:
```suggestion
            if ((element == elements[i])
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int size) {
```
olekskov, line 82: You always pass same argument to this method (size+1).
Also it is not the best decision to introduce method variable with same name as class variable. It makes reading code more complicated.
I'd recommend to make this method without argument and use class variable size instead

olekskov (approved):
Good job, approved. Please take into account below comments for further tasks

PR #928, opened 2023-04-23
completed task

sZlishchev:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        }
+    }
+
+    private void checkRange(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Illegal index: " + index);
+        }
+    }
+
+    private T fastRemote(int index) {
```
rovein, line 39: Do you mean `remove`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+    public Object[] toArray() {
+        return Arrays.copyOf(elementData, size);
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            elementData = grow();
+        }
```
rovein, line 60: Try to extract it into some helper method, e.g. `growIfNeeded`, and also check such logic occurrence in the next methods.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+    private static final int MIN_SIZE = 0;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = MIN_SIZE;
+    }
+
+    private Object[] grow() {
```
rovein, line 18: Do we actually need to return the array from it?

Let's try to return `void` and access `elementData` directly from here.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkRangeForAdd(index);
+        final int newSize = size;
```
rovein, line 68: Do you actually need it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        if (size - 1 > index) {
+            System.arraycopy(elementData, index + 1, elementData, index, size - 1);
+        }
+        size--;
+        return value;
+    }
+
+    private boolean arrayListEquals(Object first, Object second) {
+        return (first == second) || (first != null && first.equals(second));
+    }
```
rovein, line 50: Let's move this method closer to its first usage. Also, could be renamed to `elementsEqual`.

rovein (changes requested):
Good job! But let's improve your solution a little bit :)

rovein (approved):
Good job! Keep the rhythm!

PR #929, opened 2023-04-24
solved task

OlehPryshchepa:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        }
     }
 
     @Override
     public void add(T value, int index) {
-
+        isValidIndexForAdd(index);
+        if (size + 1 > elements.length) {
+            resize();
+        }
```
rovein, line 31: Could be extracted into a separate method, e.g. `resizeIfNeeded`.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        if (list.size() + size > elements.length) {
+            resize();
+        }
+        int indexOfLastElement = 0;
+        for (int i = 0; i < elements.length; i++) {
+            if (elements[i] == null) {
+                indexOfLastElement = i;
+                break;
+            }
+        }
```
rovein, line 48: `size - 1` is your `indexOfLastElement`, isn't it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        for (int i = 0; i < size; i++) {
+            if (elements[i] != null && elements[i].equals(element)) {
+                deletedElement = elements[i];
+                remove(i);
+                return deletedElement;
+            } else if (elements[i] == element) {
+                deletedElement = elements[i];
+                remove(i);
+                return deletedElement;
+            }
```
rovein, line 90: Do wee need two conditions with the same body?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        }
+        return true;
+    }
+
+    private boolean isValidIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("wrong index: " + index
+                    + " for ArrayList size: " + size);
+        }
+        return true;
```
rovein, line 121: The return value of these methods is never used. Let's better rename them and change the signature.
```suggestion
    private void verifyIndexForAdd(int index) {
        if (index > size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("invalid index: " + index + " not exist");
        }
    }

    private void verifyIndex(int index) {
        if (index >= size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("wrong index: " + index
                    + " for ArrayList size: " + size);
        }
```

rovein (changes requested):
Good job! But let's improve your solution ;)

PR #930, opened 2023-04-24
Implemented the necessary methods

GradientAspen:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+
+        T[] newValues = (T[]) new Object[newCapacity];
+        System.arraycopy(values, 0, newValues, 0, size);
+        return newValues;
+    }
+
     @Override
     public void add(T value) {
+        if (size == values.length) {
+            values = growSize();
```
olekskov, line 29: You have problem with naming here. Method growSize() is expected to be void or return int according to it's name but it returns array.
Please change naming to be more obvious (like getGrownArray() etc.) or rework it to be void so you will use like
if (size == values.length) {
             growArraySize();

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        } else {
+            return values[index];
+        }
+
     }
 
     @Override
-    public void set(T value, int index) {
+    public void set(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size || index < 0) {
```
olekskov, line 72: you have similar checks nearly in every method, please extract them to separate method(s) and reuse

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+
+    private T[] getGrownArray() {
+        int newCapacity = 0;
+        int oldCapacity = values.length;
+        newCapacity = (int) (oldCapacity * DEFAULT_SIZE_GROW);
+
+        T[] newValues = (T[]) new Object[newCapacity];
+        System.arraycopy(values, 0, newValues, 0, size);
+        return newValues;
+    }
```
olekskov, line 30: 1. Please place all private methods after all public ones
2. I'd not create separate method for 1 if statement. May be combined:
```suggestion
    private void checkSize() {
        if (size < values.length) {
            return;
        }
        int oldCapacity = values.length;
        int newCapacity = (int) (oldCapacity * DEFAULT_SIZE_GROW);

        T[] newValues = (T[]) new Object[newCapacity];
        System.arraycopy(values, 0, newValues, 0, size);
        values = newValues;
    }
```

src/main/java/core/basesyntax/ArrayList.java
olekskov: not fixed, you have similar code on lines 63-65, 73-75, 82-84
please introduce some void method like checkIndex or something like that that will contain this logic

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        checkSize();
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index");
+        }
```
uliana-andrushchuk, line 27: You can use `checkIndex()`method here if you add check` if (index == size) `

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+
+        final T removeValue = (T) values[index];
+        final T[] resultCopy = (T[]) new Object[values.length - 1];
+        System.arraycopy(values, 0, resultCopy, 0, index);
+        System.arraycopy(values, index + 1, resultCopy, index, size - (index + 1));
+        values = resultCopy;
```
olekskov, line 66: Too many heavy operations for this logic.
You should use 1 arraycopy() method to move tail of array 1 position to the left (you can copy and insert in same array)
e.g.
was [0, 1, 2, 3, 4], we deleting 2
we copying 3 and 4 for one position to the left, obtaining [0, 1, 3, 4, 4]
and then we reducing size variable which cuts unimportant tail: [0, 1, 3, 4]

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public void add(T value, int index) {
+        if (index != size) {
+            checkIndex(index);
+        }
+        checkSize();
+        T[] resultCopy = (T[]) new Object[values.length];
+        System.arraycopy(values, 0, resultCopy, 0, index + 1);
+        resultCopy[index] = value;
+        System.arraycopy(values, index, resultCopy, index + 1, size - index);
```
olekskov, line 32: same situation here, we should copy tail of array 1 position to the right and change value of index without creating new array or double usage of arraycopy() method

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (values[i] == element || values[i] != null && values[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("informative message");
```
olekskov, line 73: good joke but never do like this on real project :)

olekskov (approved):
Good job!

PR #931, opened 2023-04-25
Created realization ArrayList which implements the List interface.

Stipaxa-hub:
Created realization ArrayList which implements the List interface.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -4,6 +4,7 @@ import java.util.NoSuchElementException;
 
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_FACTOR = 1.5;
     private T[] elementData;
     private int size;
 
@@ -67,8 +68,7 @@ public class ArrayList<T> implements List<T> {
                 return removedElement;
             }
         }
-        checkNoSuchElementException(removedElement);
-        return removedElement;
+        throw new NoSuchElementException("Value doesn't exist! " + element);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        T removedElement = null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == element
+                    || elementData[i] != null && elementData[i].equals(element)) {
+                removedElement = remove(i);
+                return removedElement;
+            }
+        }
+        checkNoSuchElementException(removedElement);
+        return removedElement;
```
okuzan, line 71:
```suggestion
        throw new NoSuchElementException("Value does not exist! " + element);
```
we will reach this point only if we didn't find equal element

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        return size == 0;
+    }
+
+    private void growIfArrayFull() {
+        if (size == elementData.length) {
+            int oldLength = size;
+            int newLength = oldLength + (oldLength >> 1);
+            T[] newData = (T[]) new Object[newLength];
+            System.arraycopy(elementData, 0, newData, 0, size);
+            elementData = newData;
```
okuzan, line 90:
```suggestion
        T[] data = elementData;
        elementData = (T[]) new Object[(int) (size * GROW_FACTOR)];
        System.arraycopy(data, 0, elementData, 0, size);

```

okuzan (changes requested):
Good job so far!

uliana-andrushchuk (approved):

PR #933, opened 2023-04-26
task solved

MykytaKat:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private int size = 0;
+    private T[] objects;
+
+    public ArrayList() {
+        objects = (T[]) new Object[10];
```
sarakhmen, line 11: magic number

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        increasingTheSize();
```
sarakhmen, line 16: as described in common mistakes, let's rename your 'increasingTheSize()' method to make it's purpose more clear

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private int size = 0;
+    private T[] objects;
```
sarakhmen, line 8:
```suggestion
    private T[] objects;
```
I'd recommend renaming to 'elements' or 'values'

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (size <= index || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index.");
+        }
```
sarakhmen, line 28: this code could be moved to a separate method to avoid code duplication

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        objects[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            increasingTheSize();
+            objects[size++] = list.get(i);
```
sarakhmen, line 39: let's reuse your add() method here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (size <= index || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index.");
+        }
```
sarakhmen, line 47: as described before, let's move to the separate method and reuse it everywhere. Also exception message isn't clear enough, let's add the index number

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (size <= index || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index.");
+        }
+        final T removedObject = objects[index];
+        objects[index] = null;
```
sarakhmen, line 65: this statement is redundant and could be removed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        System.arraycopy(objects, index + 1, objects, index, size - index - 1);
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(objects[i], element)) {
```
sarakhmen, line 74: as described in common mistakes, let's not use Objects class

sarakhmen (changes requested):
Good job but let's improve your solution :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int size = 0;
+    private final int defaultSize = 10;
```
maxline, line 7: make this field a constant (add static modifier). remember the naming for constants and ordering (constants should introduced before regular class fields.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int size = 0;
```
maxline, line 6: assigining to 0 is redundant for the class fields, this is default value for `int` type variables

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+
+    private void resize() {
+        if (elements.length == size) {
+            T[] increasedObjectsArray = (T[]) new Object[(int) (elements.length * 1.5)];
+            System.arraycopy(elements, 0, increasedObjectsArray, 0, elements.length);
+            elements = increasedObjectsArray;
+        }
+    }
+
+    private void indexExeption(int index) {
```
maxline, line 89: remember a method naming, name should be a verb. I'd recommend for this case rename as `checkIndex()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        if (size <= index || index < 0) {
```
maxline, line 90:
```suggestion
        if (index>= size || index < 0) 
```
looks more readable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resize() {
+        if (elements.length == size) {
```
maxline, line 82:
```suggestion
        if (size == elements.length) {
```
looks more readable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+            T[] increasedObjectsArray = (T[]) new Object[(int) (elements.length * 1.5)];
```
maxline, line 83:
```suggestion
            T[] resizedElements = (T[]) new Object[(int) (elements.length * 1.5)];
```
Try to not use word `object` in variable names

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
-
+        indexExeption(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexExeption(index);
+        final T removedObject = elements[index];
```
maxline, line 55:
```suggestion
        final T removedElement = elements[index];
```
Try to not use `object` word in variables naming

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] != null && elements[i].equals(element) || element == elements[i]) {
```
maxline, line 64: please use this approach for check equality
`(element == list[i] || (list[i] != null) && list[i].equals(element)`

maxline (changes requested):
good job! let's improve the solution

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private int size;
+    private static final int defaultSize = 10;
```
maxline, line 7: what about naming convention for the constants? and ordering? I wrote about it in previous comment.

maxline (changes requested):
Well done. Left one comment. And please check while the build is failed.

rovein (approved):
Good job!

PR #934, opened 2023-04-26
Implements my own List in ArrayList

Gamemod13:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+    private static final int SOFT_MAX_ARRAY_LENGTH;
+    private static final Object[] DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA;
+    private static final int NOT_EXISTING_INDEX;
 
+    static {
+        NOT_EXISTING_INDEX = -1;
+        DEFAULT_CAPACITY = 10;
+        SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
+        DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA = new Object[]{};
     }
```
sarakhmen, line 17: let's initialize your constants directly in fields

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+    private static final int DEFAULT_CAPACITY;
+    private static final int SOFT_MAX_ARRAY_LENGTH;
+    private static final Object[] DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA;
```
sarakhmen, line 9: This constant is definitely redundant and could be removed and replaced

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
-    public void add(T value, int index) {
+    private Object[] elementData;
+    private int size;
 
+    public ArrayList() {
+        elementData = DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA;
     }
 
     @Override
     public void addAll(List<T> list) {
```
sarakhmen, line 27: your solution is overcomplicated actually. Let's try to simplify everything. For example, here you can just iterate throughout list and reuse your add() method to add new element to your array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+            objects[i] = list.get(i);
+        }
+        return objects;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, size);
+        final Object[] elements = elementData;
```
sarakhmen, line 75: do you really need this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        return NOT_EXISTING_INDEX;
+    }
+
+    private void fastRemove(Object[] es, int i) {
+        final int newSize;
+        if ((newSize = size - 1) > i) {
+            System.arraycopy(es, i + 1, es, i, newSize - i);
+        }
+        es[size = newSize] = null;
     }
```
sarakhmen, line 118: This method is also redundant. Instead of this whole method you can just call something like this:
`System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);`
also remember to decrement the 'size' after element removal

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+    static {
+        NOT_EXISTING_INDEX = -1;
+        DEFAULT_CAPACITY = 10;
+        SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
+        DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA = new Object[]{};
     }
 
-    @Override
-    public void add(T value, int index) {
+    private Object[] elementData;
```
sarakhmen, line 19: let's make the name of this array more clear. Also it should be parametrized:
```suggestion
    private T[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int searchingIndex = foundElementIndex(element);
+        if (searchingIndex > NOT_EXISTING_INDEX) {
+            fastRemove(elementData, searchingIndex);
+            return element;
```
sarakhmen, line 87: here you can reuse your remove by index method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        }
+        throw new NoSuchElementException("Not found such element.");
+    }
+
+    private int foundElementIndex(T element) {
```
sarakhmen, line 92:
```suggestion
    private int findIndex(T element) {
```
Also let's move your private methods after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        final Object[] elements = elementData;
+        final int size = this.size;
```
sarakhmen, line 94: redundant variables

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+                    return i;
+                }
+            }
+        } else {
+            for (; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
```
sarakhmen, line 108: could be replaced by one for-loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        add(value, elementData, size);
     }
 
     @Override
-    public boolean isEmpty() {
-        return false;
+    public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        final int size;
+        Object[] elementData;
```
sarakhmen, line 129: redundant variables

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        if ((size = this.size) == (elementData = this.elementData).length) {
+            elementData = grow();
+        }
+        System.arraycopy(elementData, index,
+                elementData, index + 1,
+                size - index);
+        elementData[index] = value;
+        this.size = size + 1;
```
sarakhmen, line 137:
```suggestion
        size++;
```
'size' is actually 'this.size' after redundant variables removal

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
 
     @Override
-    public boolean isEmpty() {
-        return false;
+    public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        final int size;
+        Object[] elementData;
+        if ((size = this.size) == (elementData = this.elementData).length) {
+            elementData = grow();
```
sarakhmen, line 131:
```suggestion
            elementData = grow();
```
your grow method actually can directly change your elementsData array and not return anything

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        elementData[index] = value;
+        this.size = size + 1;
+    }
+
+    private void add(T element, Object[] elementData, int size) {
+        if (size == elementData.length) {
+            elementData = grow();
+        }
+        elementData[size] = element;
+        this.size = size + 1;
```
sarakhmen, line 145:
```suggestion
        elementData[size++] = element;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0 || elementData != DEFAULT_CAPACITY_EMPTY_ELEMENT_DATA) {
+            int newCapacity = newLength(oldCapacity,
+                    minCapacity - oldCapacity,
+                    oldCapacity >> 1);
+            return elementData = Arrays.copyOf(elementData, newCapacity);
+        } else {
+            return elementData = new Object[Math.max(DEFAULT_CAPACITY, minCapacity)];
+        }
+    }
```
sarakhmen, line 178: This logic is overcomplicated actually. Try to simplify it. Your grow() method could be implemented in 3 lines actually. Also use System.arraycopy() please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,196 @@
...
+
+    private Object[] grow() {
+        return grow(size + 1);
+    }
+
+    @SuppressWarnings("unchecked")
+    private T elementData(int index) {
+        return (T) elementData[index];
+    }
```
sarakhmen, line 187: let's remove this redundant method

sarakhmen (changes requested):
Let's improve your solution

src/main/java/core/basesyntax/ArrayList.java
Gamemod13: Can I leave these constants? I need them to check whether the new array does not exceed the maximum possible and also in order not to create a new object of an empty array every time. I tried make it without constants, but I also need it and hardkoded it directly in methods, so I think its no sense. Can i leave it?

Gamemod13: > This constant is definitely redundant and could be removed and replaced

Can I leave these constants? I need them to check whether the new array does not exceed the maximum possible and also in order not to create a new object of an empty array every time. I tried make it without constants, but I also need it and hardkoded it directly in methods, so I think its no sense. Can i leave it?

sarakhmen (approved):
Good job ;)

PR #936, opened 2023-04-26
My branch

V-Solianyk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int INITIAL_SIZE = 0;
+    private static final int INITIAL_INDEX = 0;
+    private static final String TEXT_OF_EXCEPTION = "Invalid index outside"
+            + " the dimension of the array";
+    private int size;
+    private T[] elementData;
```
maxline, line 13: let's rename in plural manner to be more clear through the code that this is not one element.
```suggestion
    private T[] elementsData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int INITIAL_SIZE = 0;
+    private static final int INITIAL_INDEX = 0;
```
maxline, line 9: this contant is redundant, we can just use `0` when check that index .

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        if (index >= size || index < INITIAL_INDEX) {
+            throw new ArrayListIndexOutOfBoundsException(TEXT_OF_EXCEPTION);
+        }
+        if (index == size - 1) {
+            size--;
+            return elementData[index];
+        } else {
+            T removeValue = elementData[index];
+            leftShiftOfElements(index);
+            return removeValue;
```
maxline, line 82: you can simplify this if else 
```
T removedElement= elements[index];
System.arraycopy(elements, index + 1, elements, index, (size - index) - 1);
size--;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        elementData = bufferElementData;
+        size++;
+    }
+
+    private void addValueToElementData(T value) {
+        elementData[size] = value;
+        size++;
+    }
+
+    private void leftShiftOfElements(int i) {
```
maxline, line 130: method is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            return removeValue;
+        }
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null) {
+                if (null == elementData[i]) {
```
maxline, line 90: use the next approach to check elements equality, and just call existing `remove(i)` method
```
 if ((elementData[i] == element  || elementData[i] != null && elementData[i].equals(element))) {
                return remove(i);
            }
 ```

File: src/main/java/core/basesyntax/ArrayList.java [hunk 6]
```diff
@@ -1,48 +1,137 @@
...
     public void add(T value, int index) {
-
+        if (index > size || index < INITIAL_INDEX) {
+            throw new ArrayListIndexOutOfBoundsException(TEXT_OF_EXCEPTION);
+        } else if (index == elementData.length - 1) {
+            increaseElementDataLength();
+            addValueToElementData(value);
+        } else {
+            rightShiftOfElements(value, index);
+        }
```
maxline, line 37: `else` clause is redundat because of exiting from the method when throwing exception in 31 line.

```suggestion
        if (index == elementData.length - 1) {
            increaseElementDataLength();
        } 
        System.arraycopy(data, index, data, index + 1, size - index);
        data[index] = value;
        size++;
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        elementData = Arrays.copyOf(elementData, (int) (elementData.length * 1.5 + 1));
+    }
+
+    private void rightShiftOfElements(T value, int index) {
+        T[] bufferElementData = (T[]) new Object[elementData.length + 1];
+        System.arraycopy(elementData, INITIAL_INDEX, bufferElementData, INITIAL_INDEX, index);
+        System.arraycopy(elementData, index, bufferElementData, index + 1, size);
+        bufferElementData[index] = value;
+        elementData = bufferElementData;
+        size++;
```
maxline, line 122: method is a bit overcomplicated
see the example above
```
  System.arraycopy(data, index, data, index + 1, size - index);
  data[index] = value;
  size++;
```
should be enough.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+    }
+
+    private void addValueToElementData(T value) {
```
maxline, line 125: probably better to no create this method because it is not only add value but increase the size. and when we just read the code in place where method is called it can be not obvious.

maxline (changes requested):
good job! let's improve the solution

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        }
+        elementsData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            trowException();
```
maxline, line 27: rename method as `checkIndex()` and `if (index > size || index < 0) {` inside the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        }
+        if (size == elementsData.length - 1) {
```
maxline, line 29: why in line 17  we check ` if (size == elementsData.length) `, but here is ` if (size == elementsData.length - 1)`?
also for the case when  `(index == size) ` you can simple call existing method `void add(T value)`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
 
     @Override
     public void addAll(List<T> list) {
-
+        T[] data = (T[]) new Object[list.size()];
+        for (int i = 0; i < list.size(); i++) {
+            data[i] = list.get(i);
+        }
+        if (elementsData.length - size < list.size()) {
+            resize();
```
maxline, line 44: `resize()` increases the size by multiplying to 1.5, what will happen if we try to addAll from the  `list` that has size  larger e.g. by 2 times than `elementsData` size?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        elementsData = Arrays.copyOf(elementsData, (int) (elementsData.length * 1.5 + 1));
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            trowException();
+        }
+    }
+
+    private void trowException() {
```
maxline, line 102: this method is redundant, combine it with `checkIndex()` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        throw new ArrayListIndexOutOfBoundsException("Invalid index outside"
```
maxline, line 103: add more details to the message, put the index value and the size of the array.

maxline (changes requested):
Good job! Let's make a bit more changes to improve the solution

src/main/java/core/basesyntax/ArrayList.java
V-Solianyk: I use this code twice, and I need to either create a constant with the necessary error text, or separate it into a separate method, how to do it better, please tell me, because I have already done both ways)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
 
     @Override
     public void addAll(List<T> list) {
-
+        T[] data = (T[]) new Object[list.size()];
+        for (int i = 0; i < list.size(); i++) {
+            data[i] = list.get(i);
+        }
+        if (elementsData.length - size < list.size()) {
+            int newSize = ((int) (elementsData.length * 1.5 + 1) + list.size());
```
maxline, line 46: this is code duplication, because we have similar code in method resize().
1 the simplest solution is to make method `addAll()` more concise and simple call the method `add(value)` in the loop
2 the second variant is add parameter to method `resize() ` which will resize not only by multiplying  by `1.5`, but adding needed quantity of elements (if we know that quantity of elements in more the `0.5 * size`)

maxline (changes requested):
Well done. Let's fix the last remark

maxline (approved):
Great work! Approved with small comment. Take a look for future please

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+            System.arraycopy(elementsData, index, elementsData, index + 1, size - index);
+            elementsData[index] = value;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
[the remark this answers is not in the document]
V-Solianyk: of course, I didn't pay attention, I'll edit it now!) thank you

PR #939, opened 2023-05-01
Task solution added

Tupadzaky:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -102,4 +87,49 @@ public int size() {
...
+
+    private void rangeCheckForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(outOfBoundsMsg(index));
```
V-Solianyk, line 99: I would add a little more information in the message, something like:
"Invalid index outside"
                + " the dimension of the array, index - " + index + System.lineSeparator()
                + "array size - " + size);

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -102,4 +87,49 @@ public int size() {
...
+        return "Index: " + index + ", Size: " + size;
+    }
+
+    private int newLength(int oldLength, int minGrowth, int prefGrowth) {
+        return oldLength + Math.max(minGrowth, prefGrowth);
+    }
+
+    private Object[] grow() {
+        return grow(size + 1);
+    }
```
V-Solianyk: I did not understand why you have 2 methods, this method should increase the size of the array

V-Solianyk (changes requested):
Great job, maybe I would tweak a few points a little:)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,96 +1,81 @@
...
 
-
-import core.basesyntax.util.ArraysSupport;
-
 import java.util.Arrays;
+import java.util.NoSuchElementException;
 
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
     private Object[] elementData;
```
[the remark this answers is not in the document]
Tupadzaky: it's the matter of where and want I want to cast. In this case I don't have to cast elementData array in constructor, but I have to cast return type of the get method.

src/main/java/core/basesyntax/ArrayList.java
Tupadzaky: grow() is called a couple times in my program, but if there were no such method, I had to use the grow(int minSize) method and pass size + 1 as an argument to it in all the calls to it. So the grow() method is just a simplification to not write unnecessary code

V-Solianyk: Thanks for the explanation, I didn't understand it right away

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
V-Solianyk: ok, I understand, maybe , it’ll be right

V-Solianyk (approved):
cool job some things I took note of)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
-    public void add(T value) {
+    public Object[] toArray() {
+        return Arrays.copyOf(elementData, size);
```
sarakhmen, line 17: as described in common mistakes let's use System.arraycopy instead :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        Object[] arrayToAdd = list.toArray();
+        int arrayToAddLength = arrayToAdd.length;
+        if (arrayToAddLength > elementData.length - size) {
+            grow(size + arrayToAddLength);
+        }
+        System.arraycopy(arrayToAdd, 0, elementData, size, arrayToAddLength);
+        size += arrayToAddLength;
```
sarakhmen, line 50: let's simplify your code here. I'd recommend iterating through the list and using your add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private String outOfBoundsMsg(int index) {
+        return "Index: " + index + " out of bounds for size: " + size;
+    }
```
sarakhmen, line 105: let's not create separate methods for exception message composing

sarakhmen (changes requested):
Good job, but let's improve your solution ;)

sarakhmen (approved):
Good job :)

PR #944, opened 2023-05-03
ArrayList v0.1

eugenetolkunov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,134 @@
...
+            grow();
+            addAll(list);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        if (size > index) {
```
ykhmaruk: you need to create second "check index" method (index >= size || index < 0). And remove this condition and exception from the methods. You can rename your first "chekIndex" method like "chekIndexForAdd".

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int NUM_ONE = 1;
```
sarakhmen, line 7: Actually I'd recommend using '1' literal directly and removing this constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkGrowIfNeed() {
```
sarakhmen, line 84:
```suggestion
    private void growIfArrayFull() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        if (size >= DEFAULT_SIZE) {
+            grow();
+        }
+    }
+
+    private T[] grow() {
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0 || elementData != null) {
```
sarakhmen, line 92: I'm not sure this 'if' is necessary here. Actually, if element.data is null, then elementData.length will cause NPE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+            int newCapacity = oldCapacity + (DEFAULT_SIZE >> NUM_ONE) + NUM_ONE;
+            return elementData = Arrays.copyOf(elementData, newCapacity);
```
sarakhmen, line 94: as described in common mistakes let's use System.arraycopy instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+    private void arrayCopyForAddByIndex(int index) {
+        System.arraycopy(elementData, index, elementData, index + NUM_ONE, size - index);
+    }
+
+    private void removeByIndex(T[] elementData, int index) {
+        System.arraycopy(elementData, index + NUM_ONE, elementData, index, size - NUM_ONE - index);
+        size--;
+    }
+
+    private boolean equalsElemets(Object a, Object b) {
```
sarakhmen, line 121: please pay attention to your method name, also let's make your 'a' and 'b' parameters of 'T' type

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        return (a == b) || (a != null && a.equals(b));
```
sarakhmen, line 122:
```suggestion
        return a == b || a != null && a.equals(b);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
     public void addAll(List<T> list) {
-
+        if (elementData.length > list.size()) {
+            for (int i = 0; i < list.size(); i++) {
+                elementData[size++] = list.get(i);
+            }
+        } else {
+            grow();
+            addAll(list);
+        }
```
sarakhmen, line 40: actually here you can simply iterate through list and call your add() method

sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int index = 0; index < elementData.length; index++) {
+            if (equalsElemets(elementData[index], element)) {
+                removeByIndex(elementData, index);
+                return element;
+            }
+        }
+        throw new NoSuchElementException();
```
rovein, line 69: If you throw an Exception, you should always add an informative message so that it would be clear from the stack trace what exactly went wrong.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+    private void grow() {
+        int oldCapacity = elementData.length;
+        T[] newElementData = (T[]) new Object[oldCapacity + (DEFAULT_SIZE >> 1) + 1];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        elementData = newElementData;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index >= size or index < 0");
```
rovein, line 97:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index >= size or index < 0. Input index " + index + " for size " + size);
```

rovein (approved):
Good job! See minor comments left.

PR #947, opened 2023-05-03
implemented task

chrisnyrka:


Gamemod13 (approved):
Good Work, havent any questions

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return size == 0;
+    }
+
+    private void toResize() {
+        if (size == elementData.length || (size - 1) == elementData.length) {
+            grow();
+        }
+    }
+
+    private boolean isEquals(T elements, T elementsData) {
```
sarakhmen, line 95: issues with parameter names

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return elements == null ? elements == elementsData : elements.equals(elementsData);
```
sarakhmen, line 96:
```suggestion
        return elements == elementsData && elements.equals(elementsData);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void toResize() {
```
sarakhmen, line 89: unclear method name

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        if (size == elementData.length || (size - 1) == elementData.length) {
```
sarakhmen, line 90:
```suggestion
        if (size == elementData.length) {
```
Why do you need that second condition?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+        }
```
sarakhmen, line 45: this validation logic is used several times, so let's create a separate method for it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
-
+        toResize();
+        if (index != size) {
+            if (index >= size || index < 0) {
+                throw new ArrayListIndexOutOfBoundsException("Incorrect index: " + index);
+            }
+            System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        }
+        elementData[index] = value;
+        size++;
```
sarakhmen, line 31: what if index == size ? I think it'll cause IndexOutOfBoundsException (in case your toResize() method has the second condition within 'if' block removed). Think why and try to fix it

sarakhmen (changes requested):
Overall good job, but let's improve your solution :)

src/main/java/core/basesyntax/ArrayList.java
chrisnyrka: I need check for null, otherwise I get a NullPointerException

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void isNeedToResize() {
```
sarakhmen, line 81:
```suggestion
    private void resizeIfNeeded() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        }
+    }
+
+    private boolean isEquals(T firstElement, T secondElement) {
+        return firstElement == null
+                ? firstElement == secondElement : firstElement.equals(secondElement);
+    }
+
+    private void grow() {
+        int newCapacity = (int) (elementData.length * 1.5);
```
sarakhmen, line 93: I'd also recommend creating a constant for your multiplier here

sarakhmen (approved):
LGTM! You can slightly improve your solution if you want ;)

src/main/java/core/basesyntax/ArrayList.java
sarakhmen: oh, yeah, sorry :)

PR #956, opened 2023-05-04
Implementation ArrayList

DV7Art:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+    }
+
+    private void rangeCheckForAdd(int index) {
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void resize() {
+        elementData = Arrays.copyOf(elementData, (int) (elementData.length * 1.5));
```
AnastasiiaVariuchuk, line 91: please, do not use the magic numbers 1.5 =>  create new private static final double GROWING_COEFFICIENT = 1.5d and use it here

AnastasiiaVariuchuk (commented):
Your solution is great, but lets improve it ;)

AnastasiiaVariuchuk (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            resize();
+        }
```
rovein, line 21: Let's collapse it in one single method.
image
```suggestion
            resizeIfFull() ... or ... growIfFull();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+        size--;
+        return removeData;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elementData[i], element)) {
```
rovein, line 68: :)
image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
rovein, line 72: If you throw an Exception, you should always add an informative message so that it would be clear from the stack trace what exactly went wrong.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    }
+
+    private void rangeCheckForAdd(int index) {
+        if (index < 0 || index > size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void resize() {
+        elementData = Arrays.copyOf(elementData, (int) (elementData.length * GROWING_COEFFICIENT));
```
rovein, line 92: Let's extract it into a variable.
```suggestion
        elementData = Arrays.copyOf(elementData, newLength);
```

rovein (changes requested):
Good job! But let's improve your solution ;)

src/main/java/core/basesyntax/ArrayList.java
DV7Art: When we can use Objects T_T

sarakhmen (approved):
Good job :)

PR #957, opened 2023-05-04
Override methods and create some new methods, pass all test.

Fafakepa:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        System.arraycopy(elementData, index + 1, elementData, index, size - 1 - index);
+        size--;
+        return removedValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size(); i++) {
+            if (isEqualElement(element, elementData[i])) {
```
vladyslavkryvets, line 70: I think u need to add a null check here

vladyslavkryvets (changes requested):
Good job, one little misstake in my opinion

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int GROWTH_RATE = 5;
+    private int size;
+    private T[] elementData;
+    private T[] bufferArray;
```
sarakhmen, line 10: this field actually shouldn't be declared here and could be replaced with local variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int GROWTH_RATE = 5;
```
sarakhmen, line 7: so each time your array is full it's resized only by 5 elements, not x1.5 times?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        elementData = bufferArray;
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index >= size()) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is not valid");
+        }
+    }
+
+    private boolean isEqualElement(T firstElement, T secondElement) {
```
sarakhmen, line 99:
```suggestion
    private boolean areElementsEqual(T firstElement, T secondElement) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        if (firstElement == secondElement
+                || (firstElement != null && firstElement.equals(secondElement))) {
+            return true;
+        }
         return false;
```
sarakhmen, line 104: this construction could be simplified

sarakhmen (changes requested):
Good job but let's improve your solution ;)

sarakhmen (approved):
Good job :)

PR #958, opened 2023-05-04
Hw array list

RostyslavOnysh:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double SIZE_MULTIPLIER = 1.5;
+    private T[] array;
+    private int size;
+
+    @SuppressWarnings("unchecked")
```
uliana-andrushchuk, line 13: Please, do not use @SuppressWarnings annotation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
uliana-andrushchuk, line 37:
```suggestion
        for (T element : list) {
            add(element);
        }
```
You can use for each loop here

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
RostyslavOnysh: Did I understand correctly that to implement this code block 

for (T element : list) {
add(element);
}

 I need to implement the Iterable interface? Or would it be better to do it some other way?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void increaseSize() {
+        if (array.length == size) {
+            int newCapacity = (int) (array.length * SIZE_MULTIPLIER);
+            array = Arrays.copyOf(array, newCapacity);
```
okuzan, line 88: we can only use System.arraycopy
it's native method and usually faster
image

okuzan (approved):

PR #960, opened 2023-05-05
implemented own ArrayList

m-pnsk:


src/main/java/core/basesyntax/List.java
```diff
@@ -17,5 +17,7 @@
 
     int size();
 
+    T[] toArray();
```
AnastasiiaVariuchuk, line 20: I'm not sure if it was necessary to make changes to this class, maybe it's better to make the implementation in ArrayList

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        return size == 0;
+    }
+
+    public boolean equals(T first, T second) {
+        return (first == second) || (first != null && first.equals(second));
+    }
+
+    private void grow() {
+        int oldSize = data.length;
+        int newSize = oldSize + (oldSize >> 1);
```
AnastasiiaVariuchuk, line 95: please, do not use the magic numbers 1 => create new private static final double GROWING_COEFFICIENT = 1.5d and use it here

AnastasiiaVariuchuk (approved):
Good job, I add one my additional suggestion, but it's minor, so I approve it ;)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
     public void addAll(List<T> list) {
-
+        int addedArraySize = list.size();
+        while (addedArraySize > data.length - size) {
+            grow();
+        }
+        for (int i = 0; i < addedArraySize; i++) {
+            data[size + i] = list.get(i);
+        }
+        size += addedArraySize;
```
sarakhmen, line 43: I'd recommend simply iterating through the list and using your add() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        data[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkRange(index);
+        T value = data[index];
+        if (size - 1 > index) {
+            System.arraycopy(data, index + 1, data, index, size - 1);
```
sarakhmen, line 63:
```suggestion
            System.arraycopy(data, index + 1, data, index, size - index - 1);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public boolean equals(T first, T second) {
+        return (first == second) || (first != null && first.equals(second));
+    }
```
sarakhmen, line 91:
```suggestion
    private boolean elementsAreEqual(T first, T second) {
        return first == second || first != null && first.equals(second);
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+
+    private void grow() {
+        int oldSize = data.length;
+        int newSize = oldSize + (oldSize >> 1);
+        if (newSize < 0) {
+            newSize = Integer.MAX_VALUE;
+        }
+        data = Arrays.copyOf(data, newSize);
```
sarakhmen, line 99: as described in common mistakes, let's use System.arraycopy instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+
+    private void growIfFull() {
+        if (size == data.length) {
+            grow();
+        }
+    }
+
+    private void checkAddRange(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Illigal index: " + index);
```
sarakhmen, line 110:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Illegal index: " + index);
```

sarakhmen (changes requested):

src/main/java/core/basesyntax/ArrayList.java
m-pnsk: I do not agree with this change. This parameter indicates the size of the new array. And it has decreased as a result by only one

PR #965, opened 2023-05-05
create arraylist

VRuban373:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void resize(int newLength) {
+        if (newLength < size) {
+            throw new IllegalArgumentException("New length is less than the current size");
+        }
+        T[] newArray = (T[]) new Object[newLength];
```
m-pnsk, line 94: This method should be private and then you don't need to check if (newLength < size). But you can leave that if in. It's not a mistake.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        return object;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementsAreEqual(element, elementData[i])) {
+                remove(i);
+                return element;
```
m-pnsk, line 74:
```suggestion
                return remove(i);
```

m-pnsk: Your remove(index) method has already returned the removed element

m-pnsk (changes requested):

m-pnsk (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            int newLength = (int) (elementData.length * GROWING_COEFFICIENT);
```
sarakhmen, line 18: I think this is a resize-related logic and should be moved to 'resize()' method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (size == elementData.length) {
+            int newLength = (int) (elementData.length * GROWING_COEFFICIENT);
```
sarakhmen, line 32: the same here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
-
+        verifyIndex(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        verifyIndex(index);
+        T object = get(index);
```
sarakhmen, line 63: if you're using your get() method here then verifyIndex() is called twice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    }
+
+    private void verifyIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private boolean elementsAreEqual(T first, T second) {
+        return (first == second) || (second != null && second.equals(first));
```
sarakhmen, line 102:
```suggestion
        return first == second || second != null && second.equals(first);
```

sarakhmen (changes requested):
Overall good job but let's slightly improve your solution :)

sarakhmen (approved):
Good job :)

PR #967, opened 2023-05-07
Arraylist implementation

shaman1437:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || size < index) {
+            throw new ArrayListIndexOutOfBoundsException("No valid index: " + index);
+        }
```
Ivan95kos, line 29: It is better to move in private method
```suggestion
        if (index < 0 || size < index) {
            throw new ArrayListIndexOutOfBoundsException("No valid index: " + index);
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        if (elements.length == size) {
+            grow();
+        }
```
Ivan95kos, line 32: do this check in the grow() method
```suggestion
        grow();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int SIZE = 10;
```
Ivan95kos, line 7:
```suggestion
    private static final int DEFAULT_SIZE = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
-
+        if (index < 0 || size < index) {
+            throw new ArrayListIndexOutOfBoundsException("No valid index: " + index);
+        }
+        if (elements.length == size) {
+            grow();
+        }
+        for (int i = size - 1; i >= index; i--) {
+            elements[i + 1] = elements[i];
+        }
```
Ivan95kos, line 35: You can use System.arraycopy() instead of a loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        size--;
+        elements[size] = null;
+        return removedValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < elements.length; i++) {
+            if (Objects.equals(element, elements[i])) {
```
Ivan95kos, line 72: Don’t use class Objects
https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#dont-use-class-objects

Ivan95kos (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: What are the reasons why you didn't fix it?

src/main/java/core/basesyntax/ArrayList.java
Ivan95kos: What are the reasons why you didn't fix it?

Ivan95kos (approved):
Good job,
with unsolved questions

PR #974, opened 2023-05-11
implemented my own ArrayList class with appropriate methods in it

vaniii1:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        grow();
+        verifyIndexForAdd(index);
+        Object[] newArray = new Object[array.length];
+        System.arraycopy(array, 0, newArray, 0, index);
+        newArray[index] = value;
+        System.arraycopy(array, index, newArray, index + 1, size - index);
```
olekskov, line 29: please use 1 arraycopy() method, you may perform copy of cells within same array without buffer one

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    private T[] array;
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_LENGTH];
+    }
+
     @Override
     public void add(T value) {
-
+        grow();
```
olekskov, line 17: Please rework naming to add information that it also performs check if growth is required.
Now it looks like you grow array on each add operation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        grow();
+        for (int i = 0; i < list.size(); i++) {
+            array[size] = list.get(i);
+            size++;
+        }
```
olekskov, line 40: If input list contains elements enough for 2+ resizes your method will fail.
I recommend to call your add(T value) method above in cycle and allow it determine how many resizes will be required

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     public T get(int index) {
-        return null;
+        verifyIndex(index);
+        return array[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        grow();
```
olekskov, line 51: we're replacing value on set operation, why check for growth is required?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        verifyIndex(index);
+        final T removedObject = array[index];
+        Object[] newArray = new Object[array.length];
+        System.arraycopy(array, 0, newArray, 0, index);
+        System.arraycopy(array, index + 1, newArray, index, size - index - 1);
+        array = (T[]) newArray;
```
olekskov, line 63: only one arraycopy() operation without buffer array is needed

olekskov (changes requested):

PR #977, opened 2023-05-13
solution for this task

aTiukhtii:


aTiukhtii:
is it necessary to cast (T[]) in line 83?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_LENGTH];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfLengthEqualsSize();
+        array[size] = value;
+        size++;
```
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if (array[i] == element || (array[i] != null && array[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("there is no such element in this array");
```
okuzan, line 67: add element to the message to make it more informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        }
+    }
+
+    private void verifyIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("wrong index input");
+        }
+    }
+
+    private void verifyIndexForAdd(int index) {
```
okuzan, line 95: you can use the main verifyIndex checker method for ` add`, if you place it below the length equals size check :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            int newLength = (int) (array.length * MULTIPLIER);
+            Object[] newArray = new Object[newLength];
+            System.arraycopy(array, 0, newArray, 0, array.length);
+            array = (T[]) newArray;
+        }
+    }
+
+    private void verifyIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("wrong index input");
```
okuzan, line 91: include index to the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growIfLengthEqualsSize() {
```
okuzan, line 80:
```suggestion
    private void growIfFull() {
```

okuzan (changes requested):
Very good job on your first attempt! Keep it up 💪

src/main/java/core/basesyntax/ArrayList.java
aTiukhtii: it didn't work because index > size and index >= size is important. (or I fixed it wrong🥲)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        T removedObject = array[index];
+        System.arraycopy(array, index + 1, array, index,size - index - 1);
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
```
olekskov, line 60: In the remove(T element) method, it's better to use 'size' instead of 'array.length' in the for loop condition, to avoid unnecessary iterations and possible mistakes due to leftover values after removing elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexVerifyForAdd(index);
+        growIfSizeFull();
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
```
olekskov, line 27: Regarding Oleh's comment: After such change you should be able to remove indexVerifyForAdd method
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        indexVerify(index);
        growIfSizeFull();
        System.arraycopy(array, index, array, index + 1, size - index);
        array[index] = value;
        size++;
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
aTiukhtii: thanks, now I understand🙂

olekskov (changes requested):
Need to fix build, other seems fine already

aaananas (approved):

aaananas (approved):

PR #979, opened 2023-05-16
Hw array list

tanneken:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] currentList;
+    private Object[] newList;
```
kozhukhovsky, line 9:
```suggestion
```

No need for such a variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+
+    public ArrayList() {
+        size = 0;
```
kozhukhovsky, line 12:
```suggestion
```

0 is default int value

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            currentList = newList;
+        }
+        currentList[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bounds");
+        }
```
kozhukhovsky, line 31:
```suggestion
        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException("Index is out of bounds");
        }
```

move to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        size = 0;
+        currentList = new Object[DEFAULT_CAPACITY];
+        newList = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == currentList.length) {
+            grow();
```
kozhukhovsky, line 20:
```suggestion
        if (size == currentList.length) {
            grow();
```

it's better to put it in a separate method, like resize()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        newList = new Object[currentList.length + (currentList.length >> 1)];
```
kozhukhovsky, line 119: in this method it is better to copy to the old array by increasing its size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        newList = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == currentList.length) {
+            grow();
+            System.arraycopy(currentList, 0, newList, 0, size);
+            currentList = newList;
```
kozhukhovsky, line 22:
```suggestion
```
there is no need for this if it will be in the grow() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        if (currentList.length < size + list.size()) {
+            newList = new Object[size + list.size()];
+            System.arraycopy(currentList, 0, newList, 0, size);
+            currentList = newList;
+        }
 
+        for (int i = 0; i < list.size(); i++) {
+            currentList[size++] = list.get(i);
+        }
     }
```
kozhukhovsky, line 59:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            }
+            if (element == null && currentList[i] == null) {
+                index = i;
+                break;
+            }
+        }
+
+        if (index == -1) {
+            throw new NoSuchElementException("There is no such element in the list");
+        }
```
kozhukhovsky, line 104: here it is better to loop through the old values, if you find it, you get the index and remove it through the remove (index) method

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element != null && element.equals(currentList[i])) {
+                return remove(i);
+            }
+            if (element == null && currentList[i] == null) {
+                return remove(i);
+            }
```
okuzan, line 81: same body, combine conditions - `element == currentList[i] || element != null && element.equals(currentList[i]))`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] currentList;
```
okuzan, line 8: I think something like `values` would be a better fit, more concise and to the point

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        }
+        indexChecking(index);
+        if (size == currentList.length) {
+            growIfArrayFull();
+        }
 
+        Object[] newList = new Object[currentList.length];
+        System.arraycopy(currentList, 0, newList, 0, index);
+        newList[index] = value;
+        System.arraycopy(currentList, index, newList, index + 1, size - index);
```
okuzan, line 34: in all methods of the class, use only one System.arraycopy - hint: you can have the same array as both src and dest arguments passed to it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexChecking(index);
+
+        Object[] newList = new Object[currentList.length];
+        System.arraycopy(currentList, 0, newList, 0, index);
+        System.arraycopy(currentList, index + 1, newList, index, size - index - 1);
```
okuzan, line 64: use once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        indexChecking(index);
+        currentList[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexChecking(index);
+
+        Object[] newList = new Object[currentList.length];
```
okuzan, line 62: you don't need this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        if (size == currentList.length) {
+            Object[] newList = new Object[currentList.length + (currentList.length >> 1)];
+            System.arraycopy(currentList, 0, newList, 0, size);
+            currentList = newList;
+        }
+    }
+
+    private void indexChecking(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bounds");
```
okuzan, line 107: add more info to the message, which index? what could be the max possible value (size)?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        for (int i = 0; i < size; i++) {
+            if (element != null && element.equals(currentList[i])) {
+                return remove(i);
+            }
+            if (element == null && currentList[i] == null) {
+                return remove(i);
+            }
+        }
+
+        throw new NoSuchElementException("There is no such element in the list");
```
okuzan, line 84: add more info - which element?

okuzan (changes requested):
Good job so far!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element != null && element.equals(values[i])
+                    || element == null && values[i] == null) {
+                return remove(i);
+            }
+        }
+
+        throw new NoSuchElementException("There is no element "
+                + element.toString() + " in the list");
```
okuzan, line 74:
```suggestion
                + element + " in the list");
```
toString will be called under the hood

okuzan (approved):
Well done! 💪

PR #981, opened 2023-05-20
Implemented List interface;

pochtalon:
Implemented List interface; 
Created 3 constants;
Created 3 variables "size", "capacity" and "array";
Created additional methods "grow", "legalBound"

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,124 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_CAPACITY = 10;
+    private static final int EMPTY = 0;
+    private static final double GROW = 1.5;
+    private int size;
+    private int capacity;
+    private T[] array;
+
+    public ArrayList() {
+        array = (T[]) new Object[START_CAPACITY];
+        size = EMPTY;
+        capacity = START_CAPACITY;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == capacity) {grow();}
+        array[size] = value;
+        size++;
+    }
+    private void grow() {
+        capacity = (int) (capacity * GROW);
+        T[] newArray = (T[]) new Object[capacity];
+        System.arraycopy(array, 0, newArray,0, size);
+        array = newArray;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
+            if (size == capacity) { grow(); }
+            T[] newArray = (T[]) new Object[capacity];
+            System.arraycopy(array, 0, newArray,0, index);
+            newArray[index] = value;
+            System.arraycopy(array, index, newArray,index + 1, size - index);
+            array = newArray;
+            size++;
+        }
+    }
 
+    private boolean legalBound(int index) {
+        if (index < 0 || index > size - 1) {
+            return true;
+        }
+        return false;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() > capacity) {grow();}
+        T[] acceptedArray =  (T[]) new Object[list.size()];
+        for (int i = 0; i < acceptedArray.length; i++) {
+            acceptedArray[i] = list.get(i);
+        }
+        System.arraycopy(acceptedArray, 0, array, size, acceptedArray.length);
+        size += acceptedArray.length;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
+            return array[index];
+        }
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
+            array[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
+            T result = array[index];
+            size--;
+            System.arraycopy(array, index + 1, array, index, size - index);
+            return result;
+        }
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T result = null;
+        int index = 0;
+        for (int i = 0; i < size; i++){
+            if (array[i] == element || array[i] != null && array[i].equals(element)) {
+                result = element;
+                index = i;
+                break;
+            }
+        }
+        if (result == null && element != null) {
+            throw new NoSuchElementException();
+        } else {
+            size--;
+            System.arraycopy(array, index + 1, array, index, size - index);
+            return result;
+        }
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == EMPTY;
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        }
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
```
olekskov, line 31: For better clarity, change the exception message from 'IndexOutOfBounds' to 'Index out of bounds for ArrayList'.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        }
+        System.arraycopy(acceptedArray, 0, array, size, acceptedArray.length);
+        size += acceptedArray.length;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
```
olekskov, line 61: For better clarity, change the exception message from 'IndexOutOfBounds' to 'Index out of bounds for ArrayList'.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        } else {
+            return array[index];
+        }
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
```
olekskov, line 70: For better clarity, change the exception message from 'IndexOutOfBounds' to 'Index out of bounds for ArrayList'.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+            System.arraycopy(array, index + 1, array, index, size - index);
+            return result;
+        }
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T result = null;
+        int index = 0;
```
olekskov, line 91: You can optimize the remove(T element) method by removing the 'result' variable and using a boolean flag, like 'elementFound', to check if the element is found. This way, you can avoid unnecessary variable assignments and null checks.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        } else {
+            array[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
```
olekskov, line 79: For better clarity, change the exception message from 'IndexOutOfBounds' to 'Index out of bounds for ArrayList'.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_CAPACITY = 10;
+    private static final int EMPTY = 0;
+    private static final double GROW = 1.5;
+    private int size;
+    private int capacity;
```
olekskov, line 10: Please remove capacity variable and use array.length where required
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    private T[] array;
+
+    public ArrayList() {
+        array = (T[]) new Object[START_CAPACITY];
+        size = EMPTY;
```
olekskov, line 15: default int value is 0, no need to explicitly set it and create additional constant
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() > capacity) {
+            grow();
+        }
```
olekskov, line 49: you have your list with capacity 10, user wants to add all values from list of 100 elements
will 1 grow be enough for this operation?

Please make more universal solution, most easy one is to reuse add(T value) in cycle and let in rescale array when required

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (legalBound(index)) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
+            return array[index];
+        }
```
olekskov, line 64: Please move exception throwing inside method legalBound(), give this method proper name and code will be much more clear
```suggestion
        checkIndex(index)
        return array[index];
```
Please recheck other methods with same situation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+                break;
+            }
+        }
+        if (result == null && element != null) {
+            throw new NoSuchElementException();
+        } else {
+            size--;
+            System.arraycopy(array, index + 1, array, index, size - index);
+            return result;
+        }
```
olekskov, line 105: Overcomplicated, you have all required data inside first cycle:
```suggestion
        for (int i = 0; i < size; i++) {
            if (array[i] == element || array[i] != null && array[i].equals(element)) {
                System.arraycopy(array, i + 1, array, i, size - i);
                size--;
                return element;
            }
        }
        throw new NoSuchElementException();
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        array[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("IndexOutOfBounds");
+        } else {
```
okuzan, line 28: you don't need else after if with trows, unwrap the code, nested code is less readable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+            if (size == array.length) {
+                grow();
+            }
+            T[] newArray = (T[]) new Object[array.length];
+            System.arraycopy(array, 0, newArray,0, index);
+            newArray[index] = value;
+            System.arraycopy(array, index, newArray,index + 1, size - index);
```
okuzan, line 35: in each method, use arraycopy only once, you can pass the same array to it twice, as src and as dest

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-
+        while (size + list.size() > array.length) {
+            grow();
+        }
+        T[] acceptedArray = (T[]) new Object[list.size()];
+        for (int i = 0; i < acceptedArray.length; i++) {
+            acceptedArray[i] = list.get(i);
+        }
+        System.arraycopy(acceptedArray, 0, array, size, acceptedArray.length);
+        size += acceptedArray.length;
```
okuzan, line 51: you have all the necessary logic in add method, here just call it in for/for-each loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (array[i] == element || array[i] != null && array[i].equals(element)) {
+                System.arraycopy(array, i + 1, array, i, size - i);
+                size--;
+                return element;
```
okuzan, line 81:
```suggestion
                return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    }
+
+    private void grow() {
+        T[] newArray = (T[]) new Object[(int) (array.length * GROW)];
+        System.arraycopy(array, 0, newArray,0, size);
+        array = newArray;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size - 1) {
```
okuzan, line 104:
```suggestion
        if (index < 0 || index >= size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds for ArrayList");
```
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Index " + index +  " out of bounds for ArrayList with size " + size);
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() > array.length) {
+            grow();
+        }
```
Ivan95kos, line 42: you call the `grow()` in the `add` method
```suggestion
```

Ivan95kos (approved):
Great job, with minor comments

PR #983, opened 2023-05-22
added jv-array homework solution

vitaliiplys:


okuzan (approved):
You nailed it on your first try! 👏

PR #985, opened 2023-05-30
Create ArrayList class which implements the List interface

kaptetos:
The ensureCapacity method checks if the current capacity is sufficient to accommodate the requested minimum capacity. If not, it increases the capacity by 1.5 times using Arrays.copyOf.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,127 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTORY = 1.5;
+    private static final int INVALID_INDEX = -1;
+
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this(DEFAULT_CAPACITY);
+    }
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity < 0) {
+            throw new IllegalStateException("Illegal capacity: " + initialCapacity);
+        }
+        this.elements = new Object[initialCapacity];
+        this.size = 0;
+    }
+
+    public void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = (int) (elements.length * GROWTH_FACTORY);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+    }
+
+    public int indexOf(T element) {
+        if (element == null) {
+            for (int i = 0; i < size; i++) {
+                if (elements[i] == null) {
+                    return i;
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
+        return INVALID_INDEX;
+    }
+
     @Override
     public void add(T value) {
+        ensureCapacity(size + 1);
+        elements[size++] = value;
 
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index:" + index);
+        }
+        ensureCapacity(size + 1);
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
 
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
+        checkIndex(index);
+        elements[index] = value;
 
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedElement = (T) elements[index];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+    }
+
+    public void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = (int) (elements.length * GROWTH_FACTORY);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
```
Ivan95kos, line 33: Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity < 0) {
+            throw new IllegalStateException("Illegal capacity: " + initialCapacity);
+        }
+        this.elements = new Object[initialCapacity];
+        this.size = 0;
+    }
+
+    public void ensureCapacity(int minCapacity) {
```
Ivan95kos, line 26: All methods used in the middle of the class must be private
```suggestion
    private void ensureCapacity(int minCapacity) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            int newCapacity = (int) (elements.length * GROWTH_FACTORY);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
+        }
+    }
+
+    private void checkIndex(int index) {
```
Ivan95kos, line 37: All private methods should be below public methods

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
+        return INVALID_INDEX;
```
okuzan, line 123: used only once. Combine conditions in if-else - the body is the same, and the loop is the same. Move the loop to the `remove` method. Remove INVALID_INDEX constant/field. Change the body to `remove(i)`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTORY = 1.5;
```
```suggestion
    private static final double GROWTH_FACTOR = 1.5;
```
factory = завод

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+    private static final int INVALID_INDEX = -1;
```
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+    public ArrayList() {
+        this(DEFAULT_CAPACITY);
+    }
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity < 0) {
+            throw new IllegalStateException("Illegal capacity: " + initialCapacity);
+        }
+        this.elements = new Object[initialCapacity];
+        this.size = 0;
```
okuzan, line 22: 0 is the default for int, this doesn't achieve anything. remove redundant `this`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = (int) (elements.length * GROWTH_FACTORY);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
```
okuzan, line 95: just remove the argument from the method and you won't need to validate it. You can get everything you need here - elements.length, GROW_FACTOR, size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedElement = (T) elements[index];
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[size - 1] = null;
+        size--;
```
```suggestion
        elements[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTORY = 1.5;
+    private static final int INVALID_INDEX = -1;
+
+    private Object[] elements;
```
okuzan, line 10:
```suggestion
    private T[] elements;
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity < 0) {
+            throw new IllegalStateException("Illegal capacity: " + initialCapacity);
+        }
+        elements = (T[]) new Object[initialCapacity];
+        size = 0;
+    }
+
+    private void ensureCapacity() {
```
Ivan95kos, line 24: All private methods should be below public methods
Move methods to end of the class

Ivan95kos (approved):
Great job, with minor comments

PR #988, opened 2023-05-31
first version commit

SieryiD:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,155 @@
...
+    public T[] elementArrayPlusSize(T[] elementArray, int increasingInSizeByInt) {
+        int futureSize = size + increasingInSizeByInt;
+        while (futureSize >= capacity) {
+            capacity += (capacity >> 1);
+        }
+        T[] arrayPlusSize = (T[]) new Object[capacity];
+        for (int a = 0; a < size; a++) {
+            arrayPlusSize[a] = elementArray[a];
+        }
+        return arrayPlusSize;
```
aaananas, line 26: image

aaananas (changes requested):
**Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected**

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementArray;
+    private int size;
+    private int capacity;
```
Ivan95kos, line 9: current capacity can be calculated based on size and array length, not need to have separate variables for them

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+
+    public ArrayList() {
+        capacity = DEFAULT_CAPACITY;
+        elementArray = (T[]) new Object[capacity];
+    }
+
+    public T[] elementArrayPlusSize(T[] elementArray, int increasingInSizeByInt) {
```
Ivan95kos, line 16: All methods that are used only in the class must be private
All private methods should be below public methods
```suggestion
    private T[] grow(T[] elementArray, int increasingInSizeByInt) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            elementArray[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("this index is not correction");
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
```
Ivan95kos, line 82: move all checks to private methdos, 
example:
```
    @Override
    public T remove(int index) {
        checkIndex(index);
        checkIndexEqualsSize(index);
        T element = elements[index];
        return element;
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            System.arraycopy(elementArray, index + 1, elementArray, index, size - index);
+            elementArray[size] = null;
+            return deletedElement;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("this index is not correction");
+        }
     }
 
     @Override
     public T remove(T element) {
```
Ivan95kos, line 94: Find the index and call the remove method by index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
```
Ivan95kos, line 127:
```suggestion
        return  size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            size++;
+            System.arraycopy(elementArray, index, elementArray, index + 1, size - index - 1);
+            elementArray[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("this index is not correction");
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
```
Ivan95kos, line 50: You can use add method
example:
```
    @Override
        for(int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
    }
```

Ivan95kos (changes requested):

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
SieryiD: в останній версії здається виправив, хіба ні? 
if (deleteElement == true) {
            return remove(numberDeletedElement);
        }

src/main/java/core/basesyntax/ArrayList.java
okuzan: use this approach, no need to handle grow here - there's logic for that in add method already. addAll is just a convenient way to add multiple times, that's it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+                numberDeletedElement = a;
+                deleteElement = true;
+                break;
+            }
+        }
+        if (deleteElement == true) {
+            return remove(numberDeletedElement);
+        } else {
+            throw new NoSuchElementException("this element is missing from the T[] elementArray");
+        }
```
okuzan, line 82: this type of problem is solved in a different manner. When you found your element and entered if-body on line 73, call  remove by index with the index. No need to save it to a temporary variable and to rely on a boolean flag. On line 78, unwrap exception throwing - if we reached that point, it means nothing was found. Also, include the element into exception message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexEqualsSize(index);
+        if (size == elementArray.length) {
+            elementArray = grow(elementArray, 1);
+        }
+        size++;
```
okuzan, line 29: if you place index checker after if, you can use the standard checker

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementArray;
```
okuzan, line 7: okay as is, but elements would be more idiomatic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
     }
+
+    private T[] grow(T[] elementArray, int increasingInSizeByInt) {
```
okuzan, line 95:
```suggestion
    private T[] grow() {
```
when you will fix addAll you won't need the second parameter, first is actually always the same

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+
+    public ArrayList() {
+        elementArray = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementArray.length) {
+            //elementArray = grow(elementArray, 1);
```
okuzan, line 17: don't commit comments

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+            elementArray = grow();
+        }
+        elementArray[size] = value;
+        size++;
     }
```
```suggestion
        elementArray[size++] = value;
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        size++;
+        System.arraycopy(elementArray, index, elementArray, index + 1, size - index - 1);
+        elementArray[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int listSize = list.size();
+        for (int a = 0; a < listSize; a++) {
```
okuzan, line 38:
```suggestion
        for (int a = 0; a < list.size(); a++) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexEqualsSize(index);
+        if (size == elementArray.length) {
+            elementArray = grow();
+        }
```
```suggestion
        if (size == elementArray.length) {
            elementArray = grow();
        }
        checkIndex(index);        
```
this way you can wouldn't need the second checker method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+
+    private T[] grow() {
+        int futureSize = size + 1;
+        int newCapacity = elementArray.length;
+        while (futureSize >= newCapacity) {
+            newCapacity += (newCapacity >> 1);
+        }
+        T[] arrayPlusSize = (T[]) new Object[newCapacity];
+        System.arraycopy(elementArray, 0, arrayPlusSize, 0, size);
+        return arrayPlusSize;
```
okuzan, line 94:
```suggestion
        private void grow() {
        int newCapacity = elementArray.length * GROW_FACTOR;
        T[] resizedArray = (T[]) new Object[newCapacity];
        System.arraycopy(elementArray, 0, resizedArray, 0, size);
        elementArray = resizedArray;
```

okuzan: it's always called to change elementArray, so may as well be void, logic a bit simplified, grow_Factor extracted to constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    public ArrayList() {
+        elementArray = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementArray.length) {
+            //elementArray = grow(elementArray, 1);
+            elementArray = grow();
```
```suggestion
            grow();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        elementArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexEqualsSize(index);
+        if (size == elementArray.length) {
+            elementArray = grow();
```
```suggestion
            grow();
```

okuzan (changes requested):

olekskov (approved):
Good job

PR #992, opened 2023-06-15
array list

ozhabrovets:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+
```
okuzan, line 7: extra empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void enaughCapacity(int needCapacity) {
```
okuzan, line 94: enough spelt incorrectly. better method name - ensure capacity. It shouldn't take any arguments, internalArray.length and constants are visible here, nothing else is needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        if (needCapacity >= internalArray.length) {
+            T[] newArray = (T[]) new Object[(int) (internalArray.length * GROW_FACTORY)];
+            System.arraycopy(internalArray, 0, newArray, 0, internalArray.length);
+            internalArray = newArray;
+        }
+    }
+
+    private void growIfArrayFull() {
```
okuzan, line 102:
```suggestion
    private void growArrayIfFull() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+                size--;
+                notExistToRemove = false;
+                return element;
+            }
+        }
+        if (notExistToRemove) {
+            throw new NoSuchElementException("Element "
+                    + element + " nor exist in List");
+        }
+        return element;
```
okuzan, line 81: you don't need to rely on boolean flag here. Unwrap exception throwing from if outside of for loop, if we reach this point, it means element is not found

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T toRemoveElement = internalArray[index];
+        for (int i = index; i < size - 1; i++) {
+            internalArray[i] = internalArray[i + 1];
+        }
```
okuzan, line 60: use system array copy
image

okuzan (changes requested):
Good job on your first try!

olekskov (approved):
Good job!

PR #996, opened 2023-06-19
Made changes to the ArrayList class

Olga-Melnichenko:
Implemented add, addAll, get, set, and remove methods. Added validations for invalid indices and threw ArrayListIndexOutOfBoundsException.  Added a remove method to remove an element by value, throwing NoSuchElementException if the element is not found.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,116 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
overpathz, line 13: Value of primitive type(int) will be zero by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elements.length) {
+            expandCapacity();
+        }
+        elements[size] = value;
```
overpathz, line 21: Minor remark.
But we can do elements[size++] = value;

overpathz (approved):

PR #1003, opened 2023-06-23
meyson-hw-jv-array-list-solution

Meysonsan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int ELEMENT_NOT_FOUND = -1;
+    private T[] elements;
+    private int size;
+
+    @SuppressWarnings("unchecked")
```
Ivan95kos, line 12: In this task you must not use this annotation @SuppressWarnings
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    private void checkIndex(int index) {
```
Ivan95kos, line 17: All private methods should be below public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+            int newCapacity = currentCapacity + (currentCapacity >> 1);
+            T[] newElements = (T[]) new Object[newCapacity];
+            System.arraycopy(elements, 0, newElements, 0, size);
+            elements = newElements;
+        }
+    }
+
+    private int indexOf(T element) {
+        for (int i = 0; i < elements.length; i++) {
+            if (Objects.equals(element, elements[i])) {
```
Ivan95kos, line 112: Don’t use class Objects.
https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
```
Ivan95kos, line 67: You can rewrite like this:
```
    @Override
    public T remove(int index) {
        checkIndex(index);
        checkIndexEqualsSize(index);
        T element = elements[index];
        return element;
    }
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Meysonsan: > You can rewrite like this:
> 
> ```
>     @Override
>     public T remove(int index) {
>         checkIndex(index);
>         checkIndexEqualsSize(index);
>         T element = elements[index];
>         System.arraycopy(elements, index + 1, elements, index, --size - index);
>         return element;
>     }
> ```

I don't understand how to implement the checkIndexEqualsSize(index) method, I get two errors when passing tests
java.lang.AssertionError: Test failed! Returned value should be "Java", but was "null"
 
java.lang.Exception: Unexpected exception, expected<java.util.NoSuchElementException> but was<core.basesyntax.ArrayListIndexOutOfBoundsException>

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                    "Index " + index + " out of bounds for size: " + size);
+        }
+    }
+
+    public void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index " + index + " out of bounds for size: " + size);
+        }
+    }
```
olekskov, line 27: These method shouldn't be public.
Please change access modifiers and move the to the bottom after public methods
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException(
                    "Index " + index + " out of bounds for size: " + size);
        }
    }

    private void checkIndexForAdd(int index) {
        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException(
                    "Index " + index + " out of bounds for size: " + size);
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        int currentCapacity = elements.length;
+        if (minCapacity > currentCapacity) {
+            int newCapacity = currentCapacity + (currentCapacity >> 1);
+            T[] newElements = (T[]) new Object[newCapacity];
+            System.arraycopy(elements, 0, newElements, 0, size);
+            elements = newElements;
+        }
+    }
+
+    public int indexOf(T element) {
```
olekskov, line 101:
```suggestion
    private void ensureCapacity(int minCapacity) {
        int currentCapacity = elements.length;
        if (minCapacity > currentCapacity) {
            int newCapacity = currentCapacity + (currentCapacity >> 1);
            T[] newElements = (T[]) new Object[newCapacity];
            System.arraycopy(elements, 0, newElements, 0, size);
            elements = newElements;
        }
    }

    private int indexOf(T element) {
```

olekskov (changes requested):
Solution is fine overall, just need to fix access modifiers and methods order

olekskov (approved):
Good job

PR #1016, opened 2023-06-29
All tests are passed

andriipikhur:
Maven cleaned package

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,140 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() { // Constructor
+        elementData = new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacityInternal(size + 1);
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        rangeCheckForAdd(index);
+        ensureCapacityInternal(size + 1);
+        System.arraycopy(elementData, index, elementData, index + 1,size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int oldElementDataSize = size;
+        ensureCapacityInternal(size + list.size());
+        for (int i = 0; i < list.size(); i++) {
+            elementData[oldElementDataSize++] = list.get(i);
+        }
+        size = size + list.size();
     }
+
+    //////// Methods for addition
+
+    private void ensureCapacityInternal(int minCapacity) {
+        if (minCapacity - elementData.length > 0) {
+            grow(minCapacity);
+        }
+    }
+
+    private void grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        if (newCapacity - minCapacity < 0) {
+            newCapacity = minCapacity;
+        }
+        if (newCapacity - MAX_ARRAY_SIZE > 0) {
+            throw new OutOfMemoryError();
+        }
+        elementData = Arrays.copyOf(elementData, newCapacity);
+    }
+
+    private void rangeCheckForAdd(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    /////////////// GET - SET //////////////
 
     @Override
     public T get(int index) {
-        return null;
+        rangeCheck(index);
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        rangeCheck(index);
+        elementData[index] = value;
     }
+
+    private void rangeCheck(int index) {
+        if (index < 0 || index >= size) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            elementData[oldElementDataSize++] = list.get(i);
+        }
+        size = size + list.size();
+    }
+
+    //////// Methods for addition
+
+    private void ensureCapacityInternal(int minCapacity) {
+        if (minCapacity - elementData.length > 0) {
```
Rommelua, line 45:
```suggestion
        if (minCapacity > elementData.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+            grow(minCapacity);
+        }
+    }
 
+    private void grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        if (newCapacity - minCapacity < 0) {
+            newCapacity = minCapacity;
+        }
```
Rommelua, line 55:
```suggestion
        newCapacity = Math.max(minCapacity, newCapacity);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        if (newCapacity - MAX_ARRAY_SIZE > 0) {
+            throw new OutOfMemoryError();
```
Rommelua, line 57: 1. is it a possible condition?
2. never throw Error - use RuntimeEx with an informative message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        }
+        elementData = Arrays.copyOf(elementData, newCapacity);
```
Rommelua, line 59: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
+        if (element == null) {
+            for (int i = 0; i < size; i++) {
+                if (elementData[i] == null) {
+                    fastRemove(i);
+                    return element;
```
Rommelua, line 108:
```suggestion
                    return remove(i);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,140 @@
...
+                }
+            }
+            throw new NoSuchElementException("Element" + element + "is not present in ArrayList");
+        }
         return null;
     }
 
+    private void fastRemove(int index) {
```
Rommelua, line 123: I do not think we need it)

Rommelua (changes requested):
Good work

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: Fixed. Thank you)   this is embarrassing)))

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: Upgraded

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: This method is used for removing records. It does not look like it is suitable for addition where we need new size. New intermediate variable will be needed as destination.

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: Yes, it was possible to make it shorter.

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: I do not understand this, and it doesn't compile with return remove(i).
Java ArrayList class explains that this method is to "@return the element that was removed from the list".

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: Added check for not to attempt to grow new Array above MAX int.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,136 @@
...
+        boolean success = false;
+        int numMoved;
+        for (int i = 0; i < size; i++) {
+            if (element == null || element.equals(elementData[i])) {
+                numMoved = (size - 1) - i;
+                System.arraycopy(elementData, i + 1, elementData, i, numMoved);
+                elementData[--size] = null;
+                success = true;
+                return element;
+            }
```
aaananas, line 119: reuse remove(i)

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: image
it does not pass tests. Doesn't this cause recursion?

andriipikhur: I meant with remove (i). Also it doesn't make sense to return removed object in method where object is removed by object-citeria. It would make sense to make such return in remove(index).

aaananas, line 119:
```suggestion
    public T remove(T element) {
        for (int i = 0; i < size; i++) {
            if (element == elementData[i] || element != null && element.equals(elementData[i])) {
                return remove(i)l
            }
```

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
andriipikhur: Understood)

Rommelua (approved):

PR #1017, opened 2023-06-30
implemented ArrayList

JeniaSan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void changeSize() {
```
nacenik, line 91: according to the name - this method should change size every time but it's not true. 
Rename this method to `ensureCapacity`, `checkCapacity` or sth like this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        if (size == capacity) {
+            capacity *= RESIZE_FACTOR;
+            T[] tmp = elements;
+            elements = (T[]) new Object[capacity];
+            System.arraycopy(tmp, 0, elements, 0, size);
```
nacenik, line 96: only this logic is related to the name `changeSize` but, according to the name, it could decrease too, so, better to create an additional method with name `grow`, `increaseCapacity`, or sth like this for this logic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't access element at index " + index);
+        }
     }
 }
+
```
nacenik, line 112: redundant line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkRemoveIndex(index);
+        final T removeElement = elements[index];
+        for (int i = index; i < size - 1; i++) {
+            elements[i] = elements[i + 1];
+        }
+        elements[size - 1] = null;
```
nacenik, line 57: but what about using `System.arraycopy` to delete the last element? That's quicker and easier than a loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     public T remove(T element) {
-        return null;
+        if (element == null) {
+            size--;
+            return null;
+        }
+        T removeElement = null;
+        for (int i = 0; i < elements.length; i++) {
+            if (ObjectUtil.equals(elements[i], element)) {
+                removeElement = remove(i);
```
nacenik, line 71: you can just return the value here, instead of using a new variable and doing an additional check on values null, cause your ArrayList could have null values inside, and if users want to delete this null from your ArrayList, they will get an exception, but it shouldn't

nacenik (changes requested):

nacenik (approved):
LGTM! 👍

PR #1021, opened 2023-07-02
Version1-Stas

stas83ggg:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((element == null && elementData[i] == null)
+                    || (element != null && element.equals(elementData[i]))) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    public void checkIndex(int index) {
```
okuzan, line 100: The method 'checkIndex(int index)' should be private as it is an internal helper method not intended for external use.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        int newCapacity = (int) (elementData.length * FACTOR_GROW);
+        T[] newElementData = (T[]) new Object[newCapacity];
+        System.arraycopy(elementData, 0, newElementData, 0, size);
+        elementData = newElementData;
+    }
+
+    private int indexOf(T element) {
+        for (int i = 0; i < size; i++) {
+            if ((element == null && elementData[i] == null)
+                    || (element != null && element.equals(elementData[i]))) {
```
okuzan, line 93:
```suggestion
            if (element == elementData[i]
                    || element != null && element.equals(elementData[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    public void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 102: include size in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            expandCapacity();
+        }
```
```suggestion
        ensureCapacity();
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
stas83ggg: Where did You get "ensureCapacity()" method ? I don't have such method.
And if I delete checking "if (size == elementData.length)" I will get test failed "checkingResize
Снимок экрана 2023-07-05 в 21 44 52
"

That is why I think it will be better to leave checking size.

Rommelua (approved):

PR #1024, opened 2023-07-04
Homework done/add implementation to custom ArrayList

VitaliiLeshchenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int findIndexByValue(T value) {
+        int i;
+        for (i = 0; i < size; i++) {
+            if ((value == null && elementData[i] == null)
+                    || (value != null && value.equals(elementData[i]))) {
```
liliia-ponomarenko, line 82:
```suggestion
            if (value == null && elementData[i] == null
                    || value != null && value.equals(elementData[i])) {
```
these brackets are redundant https://introcs.cs.princeton.edu/java/11precedence/

liliia-ponomarenko (approved):
🔥

PR #1027, opened 2023-07-05
jv-array-list

Alex-Solod:
Created ArrayList class which implements the List interface.
Override the methods.
Throw custom ArrayListIndexOutOfBoundsException in checkIndex and checkIndexForAdd methods.
NoSuchElementException in remove(T element) method.


kozhukhovsky (approved):
Very good job!

PR #1028, opened 2023-07-05
Hw

padu12:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_SIZE = 10;
+    private Object[] innerArray;
+    private int size = 0;
+
+    public ArrayList() {
+        innerArray = new Object[DEFAULT_ARRAY_SIZE];
+        size = 0;
+    }
```
JeniaSan, line 13: Do you really need constructor ? You can initialize innerArray on 7th line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
-        return null;
+        T removedItem = null;
+        for (int i = 0; i < size; i++) {
+            T castedObject = (T) innerArray[i];
+            if ((castedObject == element) || (castedObject != null
+                    && castedObject.equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("There is no element " + element);
```
JeniaSan, line 78: you can simplify this method , just use remove(int index) , and give in this method the index of the wanted object  (move the index search method into a separate method)

JeniaSan (changes requested):
nice solution , but has some minor mistakes

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_SIZE = 10;
+    private Object[] innerArray = new Object[DEFAULT_ARRAY_SIZE];
+    private int size = 0;
```
JeniaSan, line 8: int have default 0 value, you don't need to initialize this variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_SIZE = 10;
```
JeniaSan, line 6: better call this constant capacity, because size and capacity are different concepts

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        if (size == 0) {
+            add(value);
+        } else if (size == 1) {
+            innerArray[1] = innerArray[0];
+            innerArray[0] = value;
+            size++;
+        } else {
+            System.arraycopy(innerArray, index, innerArray, index + 1, size - index);
+            innerArray[index] = value;
+            size++;
```
JeniaSan, line 32: you can simplify this method like this 
```suggestion
    checkIndex(index);
    growIfArrayFull();
    System.arraycopy(innerArray, index, innerArray, index + 1, size - index);
    innerArray[index] = value;
        size++;
```

JeniaSan (changes requested):
good job, but you need to fix some minor mistakes

JeniaSan (approved):
LGTM

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] innerArray = new Object[DEFAULT_CAPACITY];
```
nacenik, line 7: All non-constant fields should be initialized in the constructor
Also, make it final

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size && index != 0) {
+            throw new ArrayListIndexOutOfBoundsException("There is no element at index " + index);
+        }
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("You can't input a negative index. "
+                    + index + " is below zero!");
+        }
```
nacenik, line 86: what about merging these blocks to one if block? 
Also, be careful with message. cause you are working with index, not with element here 
```suggestion
        if (index >= size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index + "out of bounds");
        }
```

nacenik (changes requested):
Good solution, let's improve it :)

nacenik (approved):
LGTM! 👍

PR #1031, opened 2023-07-07
Hope it's done

mate-academy-student-2023:
Please check it. Thank you!

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_SIZE = 10;
+
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[INITIAL_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
-
+        validateIndex(index, size + 1);
+        ensureSize(size + 1);
+        if (index < size) {
+            System.arraycopy(elements, index, elements, index + 1, size - index);
+        }
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null || list.size() == 0) {
+            return;
+        }
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get(int index) {
-        return null;
+        validateIndex(index);
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        validateIndex(index);
+        elements[index] = value;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        final T value = (T) elements[index];
+        size--;
+        if (index < size) {
+            System.arraycopy(elements, index + 1, elements, index, size - index);
+        }
+        elements[size] = null;
+        return value;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((elements[i] == element)
+                    || (elements[i] != null && elements[i].equals(element))) {
+                remove(i);
+                return element;
+            }
+        }
+        throw new NoSuchElementException("No such element in the list: " + element);
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureSize(int size) {
+        if (size >= elements.length) {
+            Object[] newArray = new Object[elements.length * 3 / 2];
+            System.arraycopy(elements, 0, newArray, 0, elements.length);
+            elements = newArray;
+        }
+    }
+
+    private void validateIndex(int index) {
+        validateIndex(index, size);
+    }
+
+    private void validateIndex(int index, int upperBound) {
+        if (index >= upperBound) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid: "
+                    + index + "; index greater than array size; array size: " + size);
+        }
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid: "
+                    + index + "; index should be positive; array size: " + size);
+        }
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_SIZE = 10;
+
+    private Object[] elements;
```
Ivan95kos, line 8: You must use the parameterized type `T`
```suggestion
    private T[] elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        validateIndex(index);
+        elements[index] = value;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
```
Ivan95kos, line 55: You should not use this annotation
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null || list.size() == 0) {
+            return;
+        }
```
Ivan95kos, line 35: It is better to throw exeption if list is null or empry

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size == 0;
+    }
+
+    private void ensureSize(int size) {
+        if (size >= elements.length) {
+            Object[] newArray = new Object[elements.length * 3 / 2];
+            System.arraycopy(elements, 0, newArray, 0, elements.length);
+            elements = newArray;
+        }
+    }
```
Ivan95kos, line 95: use constants instead of numbers
```
    private void grow() {
        if (elements.length == size) {
            T[] newArrays = (T[]) new Object[(int) (elements.length * GROW_FACTORY)];
            System.arraycopy(elements, 0, newArrays, 0, elements.length);
            elements = newArrays;
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[INITIAL_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        add(value, size);
```
Ivan95kos, line 17: method `add` should have own realization

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        validateIndex(index, size + 1);
```
Ivan95kos, line 22: All magic numbers, strings, symbols in your code should be constants.
[this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

Ivan95kos (changes requested):

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: The one here is not a "magic number" but the extent to size of the array to tell that we should check the index bound including next element,

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Is it mentioned somewhere?
According to this article I've intentionally created add() in that way - 

https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html

Don’t create repeating code.
If the logic of your code repeats - move it to the separate private method. For example, the index check for get(), set() and remove() methods should be moved to the separate method.
There is even a software design principle called [DRY](https://dzone.com/articles/software-design-principles-dry-and-kiss) that urges not to repeat yourself.

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Agree.

mate-academy-student-2023: It makes sense.
Thanks!

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Is it a new requirement?
No word on it nor within README nor the comments.
Please advise. Thanks!

Create your own ArrayList class which implements the List interface. Override the methods required.

Make your ArrayList behave the same way as java.util.ArrayList (it should have the default capacity, the internal array should grow 1.5 times when it is full, etc.).

Throw custom ArrayListIndexOutOfBoundsException in case the index passed to any of the methods is invalid. Throw NoSuchElementException in remove(T element) if there is no such element present.

https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html

mate-academy-student-2023: Anyway, the "elements" attribute is not shared beyond the class implementation.

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Do you mean there should not be code pieces that cause unchecked warnings?

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Agree for "null" input.
Questionable for the empty list.

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Following your logic...
Should I change these lines? i.e. replace 1 to a constant in "index + 1"?

System.arraycopy(elements, index + 1, elements, index, size - index);

please advise. thanks!

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Or... do you want me to play with reflection? np.

mate-academy-student-2023: Ok, Element class to hold value added.
Does it look better? ;-)

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: There are still unchecked warnings by the compiler.
Do you suppose it is better to keep it?

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: See another version based on reflection.
Did you really want it to be that way? ;-)

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Ok, as you wish. 
Committed. 
:-)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null) {
+            throw new NullPointerException("Input list is null");
+        }
+        if (list.size() == 0) {
+            return;
+        }
```
olekskov, line 37: Code makes no impact, if list size is 0 then next cycle is not triggered and we have same result
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+    }
+
+    private T[] createArray(T value, int arraySize) {
+        if (value == null && elements != null && elements.length > 0) {
+            value = elements[0];
+        }
+        return (T[]) Array.newInstance((value != null)
+                            ? value.getClass()
+                            : Object.class, arraySize);
+    }
```
olekskov, line 108: I can't get idea why simple things are such over-engeneered.
Why we create our list without inner array and wait for add() method to be called to make it.
Add constructor and make methods simplier:
```suggestion
    private void addCellForNewElement() {
        size++;
        if (size > elements.length) {
            T[] newArray = (T[]) Object[(int) (elements.length * GROW_FACTOR)];
            System.arraycopy(elements, 0, newArray, 0, elements.length);
            elements = newArray;
        }
    }
    
    //and add constructor to the top
    public ArrayList() {
        elements = (T[]) Object[INITIAL_SIZE];
    }
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Right. Thank you!

src/main/java/core/basesyntax/ArrayList.java
mate-academy-student-2023: Well, your mate mentioned I have to use T objects, not "Object" type.
So, what I did was just an approach to get an actual T type object array when it is possible.
In other words, with String objects, it created a real String[] array behind the scenes. Not an Object[] one.
Actually, my original version included the way you mentioned except for the type of elements array.
It was Object[], not T[]. Having it as a T[] initialized by Object[] doesn't look better than the Object[] elements array.
Not sure what was the idea to tell me to use T[] without initializing it with T[]...
https://github.com/mate-academy/jv-array-list/pull/1031#discussion_r1257294590

Any ideas?

Ivan95kos (approved):
If you disagree with comments or have additional questions
It is better to ask us in general chat of slack
Unfortunately, we don't get comments from GITGUB

PR #1041, opened 2023-07-11
created ArrayList class

Someboty:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    public ArrayList(int capacity) {
+        if (capacity > 0) {
+            data = (T[]) new Object[capacity];
+        } else if (capacity == 0) {
+            data = (T[]) new Object[]{};
+        } else {
+            throw new ArrayListIndexOutOfBoundsException(NEGATIVE_INDEX_ERROR);
+        }
+    }
+
```
Den-moroz, line 28: I'm not sure that you need to create two constructors. I think the second constructor is useless, also if u throw error it should be IllegalArgumentException.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
-
+        if (size + list.size() > data.length) {
+            this.ensureCapacity();
+            this.addAll(list);
+        }
+        for (int i = size; i < (size + list.size()); i++) {
+            data[i] = list.get(i - size);
+        }
+        size += list.size();
     }
```
Den-moroz, line 63: this can be done much more easily with a single loop.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        isIndexValid(index + 1);
+        final T result = data[index];
+        if (index == size - 1) {
+            data[index] = null;
+        } else {
+            System.arraycopy(data, index + 1, data, index, size - index);
+            data[size - 1] = null;
+        }
+        size--;
+        return result;
```
Den-moroz, line 88: first use checkIndex instead isIndexValid() then you can not use the if condition, instead use
```System.arraycopy(data, index + 1, data, index, size - index - 1);
        data[size - 1] = null;
        size--;
        return result;

File: src/main/java/core/basesyntax/ArrayList.java [hunk 4]
```diff
@@ -1,48 +1,132 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return this.size() == 0;
```
Den-moroz, line 114: you can just use 'size'

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    }
+
+    private void ensureCapacity() {
```
Den-moroz, line 117: I would rather call this method increaseCapacity.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        T[] newData = (T[]) new Object[(int)(size * CAPACITY_MULTIPLY_BY)];
+        System.arraycopy(data, 0, newData, 0, size);
+        data = newData;
+    }
```
Den-moroz, line 121: I would not advice using size here, I would create a separate capacity value.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int NATURAL_NUMBER_BORDERLINE = 0;
```
Den-moroz, line 6: You can do without this variable.

Den-moroz (changes requested):
Let's fix some mistakes)

src/main/java/core/basesyntax/ArrayList.java
Someboty: I think it should stay because the original ArrayList has it. But the exception type is wrong, agree

Den-moroz: I'm not saying that it's a mistake, just that it's not necessary because this method is not in the interface and you don't use it anywhere else

src/main/java/core/basesyntax/ArrayList.java
Someboty: I hope you mean that the if() check is redundant. Looks like it is.

Because size update is important

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        }
+        size += list.size();
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isIndexValid(index + 1);
+        return data[index];
     }
```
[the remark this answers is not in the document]
Someboty: But why if only difference is index +1?

src/main/java/core/basesyntax/ArrayList.java
Den-moroz:
```
            add(list.get(i));
        }
```
that's all what u need

src/main/java/core/basesyntax/ArrayList.java
Den-moroz: I expected something like this
```
        checkIndex(index);
        final T removedElement = data[index];
        System.arraycopy(data, index + 1, data, index, size - index - 1);
        data[size - 1] = null;
        size--;
        return removedElement;
```
but if you don't agree, you can leave it as it was and the mentor will decide  what is right. It's ok)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+            if ((data[i] != null && data[i].equals(element))
+                    || (data[i] == null && element == null)) {
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException(INDEX_OUT_OF_SIZE_ERROR);
+        }
+        return this.remove(index);
```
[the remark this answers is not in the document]
Someboty: But why should we have separate method, that works in one place? It`s not so big to separate it

src/main/java/core/basesyntax/ArrayList.java
Someboty: Wow, it really easier, than mine :`(

src/main/java/core/basesyntax/ArrayList.java
Someboty: I'm really mad at myself because I tried to write something like this and it didn't work, so I added this check. And now it works

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
```
Den-moroz, line 90: u can replace it on a much simpler and correctly:
```
int i = -1;
            if (ObjectUtil.equals(data[i], element)) {
                index = i;
            }
        }
if (i == -1) {
}
return remove(index);
```
also u can't use equals(), read this in common mistakes.

Den-moroz (changes requested):
Good job, let's fix the remaining errors!
After you have added a new commit, you have to choose "again st review" you can read more info in slack, also I wouldn't recommend  you calling commits such a long names.
In the end, I would still recommend creating a checkIndex and checkCapacity method, but this is wishful thing.

src/main/java/core/basesyntax/ArrayList.java
Someboty: ObjectUtil almost same as Objects, you cant`t use it
What do you mean I can`t use equals? There is not such thing in common mistakes, only about Objects

Someboty: I don`t think I should remove break too. It optimises program a little bit

Den-moroz (approved):
Ok, your choice. Good job!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int capacity) {
+        if (capacity > 0) {
+            data = (T[]) new Object[capacity];
+        } else if (capacity == 0) {
+            data = (T[]) new Object[]{};
```
nacenik, line 22: if your capacity == 0, you should create arrayList with default capacity
```suggestion
            data = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        System.arraycopy(data, 0, newData, 0, size);
+        data = newData;
+    }
+
+    private void isIndexValid(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_OUT_OF_SIZE_ERROR);
+        }
     }
+
```
nacenik, line 114: redundant line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    private void increaseCapacity() {
+        T[] newData = (T[]) new Object[(int)(size * CAPACITY_MULTIPLY_BY)];
+        System.arraycopy(data, 0, newData, 0, size);
+        data = newData;
+    }
+
+    private void isIndexValid(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_OUT_OF_SIZE_ERROR);
```
nacenik, line 111: add info about which index is out of bounds

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_MULTIPLY_BY = 1.5;
+    private static final String NEGATIVE_INDEX_ERROR = "Index should be natural number";
+    private static final String INDEX_OUT_OF_SIZE_ERROR = "Index out of list length";
```
```suggestion
    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index out of bounds";
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_MULTIPLY_BY = 1.5;
+    private static final String NEGATIVE_INDEX_ERROR = "Index should be natural number";
```
nacenik, line 8:
```suggestion
    private static final String NEGATIVE_INDEX_MESSAGE = "Index should be natural number";
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            throw new IllegalArgumentException(NEGATIVE_INDEX_ERROR);
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= data.length) {
+            increaseCapacity();
+        }
```
nacenik, line 32: You can create new private method called `checkIndex` with this logic and reuse it, cause you have the same block of code in `add(T value, int index)` method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_MULTIPLY_BY = 1.5;
+    private static final String NEGATIVE_INDEX_ERROR = "Index should be natural number";
+    private static final String INDEX_OUT_OF_SIZE_ERROR = "Index out of list length";
+
+    private T[] data;
```
nacenik, line 11: make it final

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((data[i] != null && data[i].equals(element))
+                    || (data[i] == null && element == null)) {
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException(INDEX_OUT_OF_SIZE_ERROR);
+        }
```
nacenik, line 89: is this out of bounds or, maybe, you should show a message, that element doesn't exist? Don't forget to make your exception message informative

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_MULTIPLY_BY = 1.5;
```
nacenik, line 7:
```suggestion
    private static final double CAPACITY_MULTIPLIER = 1.5;
```

nacenik (changes requested):

nacenik (commented):
Also, add custom exception according to the task

src/main/java/core/basesyntax/ArrayList.java
Someboty: But then I won't be able to overwrite it with the new capacity

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        return size == 0;
+    }
+
+    private void increaseCapacity() {
+        T[] newData = (T[]) new Object[(int)(size * CAPACITY_MULTIPLIER)];
+        System.arraycopy(data, 0, newData, 0, size);
+        data = newData;
+    }
+
+    private void isIndexValid(int index) {
```
KseniiaMakarova, line 110:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    (index < 0) ? NEGATIVE_INDEX_MESSAGE : INDEX_OUT_OF_SIZE_ERROR);
+        }
+    }
+
+    private boolean checkIndex() {
```
KseniiaMakarova, line 117:
```suggestion
    private boolean needsIncreasing() {
```
looks more logical

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            T item = data[i];
+            if ((item != null && item.equals(element))
+                    || (item == null && element == null)) {
```
KseniiaMakarova, line 83:
```suggestion
            if (item != null && item.equals(element)
                    || item == element == null) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+                index = i;
+                break;
```
KseniiaMakarova, line 85:
```suggestion
                 return remove(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException(ELEMENT_DOES_NOT_EXISTS);
+        }
+        return remove(index);
```
KseniiaMakarova, line 91:
```suggestion
        throw new NoSuchElementException(ELEMENT_DOES_NOT_EXISTS);
```

KseniiaMakarova (changes requested):

andrii-hoienko (approved):
Good job

PR #1042, opened 2023-07-11
implemented ArrayList and added some additional methods to it

Zel1oy:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,149 @@
...
+        if (size == elementData.length) {
+            elementData = grow();
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        rangeCheckForAdd(index);
```
holsoni, line 27: You can try not creating another method for checking indexes for add, but  using if clause.
```suggestion
        if (index != size) {
            rangeCheck(index);
        }   
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,149 @@
...
+            return;
+        }
+        Object[] elementData;
+        final int s;
+        if (numNew > (elementData = this.elementData).length - (s = size)) {
+            elementData = grow(s + numNew);
+        }
+        System.arraycopy(a, 0, elementData, s, numNew);
+        size = s + numNew;
+    }
```
holsoni, line 66: Try using add() method here. It has similar logic and will automatically grow an array if it will needed. Add every element of list object using loop.

src/main/java/core/basesyntax/List.java
```diff
@@ -18,4 +18,6 @@ public interface List<T> {
     int size();
 
     boolean isEmpty();
+
+    Object[] toArray();
```
holsoni, line 22: If you rewrite the addAll() method, this line will be redundant.

holsoni (changes requested):
Excellent solution, let's improve it 👍

OleksandrSaven (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,149 @@
...
+                    }
+                }
+            } else {
+                for (; i < size; i++) {
+                    if (element.equals(es[i])) {
+                        break found;
+                    }
+                }
+            }
+            throw new NoSuchElementException();
```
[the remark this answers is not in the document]
Zel1oy: The name of Exception says everyting user gotta know already

holsoni (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] DEFAULTCAPACITY_EMPTY_ELEMENTDATA = {};
```
nacenik, line 8: redundant constant 
your array shouldn't be empty.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    private Object[] elementData;
+    private int size = 0;
+
+    public ArrayList() {
+        elementData = DEFAULTCAPACITY_EMPTY_ELEMENTDATA;
```
nacenik, line 13: you should create array with default capacity.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            elementData = grow();
+        }
```
nacenik, line 20: create new method with this block of code, call it `growIfNeeded` and reuse it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+
+    private Object[] grow() {
+        return grow(size + 1);
+    }
+
+    private Object[] grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0 || elementData != DEFAULTCAPACITY_EMPTY_ELEMENTDATA) {
+            int newCapacity = oldCapacity + (oldCapacity >> 1);
+            return elementData = Arrays.copyOf(elementData, newCapacity);
```
nacenik, line 127: don't use `Arrays.copyOf`
[Use System.arraycopy](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        } else {
+            return elementData = new Object[Math.max(DEFAULT_CAPACITY, minCapacity)];
+        }
```
nacenik, line 130: you shouldn't have condition here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+            System.arraycopy(es, i + 1, es, i, newSize - i);
+        }
+        es[size = newSize] = null;
+    }
+
+    private Object[] grow() {
+        return grow(size + 1);
+    }
+
+    private Object[] grow(int minCapacity) {
```
nacenik, line 123: why do you need to return the array from this method? It should be void. Because now you define the same variable 2 times

nacenik, line 123: why do you need 2 methods? 
```suggestion
    private void grow() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        rangeCheck(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        rangeCheck(index);
+        final Object[] es = elementData;
+        T oldValue = (T) es[index];
```
nacenik, line 61: why do you need additional variable? 
```suggestion
        T oldValue = (T) elementData[index];
```

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
nacenik: not fixed. Add informative message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        return size == 0;
+    }
+
+    private void rangeCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private String outOfBoundsMsg(int index) {
```
nacenik, line 107:
```suggestion
    private String createIndexOutOfBoundsMsg(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+                for (; i < size; i++) {
+                    if (element.equals(es[i])) {
+                        break found;
+                    }
+                }
+            }
+            throw new NoSuchElementException();
+        }
+        fastRemove(es, i);
+        return element;
```
nacenik, line 88: You can simplify: 

- find element index. 
- call method `remove(int index)`
- if element not found - throw an exception.

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+                    }
+                }
+            } else {
+                for (; i < size; i++) {
+                    if (element.equals(es[i])) {
+                        break found;
+                    }
+                }
+            }
+            throw new NoSuchElementException("Element you want to remove does not exit");
```
nacenik, line 81: [using labels is a vary bad practice. you can do the same using just one loop and if inside this loop.](https://stackoverflow.com/questions/33689466/labels-in-java-bad-practice)

nacenik (changes requested):
Good solution, but let's avoid using labels. it's a bad practice

Vasul-Pankiv (approved):

PR #1047, opened 2023-07-11
added an implementation of the custom ArrayList

ivan0dyatlyukkk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int indexOfElement = getIndexOfValue(element);
+        if (indexOfElement != -1) {
+            return remove(indexOfElement);
+        }
+        throw new NoSuchElementException();
```
IvanFadieiev, line 90: Maybe it's better to add some message for this exception?

IvanFadieiev (changes requested):
Nice! Some small details to change.

IvanFadieiev (approved):

andrii-hoienko (changes requested):
Good solution, respect for using recursion, but there are some small issues to resolve

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+
+    private void isIndexValidForAdding(int index) {
+        if (index > currentCapacity || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index - " + index
+                                                        + " is invalid to add to the arraylist!");
+        }
+    }
+
+    private int getIndexOfValue(T value) {
+        for (int i = 0; i < currentCapacity; i++) {
```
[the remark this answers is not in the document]
ivan0dyatlyukkk: Sorry, I'm a bit confused. If I use the for-each loop in this method, how will I find an index of the suitable element?
Or will be better if I use like it?
```suggestion
        int count = 0;
        for (T element : elements){
             if (value.equals(element)){
                 return count;
              }
              count++;
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,154 @@
...
+        if (index >= currentSize || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index - " + index
+                                                        + " is invalid to use!");
+        }
+    }
+
+    private void isIndexValidForAddingValues(int index) {
+        if (index > currentSize || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The index - " + index
+                                                        + " is invalid to add to the arraylist!");
```
nacenik, line 142:
```suggestion
                                                        + " is out of bounds");
```

nacenik (approved):
LGTM! 👍

PR #1049, opened 2023-07-11
implemented ArrayList class and methods

holsoni:


Zel1oy (approved):
Approved. Good and simple implementation)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        int newSize = size - 1;
+        T element = array[index];
+        System.arraycopy(array, index + 1, array, index, newSize - index);
```
IhorKalaur, line 60: What if array.length==size. In this case we will not have anything to replace the last element and it will be unchanged

IhorKalaur (changes requested):
Good, but there are a few details that need to be worked out.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] array;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_CAPACITY];
```
```suggestion
        array = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= array.length) {
+            grow(size);
```
KseniiaMakarova, line 17:
```suggestion
            grow();
```
size is available in private method because it is a field

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    private T[] array;
+
+    public ArrayList() {
+        this.array = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= array.length) {
```
KseniiaMakarova, line 16:
```suggestion
        if (size == array.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public int getIndexOf(T element) {
```
KseniiaMakarova, line 84: should this helper method be private?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((array[i] == element) || array[i] != null && array[i].equals(element)) {
```
KseniiaMakarova, line 86:
```suggestion
            if (array[i] == element || array[i] != null && array[i].equals(element)) {
```
&& is evaluated before ||

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/List.java
```diff
@@ -18,4 +18,6 @@ public interface List<T> {
     int size();
 
     boolean isEmpty();
+
+    int getIndexOf(T element);
```
nacenik, line 22: don't add new method to the interface. This method is helper method.

src/main/java/core/basesyntax/ArrayList.java
nacenik: not fixed.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            grow();
+        }
```
nacenik, line 18: you can create a new method for this block of code and reuse it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    @Override
```
nacenik, line 84:
```suggestion
```

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private void grow() {
+        int newCapacity = size + (size >> 1);
+        T[] oldArray = array;
+        array = (T[]) new Object[newCapacity];
+        System.arraycopy(oldArray, 0, array, 0, size);
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds!");
```
nacenik, line 104: what index? Add more info to the message

nacenik (approved):
LGTM! 👍

PR #1050, opened 2023-07-11
implemented default methods in ArrayList class

CodeWebWeaver:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int requiredCapacity = size + list.size();
+        ensureCapacity(requiredCapacity);
+        for (int i = 0; i < list.size(); i++) {
+            elementData[size++] = list.get(i);
+        }
```
ivan0dyatlyukkk, line 44: I think that we can reuse the previous method add(int index). It will be better if we are talking about the DRY principle
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

ivan0dyatlyukkk (changes requested):
It's a nice solution! However, I spotted a slight issue. I hope you fix it ASAP :)

src/main/java/core/basesyntax/ArrayList.java
Skarbey: Thanks, agree

ivan0dyatlyukkk (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size == 0;
+    }
+
+    private void ensureCapacity(int requiredCapacity) {
+        if (requiredCapacity > elementData.length) {
+            int newCapacity = (int) (elementData.length * GROWTH_FACTOR);
+            if (newCapacity < requiredCapacity) {
+                newCapacity = requiredCapacity;
+            }
+            elementData = Arrays.copyOf(elementData, newCapacity);
```
nacenik, line 94: don't use `Arrays.copyOf`
[use `System.arraycopy`](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-systemarraycopy-to-move-your-array-elements)

nacenik (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+    private int size;
+    private T[] elementData;
+
+    public ArrayList() {
+        this(DEFAULT_CAPACITY);
+    }
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity >= 0) {
+            this.elementData = (T[]) new Object[initialCapacity];
```
nacenik, line 18: so, count you have array with size 0? How will it be increased? 
```suggestion
        if (initialCapacity >= DEFAULT_CAPACITY) {
            this.elementData = (T[]) new Object[initialCapacity];
         } else if (initialCapacity > 0) {
            this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

nacenik (approved):
Good solution, you can improve it if you want :)

src/main/java/core/basesyntax/ArrayList.java
Skarbey: I really didn't keep in mind that. Thanks!

PR #1055, opened 2023-07-11
implemented methods to own custom ArrayList class

ArtemReizer:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        if (elementData.length < size + 1) {
+            resize();
+        }
+    }
+
+    private void resize() {
+        newCapacity = (int) (elementData.length * RESIZE_CONSTANT);
+        Object[] temp = elementData;
+        elementData = new Object[newCapacity];
+        System.arraycopy(temp,0,elementData,0,size);
```
Vladiclevch, line 114: Hi, I think you can combine these two methods into one using if else. Everything else looks good i think.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        }
+    }
+
+    private void resizeCheck() {
+        if (elementData.length < size + 1) {
+            resize();
+        }
+    }
+
+    private void resize() {
```
ZVIIR, line 110: I think it'll be better to rename methods. Resize means, that you can resize your storage to a bigger or lower value. But your methods are used only for growth.

Vladiclevch (changes requested):
Try to combine these two methods into one using if else.

ZVIIR (changes requested):
Great solution, but need some changes.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        newCapacity = (int) (elementData.length * RESIZE_CONSTANT);
```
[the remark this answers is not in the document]
ZVIIR: made this value local*

Vladiclevch (approved):
I think code looks good

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double RESIZE_CONSTANT = 1.5;
+    private static final int EMPTY_SIZE = 0;
+    private Object[] elementData;
```
andrii-hoienko, line 9: It is better to use T[] elementData here to avoid redundant casts in your code below

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    }
+
+    private void indexOutOfBoundCheck(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " is out of bounds " + size);
+        }
+    }
+
+    private void resize() {
```
andrii-hoienko, line 104: name resize means that you will be resizing array list every time this method called, but this is not true

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    }
+
+    private void checkForAdd(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " is out of bounds " + size);
+        }
+    }
+
+    private void indexOutOfBoundCheck(int index) {
```
andrii-hoienko, line 97: Please think how to unite indexOutOfBoundCheck and checkForAdd methods in one

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " is out of bounds " + size);
+        }
+    }
+
+    private void resize() {
+        if (elementData.length == size) {
+            newCapacity = (int) (elementData.length * RESIZE_CONSTANT);
+            Object[] temp = elementData;
```
andrii-hoienko, line 107: bad variable name, not clear what stored here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+                elementData, index + 1,
+                size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
andrii-hoienko, line 37: It will be better if you will call metod add in addAll to avoid duplication

andrii-hoienko, line 37: could be replaced with foreach

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        checkForAdd(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexOutOfBoundCheck(index);
+        final T elementToRemove = (T) elementData[index];
+        for (int i = index; i < size - 1; i++) {
```
andrii-hoienko, line 62: It is possible to use System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double RESIZE_CONSTANT = 1.5;
+    private static final int EMPTY_SIZE = 0;
+    private Object[] elementData;
+    private int size;
+    private int newCapacity;
```
andrii-hoienko, line 11: is it necessary to store this variable on class level?

andrii-hoienko (changes requested):
Good, but let`s improve your solution

src/main/java/core/basesyntax/ArrayList.java
nacenik: not fixed. Instead of doing copy in loop use `System.arraycopy`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    }
+
+    private void indexOutOfBoundCheck(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " is out of bounds " + size);
+        }
+    }
+
+    private void grow() {
```
nacenik, line 112: name said that you array grow every time you call this method but it doesn't. Rename it. 
```suggestion
    private void growIfFull() {
```

nacenik (changes requested):
Good solution, let improve it :)

Vasul-Pankiv (approved):

PR #1056, opened 2023-07-11
added solution

IhorKalaur:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        growIfNeeded(list);
+        int listSize = list.size();
+        for (int i = 0; i < listSize; i++) {
+            elements[size + i] = list.get(i);
```
Pafmortim, line 44: Maybe you can use your add() method here. in this case you wouldn't need  growIfNeeded() overloading. But as soon I'm not sure I leave it fo mentor judging

Pafmortim (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private transient Object[] elements;
```
andrii-hoienko, line 8: This should be 'T[] elements' to not cast evety time you need interaction wich elements content

andrii-hoienko, line 8: Why this is transient ?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int initCapacity) {
+        if (initCapacity <= 0) {
+            throw new IllegalArgumentException(
+                    "Can't create array with negative length");
```
andrii-hoienko, line 18: message says only about negative initCapacity but there will be error with initCapacity = 0

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        }
+        elements = new Object[initCapacity];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfNeeded();
```
andrii-hoienko, line 25: needed do not add context, it is not clear whe this 'needed' happens, it will be better to call it grow when full or something

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+
+    private void growIfNeeded(List<T> list) {
+        if (list != null && elements.length < size + list.size()) {
+            Object[] newArray = new Object[size + list.size()];
+            System.arraycopy(elements, 0, newArray, 0, size);
+            elements = newArray;
+        }
+    }
+
+    private void addElementArrayListIndexOutOfBoundException(int index) {
```
andrii-hoienko, line 119: bad name for a method, usually Exceptions end with the word Exception, not methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index: " + index + " out of array size: " + size);
+        }
+    }
+
+    private void arrayListIndexOutOfBoundException(int index) {
```
andrii-hoienko, line 126: code duplication, think how to unite arrayListIndexOutOfBoundException and addElementArrayListIndexOutOfBoundException

src/main/java/core/basesyntax/ArrayList.java
andrii-hoienko: That is good point you can just call method add in loop body for adding elements of list provided

keda-vlad (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index: " + index + " out of array size: " + size);
+        }
+    }
+
+    private void removeElement(int index) {
+        if (index == elements.length - 1) {
+            elements[size - 1] = null;
```
andrii-hoienko, line 135: Imagine we want to remove index 9, and array list is full (lenth 10, size 9). When you write [size - 1] element with index 8 will be removed instead of 9

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+    }
+
+    private void arrayListIndexOutOfBoundException(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index: " + index + " out of array size: " + size);
+        }
+    }
+
+    private void removeElement(int index) {
```
andrii-hoienko, line 133: Better to call remove element by index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        T removedElement = (T)elements[index];
+        removeElement(index);
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int indexRemovedElement = -1;
+        for (int i = 0; i < size; i++) {
```
andrii-hoienko, line 72: It will be better if you extract this loop to method and call it getElementIndex or something

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+    }
+
+    private void growIfNeeded() {
+        if (elements.length == 1 && elements.length == size) {
+            Object[] newArray = new Object[DEFAULT_CAPACITY];
+            System.arraycopy(elements, 0, newArray, 0, size);
+            elements = newArray;
+        }
+        if (elements.length == size) {
+            Object[] newArray = new Object[(int)((double)elements.length * GROWTH_FACTOR)];
```
andrii-hoienko, line 105: is it really reqired to cast first to double, than to int?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growIfNeeded() {
+        if (elements.length == 1 && elements.length == size) {
+            Object[] newArray = new Object[DEFAULT_CAPACITY];
+            System.arraycopy(elements, 0, newArray, 0, size);
```
andrii-hoienko, line 101: Code duplication

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growIfNeeded() {
+        if (elements.length == 1 && elements.length == size) {
```
andrii-hoienko, line 99: it is possible to insert this if condiotion after  if (elements.length == size), so It will be nested if first one will be if (elements.length == size) and than if (elements.length == 1)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+            elements = newArray;
+        }
+        if (elements.length == size) {
+            Object[] newArray = new Object[(int)((double)elements.length * GROWTH_FACTOR)];
+            System.arraycopy(elements, 0, newArray, 0, size);
+            elements = newArray;
+        }
+    }
+
+    private void growIfNeeded(List<T> list) {
```
andrii-hoienko, line 111: I think it is better to use only one growIfNeeded method for all grow purposes

andrii-hoienko (changes requested):
Let`s improve your solution

src/main/java/core/basesyntax/ArrayList.java
IhorKalaur: I saw that this keyword is used in the original Arrailist, so I thought it would be appropriate to use it here. As far as I understand, the data of this field will not be saved when trying to transfer data over the network. I'm not sure it's necessary

src/main/java/core/basesyntax/ArrayList.java
IhorKalaur: You are right, in this case will be automatic cast to double

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+            throw new NoSuchElementException("Can't find element: " + element);
+        }
+        return index;
+    }
+
+    private void fastRemoveElement(int index) {
+        if (index == elements.length - 1) {
+            elements[size - 1] = null;
+        } else {
+            System.arraycopy(elements,index + 1, elements, index, size - index - 1);
```
nacenik, line 147: How do you think, what happens if you delete 3 time in the row from center position? What will your array contains? Do you need to clear lans position after each copy, or not?

src/main/java/core/basesyntax/ArrayList.java
nacenik: not fixed. you have a duplicated lines

src/main/java/core/basesyntax/ArrayList.java
nacenik: Agree with Andrii. Do sth like 
```
if (first condition) {
  if (second condition) {
    ...
  } else {
    ...
  }
}
```
instead of do if-if containing a part of the same conditions

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+        growIfFuel();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (T element: list) {
```
nacenik, line 44:
```suggestion
        for (T element : list) {
```

nacenik (changes requested):
Good solution, let's improve it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,151 @@
...
+            public T next() {
+                if (!hasNext()) {
+                    throw new NoSuchElementException();
+                }
+                return elements[index++];
+            }
+        };
+    }
+
+    private void growIfFuel() {
```
nacenik, line 106: typo mistake 
```suggestion
    private void growIfFull() {
```

src/main/java/core/basesyntax/ArrayList.java
IhorKalaur: You are right, unused objects will remain in memory. I missed this moment


> How do you think, what happens if you delete 3 time in the row from center position? What will your array contains? Do you need to clear lans position after each copy, or not?

src/main/java/core/basesyntax/ArrayList.java
IhorKalaur: I think it's possible to create an empty array, but maybe it make sense to limit this action

nacenik (approved):
LGTM! 👍

PR #1059, opened 2023-07-11
Implement ArrayList

hlibmurphy:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+            listArray = extendArray();
+        }
 
+        listArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size + 1 || index < 0) {
```
okuzan, line 28: The condition to check the validity of the index should be 'index > size' instead of 'index > size + 1' because adding an element at index equal to the size is a valid operation in ArrayList, it's similar to append at the end.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+
+        newArray[index] = value;
+        listArray = newArray;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            if (list.get(i) != null) {
```
okuzan, line 56: Checking for null isn't necessary, as ArrayList allows null elements. If the list contains null elements, they should be added to the ArrayList.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     public void add(T value, int index) {
+        if (index > size + 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+
+        if (size + 1 >= listArray.length) {
+            listArray = extendArray();
+        }
 
+        size++;
```
okuzan, line 36: The 'size++' should be moved after the newArray is filled, because if an exception occurs while filling the newArray, the size variable will be incorrect.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size + 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+
+        if (size + 1 >= listArray.length) {
```
okuzan, line 32: The condition for extending the array should be 'if (size >= listArray.length)' instead of 'if (size + 1 >= listArray.length)'. It's because the size of the array hasn't been incremented yet.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_SIZE = 10;
+    private int size;
+    private T[] listArray;
+
+    public ArrayList() {
+        size = 0;
```
okuzan, line 12: redundant, since 0 is default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+                isElementInArray = true;
+                break;
+            }
+        }
+
+        if (!isElementInArray) {
+            throw new NoSuchElementException("No such element");
+        }
+
+        return removedElement;
```
okuzan, line 106: think about how this can be simplified

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        size--;
+
+        return removedItem;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T removedElement = null;
+        boolean isElementInArray = false;
```
okuzan, line 93: can we do without this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private T[] extendArray() {
+        return Arrays.copyOf(listArray, listArray.length + listArray.length / 2);
```
okuzan, line 126: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        T[] newArray = (T[]) new Object[listArray.length];
+        for (int i = 0; i <= size; i++) {
+            if (i < index) {
+                newArray[i] = listArray[i];
+            }
+
+            if (i > index) {
+                newArray[i] = listArray[i - 1];
+            }
+        }
```
okuzan, line 47: image
one system arraycopy call is enough for each method, no temp array is needed anywhere except for resizing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+    }
+
     @Override
     public void add(T value) {
+        if (size >= listArray.length) {
+            listArray = extendArray();
+        }
 
+        listArray[size] = value;
+        size++;
```
okuzan, line 23:
```suggestion
        listArray[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_ARRAY_SIZE = 10;
+    private int size;
+    private T[] listArray;
```
```suggestion
    private T[] values;
```
naming's a bit confusing. Let's call it elements or values

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index > size + 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+
+        if (size + 1 >= listArray.length) {
+            listArray = extendArray();
+        }
```
okuzan, line 34: move the if inside the grow method, rename it accordingly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        checkIndex(index);
+
+        return listArray[index];
     }
 
     @Override
-    public void set(T value, int index) {
+    public void set(T value, int index) throws ArrayListIndexOutOfBoundsException {
```
okuzan, line 70: no need to declare throwing a runtime exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        checkIndex(index);
+
+        T removedItem = listArray[index];
+        for (int i = index + 1; i < size; i++) {
+            listArray[i - 1] = listArray[i];
+        }
```
okuzan, line 83: we can't move element in for loop in this task

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+
+        size--;
+
+        return removedItem;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T removedElement = null;
```
okuzan, line 92: can we do without this variable?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
```
okuzan, line 121: include size and index in the message

okuzan (changes requested):
Here are some ideas on how to make your solution better:

Ivan95kos (approved):
Nice

PR #1060, opened 2023-07-11
Implement ArrayList

NikitaKryvenets:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element || elements[i] != null && elements[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found(");
```
lazarchuk-v, line 63: add element which was not found to error description 
"Element not found: " + element

lazarchuk-v (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double MAGNIFICATION_FACTOR = 1.5;
+    private static int defaultCapacity = 10;
```
nacenik, line 7: it should be constant 
if you need to know current array length, just use `elements.length`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,94 @@
...
+    }
+
+    private void indexCheck(int listIndex) {
+        if (listIndex < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Entered negative index " + listIndex);
+        }
+        if (listIndex >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Entered index " + listIndex
+                    + " exceeds the size");
+        }
```
nacenik, line 92: you can merge these ifs to one if and write a message: "Index [" index + "] out of bound" or sth like this

nacenik (changes requested):
Good solution, let's improve it :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final double MAGNIFICATION_FACTOR = 1.5;
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
nacenik, line 13: int is 0 by default 
```suggestion
```

nacenik (approved):
Approved!

PR #1064, opened 2023-07-12
Completed task, implemented all methods List to ArrayList

keda-vlad:


Vladiclevch (approved):
my solution is different than yours but it seems to work fine for me

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
```
nacenik, line 8: Why do you need to have this? Your array should never be empty. How will you change a capacity if array will be empty?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity >= 0) {
+            elementData = new Object[initialCapacity];
+        } else {
+            throw new IllegalArgumentException("Illegal Capacity: " + initialCapacity);
+        }
+    }
+
+    public ArrayList() {
+        elementData = EMPTY_ELEMENT_DATA;
```
nacenik, line 21:
```suggestion
        elementData = new Object()[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList(int initialCapacity) {
+        if (initialCapacity >= 0) {
+            elementData = new Object[initialCapacity];
```
nacenik, line 14: How will you increase array size if it will be 0 or 1? Or how much time do you need to create new array if you start size will be 2? 
```suggestion
        if (initialCapacity >= DEFAULT_CAPACITY) {
            elementData = new Object[initialCapacity];
        } else if (initialCapacity > 0) {
           elementData = new Object()[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+                size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list == null) {
+            return;
```
nacenik, line 45: in this case better to throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        }
+        while (list.size() > elementData.length - size) {
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            elementData[size] = list.get(i);
+            size++;
+        }
```
nacenik, line 53: what if your capacity will be 10 and you need to add all element from list 30 capacity? 
I'd recommend you just calling method `add(T element)` inside loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+    private void rangeCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0 || elementData != EMPTY_ELEMENT_DATA) {
+            int newCapacity = (int) (oldCapacity * 1.5);
```
nacenik, line 113: all magic numbers should be constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+
+    private void rangeCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void grow(int minCapacity) {
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0 || elementData != EMPTY_ELEMENT_DATA) {
```
nacenik, line 112: do you really need this if-else? I think no

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+            int newCapacity = (int) (oldCapacity * 1.5);
+            elementData = Arrays.copyOf(elementData, newCapacity);
```
nacenik, line 114: use `System.arraycopy()`

nacenik (changes requested):
Let's improve your solution.

src/main/java/core/basesyntax/ArrayList.java
keda-vlad: in the ArrayList class, when an instance is created through an empty constructor, an empty array is created and when the first element is added, the array increases to the default value due to the call to the 'grow' method, so I did the same.

src/main/java/core/basesyntax/ArrayList.java
keda-vlad: if my capacity default 10, and i need add list with size 30 (what capacity isn`t important if i understand right mechanics) and size already 5 for example. I check if list can fit in in free space (30 > 10 - 5) if not grow until can

Vasul-Pankiv (approved):

PR #1067, opened 2023-07-14
Added throw NoSuchElementException, and addded ArrayList class which …

hanshyn:
Added throw NoSuchElementException, and added ArrayList class which implements the List, override the methods

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] dataArray;
+
+    public ArrayList() {
+        dataArray = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == dataArray.length) {
+            extensionArray();
+        }
+        dataArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexInclusive(index);
+        if (size == dataArray.length) {
+            extensionArray();
+        }
 
+        if (index != size) {
+            System.arraycopy(dataArray, index, dataArray, index + 1, size - index);
+        }
+
+        dataArray[index] = value;
+        size++;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public void add(T value, int index) {
+        checkIndexInclusive(index);
+        if (size == dataArray.length) {
+            extensionArray();
+        }
+
+        if (index != size) {
+            System.arraycopy(dataArray, index, dataArray, index + 1, size - index);
+        }
```
olekskov, line 32: may be simplified and method checkIndexInclusive() is not really required at all:
```suggestion
        if (size == dataArray.length) {
            add(value);
            return;
        }
        checkIndex(index);
        System.arraycopy(dataArray, index, dataArray, index + 1, size - index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " Size" + size);
+        }
+    }
+
+    private void checkIndexExclusive(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " Size" + size);
+        }
     }
```
olekskov, line 110: After changes above:
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index + " Size" + size);
        }
    }
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (size == dataArray.length) {
+            increaseArrayCapacity();
+        }
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " Size" + size);
+        }
```
olekskov, line 30: Please reuse methods you already have implemented
```suggestion
        if (size == dataArray.length) {
            add(value);
            return;
        }
        checkIndex(index);
```

olekskov (approved):
Good job, approved

PR #1077, opened 2023-08-01
Implemented ArrayList class

korets270:
Implemented ArrayList class

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,141 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object [] DEFAULT_EMPTY_ELEMENTDATA = {};
+    private int size;
+    private Object[] elementData;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int initialSize) {
+        if (initialSize > 0) {
+            elementData = new Object[initialSize];
+        } else if (initialSize == 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        } else {
+            elementData = DEFAULT_EMPTY_ELEMENTDATA;
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        resizeArrayIfNeeded();
+        elementData[size] = value;
```
overpathz, line 40: Minor thing, but you can(not necessarily) simplify this line to `elementData[size++] = value;` and don't write size++ on the next line.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+            checkIndex(index, size);
+        }
+        resizeArrayIfNeeded();
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
```
overpathz, line 56: What about just (in add method check performed to grow our array)?
```
    add(list.get(i));
}
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
-        return null;
+        checkIndex(index, size);
+        T deletedElement = (T) elementData[index];
+        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+        size--;
+        return deletedElement;
     }
 
     @Override
     public T remove(T element) {
```
overpathz, line 90: Good. But for better readability you can something like that, if you want:
```
int index = indexOf(element); (add indexOf method)
      if (index != -1) {
          return remove(index);
      }
```

overpathz (approved):
Nice

PR #1078, opened 2023-08-02
implemented tasks in Readme file

Vitalii85:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        this.checkIndex(index);
+        final T element = arrayList[index];
+        for (int i = index; i < size - 1; i++) {
+            arrayList[i] = arrayList[i + 1];
+        }
```
aaananas, line 66: lets use System.arrayCopy

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    private void grow() {
+        capacity += capacity >> 1;
+        T[] newArrayList = (T[]) new Object[capacity];
+        System.arraycopy(arrayList, 0, newArrayList, 0, size);
+        arrayList = newArrayList;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index passed is invalid");
```
okuzan, line 98: include index and size in the message to make debugging easier

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (arrayList[i] == null && element == null
+                    || arrayList[i] != null && arrayList[i].equals(element)) {
+                return this.remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
okuzan, line 76: include element in the exception message to make problem immediately obvious

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        this.checkIndex(index);
+        final T element = arrayList[index];
+        System.arraycopy(arrayList, index + 1, arrayList, index, size - index - 1);
+        size--;
+        arrayList[size] = null;
```
```suggestion
        arrayList[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+            this.add(arrayList[size - 1]);
+            System.arraycopy(arrayList, index, arrayList, index + 1, size - index - 2);
+            arrayList[index] = value;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list != null) {
```
okuzan, line 39: we don't need this check, original java.util impl fails with NPE too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public void add(T value, int index) {
-
+        if (index == size) {
+            this.add(value);
+        } else {
+            this.checkIndex(index);
+            this.add(arrayList[size - 1]);
+            System.arraycopy(arrayList, index, arrayList, index + 1, size - index - 2);
+            arrayList[index] = value;
+        }
```
okuzan, line 34: let's return early here and unwrap the major chunk of code
```suggestion
        if (index == size) {
            this.add(value);
            return;
        } 
            this.checkIndex(index);
            this.add(arrayList[size - 1]);
            System.arraycopy(arrayList, index, arrayList, index + 1, size - index - 2);
            arrayList[index] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == capacity) {
+            this.grow();
+        }
+        arrayList[size] = value;
+        size++;
```
```suggestion
        arrayList[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
+    private int size;
+    private int capacity;
```
okuzan, line 9: I'd recommend to get rid of this field, it takes more hustle to keep it updated that it benefits us

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
```
okuzan, line 7: let's be laconic
```suggestion
    private T[] list;
```

okuzan (changes requested):
Remove redundant `this.` calls everywhere

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
+        add(list[size - 1]);
```
aaananas, line 29:
```suggestion
```
why are you adding that element separately, you can move with System.Arraycopy on 30s line

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Vitalii85: if array is full there must be an error
description in System.Arraycopy:
"Otherwise, if any of the following is true, an IndexOutOfBoundsException is thrown and the destination is not modified:
...
**_destPos+length is greater than dest.length, the length of the destination array._**"

Vitalii85: so I adding element first

Vitalii85: image

Vitalii85: i can add "value" and rewrite it with System.Arraycopy
```suggestion
        add(value);
        System.arraycopy(list, index, list, index + 1, size - index - 1);
        list[index] = value;
```
is this what you want?

Vitalii85: image

Rommelua: you forget to grow() if needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+        System.arraycopy(list, index + 1, list, index, size - index - 1);
+        list[--size] = null;
+        return element;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (list[i] == null && element == null
```
Rommelua, line 65:
```suggestion
            if (list[i] == element 
```

Rommelua (changes requested):
You should fix (or **discuss in Slack** if disagree) all mentor's comments before re-request new review. Be attentive and do not hesitate to contact us in Slack if you do not understand something.
Do not discuss comments here on github - it takes more time

sokimaaa (approved):

PR #1086, opened 2023-08-09
Implemented add, add with index, addAll, get, set, remove by index, r…

dizizzz:
…emove by element methods, size and isEmpty.

 Added method when the size reaches the capacity and indexOf to find the index of a specific element in an array.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,114 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double GROW_ARRAY_SIZE = 1.5;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        resize();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
+        return -1;
+    }
```
aaananas, line 113: combine those ifs on 106 and 100 lines, you don't need that outer if-else

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
```
aaananas, line 87:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double GROW_ARRAY_SIZE = 1.5;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
aaananas, line 13:
```suggestion
```
class field have default values - int default value is 0

aaananas (changes requested):

aaananas (approved):

PR #1088, opened 2023-08-10
Initial commit

gordgelin:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        growIfArrayFull();
+
+        if (index < 0 || index > currentTailIndex) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add by index " + index
+                + "while current index is " + currentTailIndex);
+        }
```
olekskov, line 26: 1. It's better to grow array after index check
2. It's better to reuse existing methods
```suggestion
        if (index == currentTailIndex) {
            add(value);
            return;
        }
        checkIfIndexExists(index);
        growIfArrayFull();
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        System.arraycopy(storage, index, storage, index + 1, currentTailIndex - index);
+        currentTailIndex++;
+        set(value, index);
     }
 
     @Override
     public void addAll(List<T> list) {
+        while (list.size() + currentTailIndex > storage.length) {
+            grow();
+        }
```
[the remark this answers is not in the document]
gordgelin: i just decided to grow array to proper size at once before adding elements but ok

src/main/java/core/basesyntax/ArrayList.java
gordgelin: agree, i just have stuck with set while could just use add... )))

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] storage = new Object[INITIAL_CAPACITY];
+    private int currentTailIndex = 0;
```
kozhukhovsky, line 9:
```suggestion
    private int currentTailIndex;
```

0 is default value for int

PR #1092, opened 2023-08-14
implemented ArrayList

RedArtelerist:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elements.length) {
```
ystankevych, line 17: Maybe you could check it with some method to get rid of duplicate code. But I think it's up to you. (line 17 and line 26)

ystankevych (approved):
Great solution!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        int oldCapacity = elements.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        Object[] copiedElements = new Object[newCapacity];
+        System.arraycopy(elements, 0, copiedElements, 0, size);
+        elements = copiedElements;
+    }
+
+    private int getIndexByValue(T value) {
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == value || (elements[i] != null && elements[i].equals(value))) {
```
liliia-ponomarenko, line 105:
```suggestion
            if (elements[i] == value || elements[i] != null && elements[i].equals(value)) {
```
These brackets are redundant

liliia-ponomarenko (approved):
Great job!

PR #1093, opened 2023-08-14
implemented ArrayList class

ystankevych:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+        return size() <= 0;
+    }
+
+    private boolean isNeedToResize() {
+        return size() >= elements.length;
+    }
+
+    private void grow() {
+        int newLength = size() + (size() >> 1);
+        elements = Arrays.copyOf(elements, newLength);
```
RedArtelerist, line 86: If i understand we only can use System.arraycopy, imagine we don't have Arrays class
```suggestion
        elements = System.arraycopy(...);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int position;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        position = -1;
```
RedArtelerist, line 13: It will be better if we initialize it to 0, so it will be clearer for other developers.
```suggestion
        position = 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexToGet(index);
+        T toRemove = (T) elements[index];
+        if (!isLastIndex(index)) {
+            System.arraycopy(elements, index + 1, elements, index, position - index);
+        }
+        elements[position--] = null;
```
RedArtelerist, line 61: I don't sure that we need assign it to null, we can only decrease size, so the next element addition will overwrite this value

RedArtelerist (changes requested):
Great solution, let's improve it a little

src/main/java/core/basesyntax/ArrayList.java
ystankevych: ok, I changed to System.arraycopy , but there were no restrictions on the use Arrays.copyOf ))

src/main/java/core/basesyntax/ArrayList.java
ystankevych: I think we could or could not - It really depends on your logic and approach.

src/main/java/core/basesyntax/ArrayList.java
RedArtelerist: I understand, but it's a little weird if we implement ArrayList and use helper Arrays class. Perhaps I was confused by this paragraph in the common mistakes "Use System.arraycopy() to move your array elements"

RedArtelerist (approved):
Good job

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public int size() {
-        return 0;
+        return position;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() <= 0;
```
liliia-ponomarenko, line 76:
```suggestion
        return size == 0;
```

liliia-ponomarenko (approved):
LGTM 🔥

PR #1099, opened 2023-08-18
implemented ArrayList class

mmosii:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        if (size == elements.length) {
+            grow(elements.length);
+        }
+    }
+
+    private int elementCheck(T element) {
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element || elements[i] != null && elements[i].equals(element)) {
+                index = i;
```
RedArtelerist, line 106: You can immediately return index if condition is true, this will reduce the number of loop iterations
```suggestion
                return index;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            }
+        }
+        return index;
```
RedArtelerist, line 109:
```suggestion
        return -1;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow(int capacity) {
```
RedArtelerist, line 79: `grow` it's not the best name for this method. Maybe `increaseCapacity` or something like this will be more informative for other developers. And for better readability of your code, it would be better if this method went exactly after `sizeCheck` method

RedArtelerist (changes requested):
Great solution, but try to improve a little it.

RedArtelerist (approved):
Everything is good, nice job

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    private void sizeCheck() {
+        if (size == elements.length) {
+            increaseCapacity(elements.length);
+        }
+    }
+
+    private void increaseCapacity(int capacity) {
+        int newCapacity = capacity + capacity / 2;
+        Object [] destinationArray = new Object[newCapacity];
```
Andriy-dz, line 86:
```suggestion
        Object[] destinationArray = new Object[newCapacity];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size == 0;
+    }
+
+    private void sizeCheck() {
+        if (size == elements.length) {
+            increaseCapacity(elements.length);
+        }
+    }
+
+    private void increaseCapacity(int capacity) {
```
Andriy-dz, line 84:
```suggestion
    private void grow(int capacity) {
```

Andriy-dz (approved):
Good Job!

PR #1105, opened 2023-08-21
All methods are overrided

Alexsandr-Yablunovskyi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        }
+    }
 
+    private Object[] addResize(T value, int index) {
+        Object[] update = new Object[elements.length];
+        System.arraycopy(elements, 0, update, 0, index);
+        update[index] = value;
+        System.arraycopy(elements, index, update, index + 1,
+                size - index);
+        return update;
```
RedArtelerist, line 36: Why we need new array for this operation. We can copy in the same array, and don't need return update
```suggestion
    private void addResize(T value, int index) {
        System.arraycopy(elements, index, update, index + 1,
                size - index);
        elements[index] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
 
     }
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(T element) {
+        for (int i = 0; i < size; i++) {
+            if (element == null && elements[i] == null
+                    || element != null && element.equals(elements[i])) {
```
RedArtelerist, line 63: We can simplify this
```suggestion
            if (element == elements[I] || element != null && element.equals(elements[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        elements = removeResize(index);
+        size--;
+        return removeValue;
+    }
+
+    private Object[] removeResize(int index) {
+        Object[] update = new Object[elements.length];
+        System.arraycopy(elements, 0, update, 0, index);
+        System.arraycopy(elements, index + 1, update, index, size - index - 1);
+        return update;
```
RedArtelerist, line 83: The same problem that I describe above in add method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bounds for length " + size);
+        }
+        return true;
+    }
+
+    private boolean checkAddIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bounds for length " + size);
```
RedArtelerist, line 116: In this case you can add more informative message different from above

RedArtelerist (changes requested):
Good work, try to improve it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -22,18 +22,15 @@ public void add(T value) {
...
-    private Object[] addResize(T value, int index) {
-        Object[] update = new Object[elements.length];
-        System.arraycopy(elements, 0, update, 0, index);
-        update[index] = value;
-        System.arraycopy(elements, index, update, index + 1,
-                size - index);
-        return update;
+    private Object[] addResize(int index) {
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        return elements;
```
RedArtelerist, line 33: We can make this method void, it's not necessary to return elements, because we already update it.
```suggestion
    private void addResize(int index) {
        System.arraycopy(elements, index, elements, index + 1, size - index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -22,18 +22,15 @@ public void add(T value) {
     public void add(T value, int index) {
         if (checkAddIndex(index)) {
             growIfFull();
-            elements = addResize(value, index);
+            elements = addResize(index);
```
RedArtelerist, line 25:
```suggestion
            addResize(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -59,28 +56,25 @@ public void set(T value, int index) {
...
         }
-        throw new NoSuchElementException("There is no such an element");
+        throw new NoSuchElementException("There is no such an element " + element);
     }
 
     @Override
     public T remove(int index) {
         checkIndex(index);
         T removeValue = (T) elements[index];
         elements = removeResize(index);
```
RedArtelerist, line 70:
```suggestion
        removeResize(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -59,28 +56,25 @@ public void set(T value, int index) {
...
-        size--;
+        elements[--size] = null;
         return removeValue;
     }
 
     private Object[] removeResize(int index) {
```
RedArtelerist, line 75:
```suggestion
    private void removeResize(int index) {
```

RedArtelerist (changes requested):
Now almost all are great. Fix one moment and then everything will be ideal.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+
```
liliia-ponomarenko, line 8: Redundant  empty line

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: agree

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: agree. Also, move this method after the public and give a more appropriate name or use this line in the method 
directly.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -59,28 +56,25 @@ public void set(T value, int index) {
...
         return removeValue;
     }
 
     private Object[] removeResize(int index) {
-        Object[] update = new Object[elements.length];
-        System.arraycopy(elements, 0, update, 0, index);
-        System.arraycopy(elements, index + 1, update, index, size - index - 1);
-        return update;
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        return elements;
```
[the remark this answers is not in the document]
liliia-ponomarenko: agree. Also, move this method after the public and give a more appropriate name or use this line in the method
directly.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bounds for length " + size);
+        }
+        return true;
+    }
+
+    private boolean isFull() {
+        return size >= elements.length;
     }
```
liliia-ponomarenko, line 117:
```suggestion
    private boolean isNotFull() {
        return size < elements.length;
    }
```
you can rewrite conditional instead have a lot of code in if block

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+            elements = newElements;
+        }
+    }
+
+    private boolean checkIndex(int index) {
+        if (index < 0 || index > size - 1) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bounds for length " + size);
+        }
+        return true;
```
liliia-ponomarenko, line 104:
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index > size - 1) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index
                    + " out of bounds for length " + size);
        }
```
you can make this method void instead have a lot of code in if block

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,118 @@
...
+    }
+
+    private boolean checkAddIndex(int index) {
```
liliia-ponomarenko, line 107: the same here

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it ;)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= 0;
```
Elena-Bruyako, line 74: size couldn't be less then 0, so just size == 0;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+        }
+        throw new NoSuchElementException("There is no such an element " + element);
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(int index) {
+        checkIndex(index);
+        T removeValue = (T) elements[index];
```
Elena-Bruyako, line 61: Just value or removedValue, you can't use verbs for variables

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) elements[index];
```
Elena-Bruyako, line 39: You don't need to casting to generics

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,99 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] elements;
```
Elena-Bruyako, line 8: There is should be parametrised type

Elena-Bruyako (changes requested):
Need to fixed some comments

src/main/java/core/basesyntax/ArrayList.java
Alexsandr-Yablunovskyi: If element's type is parameterized, then it needs to cast 'elements' array inside a constructor anyway. Is that a major mistake to use 'Object' type for an array?

Elena-Bruyako: Not major and this isn't a mistake. It's like another option of implementation.

Elena-Bruyako (approved):

PR #1108, opened 2023-08-22
HW array list

OlehReex:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Cayman4010, line 12: This value of size is redundant, because size = 0 by default. But it`s no critical.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size() == elementData.length) {
```
Cayman4010, line 17: Why did you use size(), but not just size?

Cayman4010 (changes requested):
Good solution! Check, please, my comments and make one more commit for the mentor reviewing.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] == element || elementData[i] != null
+                    && elementData[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("There aren't your element at list");
```
Elena-Bruyako, line 72: I would suggest to use more abstract message: 
"Element does not exist" or "No such element"

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Object[] grow() {
```
Elena-Bruyako, line 85: This method could be void
```suggestion
    private void grow() {
    Object[] newArray = new Object[(int) (elementData.length * 1.5)];
        System.arraycopy(elementData, 0, newArray, 0, size);
        elementData = newArray;
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
```
Elena-Bruyako, line 7: Let's rename to elements or values

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+
+    private Object[] grow() {
+        Object[] newArray = new Object[(int) (elementData.length * 1.5)];
+        System.arraycopy(elementData, 0, newArray, 0, size);
+        return newArray;
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index > size - 1) {
+            throw new ArrayListIndexOutOfBoundsException("There is no element at index " + index);
```
Elena-Bruyako, line 93: If you are checking the index it would be better to rename the message to 'Index does not exist' or 'Index out of bounds' because you aren't checking the element here. Not critical

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        indexValidation(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexValidation(index);
+        T element = (T) elementData[index];
+        System.arraycopy(elementData, index + 1, elementData,
```
Elena-Bruyako, line 58: You can inline this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        addIndexValidation(index);
+        if (size == elementData.length) {
+            elementData = grow();
+        }
+        System.arraycopy(elementData, index, elementData,
```
Elena-Bruyako, line 29: You can inline this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
```
Elena-Bruyako, line 16: This if statements could be replace to grow() method to avoid duplication of code
See 26 line also

Elena-Bruyako (changes requested):
Good work! Let's make your code better!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        if (size == elements.length) {
+            Object[] newArray = new Object[(int) (elements.length * 1.5)];
```
Andriy-dz, line 80: 1.5 must be constant

PR #1109, opened 2023-08-22
implemented all the methods for ArrayList

Cayman4010:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("You can`t work with index " + index);
+        }
+    }
+
+    private void correctSize() {
+        if (elements.length == size) {
+            Object[] internalArray = new Object[(int) (size * LIST_ENLARGER)];
```
don-bigdad, line 97:
```suggestion
            Object[] internalArray = new Object[(int) (size * LIST_ENLARGER)];
            You can use binary shift to the right, which increases the number on 1.5
            Object[] internalArray = new Object[size + (size >> 1)]
```

don-bigdad (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    @SuppressWarnings("unckecked")
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element || element != null && element.equals(elements[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("This element doesn`t exist in in the list");
```
Elena-Bruyako, line 70: double 'in' word

Elena-Bruyako (changes requested):
Good work!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    @SuppressWarnings("unckecked")
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element || element != null && element.equals(elements[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("This element doesn`t exist in the list");
```
Andriy-dz, line 69:
```suggestion
        throw new NoSuchElementException("This element " + element + " doesn`t exist in the list");
```
it is better to indicate which element is missing from your list - it is more informative for the user)

Andriy-dz (approved):
Perfect!

PR #1110, opened 2023-08-22
realize my own arrayList with some methods

don-bigdad:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size == 0;
+    }
+
+    private void checkBound(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound: " + index);
+        }
+    }
+
+    private void newCapacity() {
```
anigilator479, line 95: make the method name more clear what it does

anigilator479 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int INITIALIZE_SIZE = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[INITIALIZE_SIZE];
+        size = 0;
```
Andriy-dz, line 13: it is not necessary because the default value of the note is int = 0

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elements.length) {
```
Andriy-dz, line 18: if you move this check to the middle of the method, you will get rid of the duplicate check

src/main/java/core/basesyntax/ArrayList.java
Andriy-dz: agree.  [Use informative names for your variables and methods](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-informative-names-for-your-variables-and-methods)

Andriy-dz (changes requested):
Good Job! Let's inprove your solution)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bound: " + index);
+        }
+    }
+
+    private void increaseCapacity() {
+        int oldSize = size;
+        int newSize = oldSize + (oldSize >> 1);
+        Object[] newArray = new Object[newSize];
+        System.arraycopy(elements,0,newArray,0,size);
```
liliia-ponomarenko, line 101:
```suggestion
        System.arraycopy(elements, 0, newArray, 0, size);
```
Let’s use whitespaces ;)

liliia-ponomarenko (approved):
Great 🔥

PR #1111, opened 2023-08-23
ArrayList created

validatorr:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private void indexInsertion(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index" + index);
+        }
+    }
+
+    private void indexAdditionAndDeletion(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index" + index);
+        }
```
AlexRosenkreuz, line 100: You can simplify those methods to one and it'll work just fine)
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Invalid index" + index);
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void resizeIfNeeded() {
```
AlexRosenkreuz, line 81: Suggest renaming it to ensureCapacity() to make it more formal and clear.

AlexRosenkreuz (changes requested):
Great job! Just a few suggestions to improve your code :)

src/main/java/core/basesyntax/ArrayList.java
AlexRosenkreuz: I see. Just move size++ in your add method and it should work with simplified method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        indexInsertion(index);
+        ensureCapacity();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
```
AlexRosenkreuz, line 29: Change this and it should work with simplified method I mentioned earlier.
```suggestion
        size++;
        indexInsertion(index);
        ensureCapacity();
        System.arraycopy(elements, index, elements, index + 1, size - index);
        elements[index] = value;
    }
```

AlexRosenkreuz (changes requested):

src/main/java/core/basesyntax/ArrayList.java
validatorr: Yeah, everything works! Thanks for tip, it helped a lot!

AlexRosenkreuz (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
```
liliia-ponomarenko, line 6: why public?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        elements[size] = value;
+        size++;
```
liliia-ponomarenko, line 19:
```suggestion
        elements[size++] = value;
```
You can do it this way

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
```
liliia-ponomarenko, line 24: better increase  size at the end of the method because  if some issue occurs in the next code this value  doesn’t  rollback

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 39: Do you really need this annotation?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (elements.length > size + 1) {
```
liliia-ponomarenko, line 82:
```suggestion
        if (elements.length > size) {
```

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    }
+
+    private void ensureCapacity() {
+        if (elements.length > size) {
+            return;
+        }
+        Object[] newArray = new Object[(int) (elements.length * ARRAY_MULTIPLICATION)];
+        System.arraycopy(elements, 0, newArray, 0, size);
+        elements = newArray;
+
```
liliia-ponomarenko, line 86: Redundant empty line

liliia-ponomarenko (approved):
Great 🔥

PR #1113, opened 2023-08-23
implement List interface in ArrayList class

anigilator479:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_INDEX = 1.5;
```
KDanil2000, line 7: may be better will be GROW_MULTIPLIER instead GROW_INDEX

KDanil2000 (approved):
Good solution!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elements.length) {
+            growIfArrayFull();
+        }
+        elements[size] = value;
+        size++;
```
liliia-ponomarenko, line 21:
```suggestion
        elements[size++] = value;
```
Can do it this way

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == elements.length) {
+            growIfArrayFull();
+        }
```
liliia-ponomarenko, line 28: firstly, you should  check the index, and after resize

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((elements[i] != null && elements[i].equals(element))
+                    || (elements[i] == element)) {
```
liliia-ponomarenko, line 69:
```suggestion
            if (elements[i] != null && elements[i].equals(element)
                    || elements[i] == element) {
```
Redundant  brackets (read about operators' priority)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1114, opened 2023-08-23
impl some methods

SalnikovMarkiian:
have bugs with tests

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,14 +1,48 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    public static final double PERCENTAGE_ARRAY_GROWTH = 1.5D;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList(int initCapacity) {
+        if (initCapacity <= 0) {
+            throw new IllegalArgumentException();
+        }
+        elementData = new Object[initCapacity];
+    }
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
+        resizeIfNeeded();
+        elementData[size] = value;
+        size++;
+    }
 
+    private void resizeIfNeeded() {
+        if (elementData.length == size) {
+            Object[] newArray = new Object[(int) (elementData.length * PERCENTAGE_ARRAY_GROWTH)];
+            System.arraycopy(elementData, 0, newArray, 0, size);
+            elementData = newArray;
+        }
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexExsist(index, size);
+        resizeIfNeeded();
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


DenisGorda (changes requested):
Your solution didn't pass auto tests.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    private static final double PERCENTAGE_ARRAY_GROWTH = 1.5D;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList(int initCapacity) {
+        if (initCapacity <= 0) {
+            throw new IllegalArgumentException();
+        }
+        elementData = new Object[initCapacity];
+    }
```
liliia-ponomarenko, line 17: Let’s remove this constructor. You never use it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        resizeIfNeeded();
+        elementData[size] = value;
+        size++;
```
liliia-ponomarenko, line 27:
```suggestion
        elementData[size++] = value;
```
You can do it this way

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexExsist(index, size + 1);
```
liliia-ponomarenko, line 32: use verbs for method names. `checkIndex `for example

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
 
     @Override
     public void set(T value, int index) {
-
+        indexExsist(index, size);
+        elementData[index] = value;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 60: Let’s googled about this annotation. Do you really need it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    private void indexExsist(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Element is not exsist");
+        }
+    }
+
+    private int getIndexByValue(T value) {
+        for (int index = 0; index < size; index++) {
+            if (Objects.equals(elementData[index], value)) {
```
liliia-ponomarenko, line 97: [Don’t use class Objects.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#dont-use-class-objects)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexExsist(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Element is not exsist");
```
liliia-ponomarenko, line 91: Let’s  pass a more informative message (pass index for example)

liliia-ponomarenko (changes requested):
Good job! Let’s fix a few mistakes ;)

src/main/java/core/basesyntax/ArrayList.java
SalnikovMarkiian: actually, I only need this annotation to eliminate the warnings specified by the compiler. I'll take the annotations, however, does it really need to be taken away?

liliia-ponomarenko: discussed in slack

liliia-ponomarenko (approved):
LGTM 🔥

PR #1115, opened 2023-08-23
Implemented ArrayList

AlexRosenkreuz:
Implemented methods for custom ArrayList class.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,92 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        checkIndex(index);
+        ensureCapacity();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity();
+        for (int i = 0; i < list.size(); i++) {
+            elements[size++] = list.get(i);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) elements[index];
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (size == elements.length) {
+            elements = Arrays.copyOf(elements, size * 3 / 2 + 1);
```
vsych321, line 83: it would be better for to use System.arraycopy() instead of Arrays.copy, cz i guess, mentors will deny it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        size++;
+        checkIndex(index);
+        ensureCapacity();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
     }
```
vsych321, line 31: in this method everything is good, but there is no logic sequency(as for me)

vsych321 (approved):
in overall everything not bad, but as I said before, mentors might deny it by some problems with chekstyle, I'll approve, cz mb I'm wrong.Good Luck!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
liliia-ponomarenko, line 14: Default  value for int - 0, so you can remove it

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,92 @@
...
     public void add(T value) {
-
+        ensureCapacity();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        size++;
```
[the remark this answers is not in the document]
AlexRosenkreuz: If apply this change, the checkIndex() method will not work properly. I tried to resolve this but it led to code duplicate. Could you suggest how I can make checkIndex() universal for remove() and add() ?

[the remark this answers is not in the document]
Elena-Bruyako: You can use two separate method for checking index with different checking one for remove and one for add. Or leave checking index in method add() as it is.

Elena-Bruyako (approved):

PR #1116, opened 2023-08-23
ArrayList created

KDanil2000:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[--size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null ? elements[i] == null : element.equals(elements[i])) {
```
Parashchuk, line 75: Minor improvement, ternary operator better to use with assignation operations. In this case it's quite difficult to read

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Parashchuk, line 46: Create a helper function to avoid code duplicating

Parashchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
```
Parashchuk, line 51: You still need to avoid the code duplicating

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Parashchuk, line 59: For this and two above statements with duplicate code, you should create a function helper

Parashchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
```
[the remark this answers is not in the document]
Parashchuk: And you need to add the if() check inside resize() function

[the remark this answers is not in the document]
Parashchuk: P.S instead of "SOME_MAGIC_NUMBER" use name you think fit here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+            int newCapacity = elements.length + (elements.length >> SOME_MAGIC_NUMBER);
```
Parashchuk, line 85: instead of "SOME_MAGIC_NUMBER" use name you think fit here.

Parashchuk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
KDanil2000: sorry, forgot about it

Parashchuk:
We are almost there )

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
Parashchuk, line 83: You have no need to pass the argument "minCapacity" into ensureCapacity() function because the variable is available  globally inside your class. Also you don't need to check the (minCapacity > elements.lenght). So combining all of this above make a conclusion to create one method resize() without parameters, and use if() check inside there. Method ensureCapacity() not needed because the newCapacity variable can't have value less than elements' length anyway, that's mean that you no need to handle this exception.

```
        private void resize() {
            if (size ==  elements.length) {
                int newCapacity = elements.length + (elements.length >> INCREMENT_VALUE);
                Object[] newElements = new Object[newCapacity];
                System.arraycopy(elements, 0, newElements, 0, size);
                elements = newElements;
            }
        }
```

Parashchuk (changes requested):

Elena-Bruyako (approved):

PR #1117, opened 2023-08-23
ArrayList created

DenisGorda:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+                    + index);
+        }
+    }
+
+    private void isExist(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Element missing for given index "
+                    + index);
+        }
+    }
```
torbinofficial, line 91: Consider using single method for checking index
```suggestion
    private void isExistOrNext(int index, int bound) {
        if (index > bound || index < 0) { // you can use size or size-1 as bound
            throw new ArrayListIndexOutOfBoundsException("Element missing for given index "
                    + index);
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
+    private void checkArraySize() {
+        if (size + 1 < elements.length) {
+            return;
+        }
+        Object[] newArray = new Object[(int) (elements.length * PERCENTAGE_GROWTH)];
+        System.arraycopy(elements, 0, newArray, 0, elements.length);
+        elements = newArray;
```
torbinofficial, line 99: [Use informative names for your variables and methods](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#use-informative-names-for-your-variables-and-methods)

torbinofficial (changes requested):

src/main/java/core/basesyntax/ArrayList.java
DenisGorda: I believe it's better to keep both methods, as the first one is used when adding a new element and checks indices within the range of length+1, whereas in the second case, we're simply checking if an element with the desired index exists. If we were to use the same method for both cases, conflicts would arise, for instance, when we try to retrieve an element at the index of length+1.

torbinofficial (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    }
+
+    private void isExist(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Element missing for given index "
+                    + index);
+        }
+    }
+
+    private void checkArraySize() {
```
Elena-Bruyako, line 93: Your method not just checking array size it does resize the array, so I would suggest grow or resize

Elena-Bruyako (approved):
Good work! Added some minor comments related to naming

PR #1119, opened 2023-08-23
Hw array list

vsych321:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size == 0;
+    }
+
+    public void resize() {
+        if (elements.length == size) {
+            Object[] newArray = new Object[elements.length * 2];
+            System.arraycopy(elements, 0, newArray, 0, size);
+            elements = newArray;
+        }
+
```
romazan07, line 98: redundant line (#97)

romazan07 (approved):
looks good

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Andriy-dz, line 12: the primitive already has a default value of 0, so this is redundant code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (elements.length == size) {
```
Andriy-dz, line 17: this is a duplicate condition that suggests moving it to the resize method, but it will need to be renamed accordingly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            resize();
+        }
+        elements[size] = value;
+        size++;
```
Andriy-dz, line 21:
```suggestion
        elements[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add value by index");
+        }
+        if (elements.length == size) {
+            resize();
+        }
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
```
Andriy-dz, line 34:
```suggestion
        elements[index++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
+    @SuppressWarnings("unchecked")
```
Andriy-dz, line 44: let's remove these annotations

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Andriy-dz, line 47: it is better to create a separate method for this condition, because the same code already exists

src/main/java/core/basesyntax/ArrayList.java
Andriy-dz: agree with @romazan07

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void resize() {
```
Andriy-dz, line 92: why public?)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = getElementIndex(element);
+        if (index < 0) {
+            throw new NoSuchElementException("Can't remove element");
+        }
```
Andriy-dz, line 78: redundant check, even if the variable index will be -1, you can still pass it to the remove() method and everything will be fine

Andriy-dz (changes requested):
Good work! Lets improve your solution)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't add value by index");
+        }
+        if (elements.length == size) {
```
Elena-Bruyako, line 25: You have already checking the size in the resize() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+
+    private void resize() {
+        if (elements.length == size) {
+            Object[] newArray = new Object[elements.length * 2];
+            System.arraycopy(elements, 0, newArray, 0, size);
+            elements = newArray;
+        }
+    }
+
+    public void checkIndex(int index) {
```
Elena-Bruyako, line 91: this method should be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get index");
+        }
+    }
+
+    public int getElementIndex(T element) {
```
Elena-Bruyako, line 97: this method should be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        int index = -1;
```
Elena-Bruyako, line 98: why -1?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        T removedElement = get(index);
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = getElementIndex(element);
```
Elena-Bruyako, line 64: you can simplify your code 
```suggestion
        for (int i = 0; i < size; i++) {
            if (elements[i] == element || elements[i] != null && elements[i].equals(element)) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element does not exist" + element);
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: This comment is not fixed

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: This comment is not fixed

liliia-ponomarenko (changes requested):
Be attentive! Several comments weren't  fixed

liliia-ponomarenko (approved):
Great 🔥

PR #1120, opened 2023-08-23
Implement ArrayList

Parashchuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentIndex == 0;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void shouldGrow() {
+        if (currentIndex == size) {
+            int newSize = (int) (storage.length + storage.length * GROW_COEFFICIENT);
```
TetianaDenchyk, line 95: The array should grow 1.5 times when it's full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double GROW_COEFFICIENT = 0.75;
```
TetianaDenchyk, line 6: The array should grow 1.5 time when full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        shouldGrow();
+        storage[currentIndex++] = value;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > currentIndex || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    String.format("Index: %d, Size: %d", index, currentIndex)
+            );
```
TetianaDenchyk, line 28: Shouldn't it be on the previous line?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
-
+        validateIndex(index);
+        storage[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        T removedElement = storage[index];
```
TetianaDenchyk, line 59: You should cast the value to T

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < currentIndex; i++) {
+            if ((element == null && storage[i] == null)
+                    || (element != null && element.equals(storage[i]))) {
+                index = i;
+            }
+        }
```
TetianaDenchyk, line 73: You may make a separate method for this check

src/main/java/core/basesyntax/ArrayList.java
TetianaDenchyk, line 28: You may create a separate method for this check

TetianaDenchyk (changes requested):
Great job! I left several comments. You may fix if you consider it necessary

src/main/java/core/basesyntax/ArrayList.java
Parashchuk: It's redundant code casting, because we can't pass not T type values in storage. What do you think ?

src/main/java/core/basesyntax/ArrayList.java
Parashchuk: It only uses once in the program, so i don't think so. Why do you think ?

src/main/java/core/basesyntax/ArrayList.java
Parashchuk: The check uses only once in code, so there's no need to create a function. But i can create a function that throw an error because this piece of code repeats twice in program.

src/main/java/core/basesyntax/ArrayList.java
TetianaDenchyk: Yes, you are right! I overlooked that you had an array of T-type values, but not of Objects

TetianaDenchyk (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double GROW_COEFFICIENT = 0.5;
+    private static final byte DEFAULT_SIZE = 10;
```
liliia-ponomarenko, line 7:
```suggestion
    private static final int DEFAULT_SIZE = 10;
```
Let’s  use int (avoid casting)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private int size = DEFAULT_SIZE;
```
liliia-ponomarenko, line 8: redundant variable. You can avoid using this variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private int currentIndex;
+    private T[] storage;
+
+    @SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 12: Let’s  remove this annotation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    public ArrayList() {
+        storage = (T[]) new Object[DEFAULT_SIZE];
+        currentIndex = 0;
```
liliia-ponomarenko, line 15:
```suggestion
```
0 - default value for int

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        shouldGrow();
```
liliia-ponomarenko, line 20:
```suggestion
        resizeArray();
```
Let’s  give a more appropriate  name for this method

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: agree

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        return currentIndex;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentIndex == 0;
+    }
+
+    @SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 91: Let’s  remove the annotation

liliia-ponomarenko (changes requested):
Good job! Let’s  fix a few mistakes  ;)

src/main/java/core/basesyntax/ArrayList.java
Parashchuk: I don't understand, it's a constat, so it will never be neither changed nor casted.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < currentIndex; i++) {
+            if ((element == null && storage[i] == null)
+                    || (element != null && element.equals(storage[i]))) {
+                index = i;
+            }
+        }
```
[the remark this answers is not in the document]
Parashchuk: I added it only for readability

liliia-ponomarenko (approved):
Good job! 👏

PR #1122, opened 2023-08-23
implement task

romazan07:
implement task

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int START_CAPACITY = 10;
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        array = new Object[START_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    public ArrayList() {
+        array = new Object[START_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            extendArray();
```
anastasiia-fi, line 18: maybe it's better to add  "if (size == array.length)" check to extendArray() method, so you don't need to repeat it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        array[--size] = null;
+        return deletedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = getIndexByValue(element);
+        if (index == -1) {
+            throw new NoSuchElementException("There is no such element in the list");
```
anastasiia-fi, line 68: it's better to add element to message, "There is no such element in the list" > "Element" + element + "doesn't exist" (for example)

anastasiia-fi (changes requested):
great work, just some minor remarks to fix )

liliia-ponomarenko (approved):
LGTM 🔥

PR #1126, opened 2023-08-24
solution added

anastasiia-fi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double RESIZE_VALUE = 1.5;
+    private int size;
+    private T[] list;
+
+    @SuppressWarnings("unchecked")
```
Elena-Bruyako, line 11: Remove this annotation

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        T removed = (T) list[index];
+        System.arraycopy(list, index + 1, list, index, size - index - 1);
+        size--;
+        return removed;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        remove(findIndexByElement(element));
```
Elena-Bruyako, line 63: It's better to create variable for findIndexByElement(element) and pass it to remove method for better readability

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is out of bounds");
+        }
+    }
+
+    private void checkIndexWhenAdd(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is out of bounds");
```
Elena-Bruyako, line 85: I would suggest using not the same message like in checkIndex method to clarify when exactly happened the exception - during add method or others.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        list[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removed = (T) list[index];
+        System.arraycopy(list, index + 1, list, index, size - index - 1);
+        size--;
```
Elena-Bruyako, line 57: To make size correct you need to set the last element as null because you shifted elements after removing 
list[--size] = null;

Elena-Bruyako (changes requested):
Good job! Let's fix some minor comments.

liliia-ponomarenko (approved):
Good job! 👍

PR #1128, opened 2023-08-25
first_commit

Lupen07:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+
+    private int indexOf(T element) {
+        for (int i = 0; i < elements.length; i++) {
+            if (element == null) {
+                if (elements[i] == null) {
+                    return i;
+                }
+            } else if (element.equals(elements[i])) {
+                return i;
+            }
```
aaananas, line 112: lets combine those ifs, you are just overcomplicating

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkAddID(index);
+        provideCapacity(size + 1);
+        System.arraycopy(elements, index, elements,
```
aaananas, line 26:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        checkSizeAndGrow();
        System.arraycopy(elements, index, elements,
```

aaananas, line 26:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        checkSizeAndGrow();
        System.arraycopy(elements, index, elements,
```

aaananas (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+
+    private int indexOf(T element) {
+        for (int i = 0; i < elements.length; i++) {
+            if (element == null) {
+                if (elements[i] == null) {
+                    return i;
+                }
+            } else if (element.equals(elements[i])) {
+                return i;
+            }
```
olekskov, line 115: Anastasia's comment not fixed
```suggestion
            if (element == elements[i] || element != null && element.equals(elements[i])) {
                 return i;
            }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+                    + index + "is out of bounds for size: " + size);
+        }
+    }
+
+    private void checkAddID(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index "
+                    + index + "is out of bounds for size: " + size);
+        }
+    }
```
olekskov, line 95: this method is not required
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkAddID(index);
```
olekskov, line 28: Here you may safely use same checkIndex method as everywhere else as case when (index == size) will not get here due to if above
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,119 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkID(int index) {
```
olekskov, line 83: Please change ID to index everywhere. ID is common abbreviation for "identificator" and this causes confusion during code reading
```suggestion
    private void checkIndex(int index) {
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Lupen07: Hi @olekskov, I added a new code, Please check it. I hope that I did a good job now.

olekskov (approved):
Good job

PR #1129, opened 2023-08-25
My implementation of array list

pastcwdt:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private Object[] elements;
+    private int count;
+
+    public ArrayList() {
+        elements = new Object[CAPACITY];
+        count = 0;
```
RomanYuzviak, line 13: ```count``` by default has 0 value, this line is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
-        return false;
+        return size() <= 0;
+    }
+
+    private boolean isFull() {
+        return size() >= elements.length;
+    }
+
+    private void grow() {
+        int newLength = size() + (size() >> 1);
```
RomanYuzviak, line 85: ```newLength``` is based on size, which is amount of existing elements in list, but should be based on internal array length

RomanYuzviak: you can use ```size``` directly instead of a method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    private void checkIndexToGet(int index) {
+        if (index >= size() || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private int getIndexByValue(T value) {
+        for (int i = 0; i < size(); i++) {
+            T current = (T) elements[i];
+            if (Objects.equals(current, value)) {
```
RomanYuzviak, line 106: Using the ```Objects``` class is not allowed in this taks

RomanYuzviak (changes requested):
to fix

pastcwdt:
done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     @Override
     public int size() {
-        return 0;
+        return count;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() <= 0;
```
Elena-Bruyako, line 75: just size == 0;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    }
+
+    private boolean isFull() {
+        return size() >= elements.length;
+    }
+
+    private void grow() {
+        Object[] tempArr = new Object[(int) (elements.length * 1.5)];
```
Elena-Bruyako, line 83: 1.5 - should be constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private Object[] elements;
+    private int count;
```
Elena-Bruyako, line 8: it's better to name size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexToAdd(index);
+        if (isFull()) {
+            grow();
+        }
+        System.arraycopy(elements, index, elements, index + 1, size() - index);
```
Elena-Bruyako, line 28: you can use your variable count (better size) everywhere and not calling size() method. Check all places

Elena-Bruyako (changes requested):
Good job! Let's fix some minor comments

liliia-ponomarenko (approved):
Great 😊

PR #1132, opened 2023-08-28
arrayList solution.01

YuraNR:


liliia-ponomarenko (changes requested):
Let’s  fix JavaCI checks

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        T removedElemen = null;
+        checkIndex(index, size);
+        for (int i = 0; i < size; i++) {
+            if (i == index) {
+                removedElemen = (T) elements[i];
+            }
+        }
```
liliia-ponomarenko, line 58: do you really need it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Out of bounds: ");
+        }
+    }
+
+    private void resizeIfNeeded() {
+        if (size >= elements.length - 2) {
```
liliia-ponomarenko, line 100:
```suggestion
        if (size == elements.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     public T remove(T element) {
-        return null;
+        T removedElement = null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] == element || elements[i] != null && elements[i].equals(element)) {
+                index = i;
+                removedElement = (T) elements[i];
+            }
+        }
```
liliia-ponomarenko, line 73: Let’s  move it to separate method and simplify your solution

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        if (index == -1) {
+            throw new NoSuchElementException("No such element was found");
+        } else {
+            System.arraycopy(elements, index + 1, elements, index, size - 1 - index);
+            size--;
+        }
```
liliia-ponomarenko, line 79: Can you avoid using this if-else construction?

liliia-ponomarenko (changes requested):
good job! Let’s  fix a few mistakes ;)

src/main/java/core/basesyntax/ArrayList.java
YuraNR: I think I can not avoid if else, because if index != -1 in block else i must do 3 operation that dependent of index.
Or I dont know how to avoid if-else.

liliia-ponomarenko (approved):
good job! 👏

PR #1135, opened 2023-09-02
Hw array list

nastasiiaKulinich:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     public void add(T value, int index) {
-
+        elementData = growIfArrayFull();
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index
+                    + ". Your index must be in the range from 0 to " + size + " inclusive.");
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
```
olekskov, line 30: 1. I don't like idea to make resize before we checked index. Probably we will do nothing, why to resize then?
2. It's better to reuse existing methods to simplify code
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index); // as we already checked case when index == size we can reuse existing method
        growIfArrayFull();
        System.arraycopy(elementData, index, elementData, index + 1, size - index);
        elementData[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private T[] growIfArrayFull() {
+        T[] grownArray = elementData;
+        int oldCapacity = elementData.length;
+        if (size == oldCapacity) {
+            int newCapacity = oldCapacity + (oldCapacity >> 1);
+            grownArray = (T[]) new Object[newCapacity];
+            System.arraycopy(elementData, 0, grownArray, 0, size);
+        }
+        return grownArray;
+    }
```
olekskov, line 92: 1. Matter of taste but for me it seems that this method is more understandable when void as we always update class variable that is accessible from method.
2. Code seems more readable when we have validations first and then main code part which not covered by if block
```suggestion
    private void growIfArrayFull() {
        if (size < oldCapacity) {
            return;
        }
        int newCapacity = elementData.length + (elementData.length >> 1);
        grownArray = (T[]) new Object[newCapacity];
        System.arraycopy(elementData, 0, grownArray, 0, size);
        elementData = grownArray;
    }
```

olekskov (changes requested):

aaananas (approved):

PR #1143, opened 2023-09-14
initial commit

tsuki99:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] dataBase;
+    private int size;
+
+    public ArrayList() {
+        this.dataBase = new Object[DEFAULT_CAPACITY];
+    }
+
+    private void grow(int minCapacity) {
```
kozhukhovsky, line 14: first all public methods, then private ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] dataBase;
```
kozhukhovsky, line 7:
```suggestion
    private T[] dataBase;
```

src/main/java/core/basesyntax/ArrayList.java
kozhukhovsky, line 14:
```suggestion
    private void grow() {
```
you are not using minCapacity.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.dataBase = new Object[DEFAULT_CAPACITY];
+    }
+
+    private void grow(int minCapacity) {
+        if (size >= dataBase.length) {
+            int newCapacity = dataBase.length + (dataBase.length >> 1);
```
kozhukhovsky, line 16: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        dataBase[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
+        }
```
kozhukhovsky, line 34:
```suggestion
        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
        }
```

move this check to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        grow(size + 1);
+        for (int i = size; i > index; i--) {
```
kozhukhovsky, line 36: u can use system.arraycopy here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+            dataBase[i] = dataBase[i - 1];
+        }
+        dataBase[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        grow(size + list.size());
```
kozhukhovsky, line 45:
```suggestion
```
grow is already used in the add method, no need to check twice

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
+        }
```
kozhukhovsky, line 55:
```suggestion
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
        }
```

move this check to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return (T) dataBase[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
+        }
```
kozhukhovsky, line 63: same for this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        dataBase[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Entered invalid index, try again !");
+        }
```
kozhukhovsky, line 71: same for this check

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        final T removedValue = get(index);
```
kozhukhovsky, line 72:
```suggestion
        final T removedValue = dataBase[index];
```
There is already an index check, the get method adds another one to us

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == dataBase[i]) || (element != null && element.equals(dataBase[i]))) {
+                final T removedValue = (T) dataBase[i];
+                System.arraycopy(dataBase, i + 1, dataBase, i, size - i - 1);
+                dataBase[size - 1] = null;
+                size--;
+                return removedValue;
```
kozhukhovsky, line 87:
```suggestion
                return remove(i);
```

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/ArrayList.java
tsuki99: it has no sense cuz in that case im using "index > size" and in other - "index >= size" otherwise it doesnt work, But i did it to other cases

olekskov (approved):

PR #1144, opened 2023-09-15
add ArrayList implementation

VoitseshukV:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
+        final T result = (T) values[index];
+        for (int i = index; i < size - 1; i++) {
+            values[i] = values[i + 1];
+        }
+        size--;
+        values[size] = null;
```
```suggestion
        values[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(values[i], element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
okuzan, line 83:
```suggestion
        throw new NoSuchElementException("Element not found: " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkAndExpandArray() {
+        if (values == null) {
+            values = new Object[DEFAULT_CAPACITY];
+        } else if (values.length == size) {
+            values = Arrays.copyOf(values, values.length + (values.length >> 1));
```
okuzan, line 100: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        }
     }
 
     @Override
-    public T get(int index) {
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds");
+        }
```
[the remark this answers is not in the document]
VoitseshukV: One of the conditions in the common code is slightly different. I considered both options, but it hurts code clarity :(

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkAndExpandArray() {
+        if (values == null) {
+            values = (T[]) new Object[DEFAULT_CAPACITY];
```
olekskov, line 87: Null check here is meaningless. There is no reason to have null array at any stage.
Please add no-args constructor that will initialize array during our list instance creation and remove check from here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+            values = newValues;
+        }
+    }
+
+    private void checkBounds(int index, int availableSize)
+            throws ArrayListIndexOutOfBoundsException {
+        if (index >= availableSize || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " out of bounds, available array size: " + availableSize);
+        }
```
olekskov, line 101: 1. ArrayListIndexOutOfBoundsException is a runtime exception. Please remove it from methods headers
2. We don't really need second argument in our method, better to use as simple solutions as possible

```suggestion
    private void checkBounds(int index) {
        if (index >= size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Index " + index
                    + " out of bounds, available array size: " + size);
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        if (index == size) {
+            add(value);
+        } else {
+            checkAndExpandArray();
+            for (int i = size; i > index; i--) {
+                values[i] = values[i - 1];
+            }
+            values[index] = value;
+            size++;
+        }
```
olekskov, line 29: 1. In order previous comment to work here we need to swap logic a bit
2. It's better to use structures with as less tabulations of code as possible
3. Use arraycopy method instead of cycle (we can perform copy using same array as source and destination)
```suggestion
    public void add(T value, int index) {
        if (index == size) {
            add(value);
            return;
        }
        checkBounds(index);
        checkAndExpandArray();
        System.arraycopy(....);
        values[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        checkBounds(index, size);
+        final T result = (T) values[index];
+        for (int i = index; i < size - 1; i++) {
+            values[i] = values[i + 1];
+        }
```
olekskov, line 60: use arraycopy() method here instead

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (list.isEmpty()) {
+            return;
+        }
```
olekskov, line 38: This check can be removed as in case of empty list cycle will be automatically skipped
```suggestion
```

olekskov (approved):
Good job

PR #1147, opened 2023-09-18
Roma Menzheres's realisation of "ArrayList" task

RomanMenzheres:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,146 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            growIfNeed();
+            elements[size] = list.get(i);
+            size++;
+        }
```
olekskov, line 63: reuse existing methods
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,146 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index " + index);
+        }
```
olekskov, line 70: extract repeating logic to separate private method and reuse in other methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,146 @@
...
+            }
+        }
+
+        return NOT_FOUND;
+    }
+
+    private void growIfNeed() {
+        int valuesLength = elements.length;
+        if (size == valuesLength) {
+            T[] extendedValues = (T[]) new Object[valuesLength + (int) (valuesLength * 1.5)];
```
olekskov, line 141: extract magic value to constant
```suggestion
            T[] extendedValues = (T[]) new Object[valuesLength + (int) (valuesLength * GROW_FACTOR)];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,146 @@
...
+        if (index == size - 1) {
+            elements[index] = null;
+        } else if (index == 0) {
+            T[] newElements = (T[]) new Object[elements.length];
+            System.arraycopy(elements, 1, newElements, 0, size);
+            elements = newElements;
+        } else {
+            T[] newElements = (T[]) new Object[elements.length];
+            System.arraycopy(elements, 0, newElements, 0, index);
+            System.arraycopy(elements, index + 1, newElements, index, size - index);
```
olekskov, line 100: arraycopy is very heavy method, use it only once
you may perform copy having same array as source and destination

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        T oldValue = elements[index];
+        if (index == size - 1) {
+            elements[index] = null;
+        } else if (index == 0) {
+            System.arraycopy(elements, 1, elements, 0, size);
+        } else {
+            System.arraycopy(elements, index + 1, elements, index, size - index);
+        }
+
+        size--;
```
okuzan, line 80:
```suggestion
        System.arraycopy(elements, index + 1, elements, index, --size - index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
     public void add(T value) {
+        growIfNeed();
 
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        isIndexValid(index, index > size);
```
okuzan, line 34: The only place where condition differs is here. Put that exclusive check here, without delegating to utility checkers, and then you can remove that weird second argument :) If you had more "diversity" in terms of acceptable limits, it would make sense to pass said limit (if the condition is >= you would pass size + 1 here and size everywhere else ) - passing a boolean is a bit unusual and redundant - you pass index twice

okuzan (changes requested):
Props to you for the interesting approach :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        isIndexValid(index);
+
+        T oldValue = elements[index];
+        if (index == size - 1) {
+            elements[index] = null;
+        }
+
```
```suggestion
```
arraycopy will handle this, no need to do it manually

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
+            T[] extendedArray =
+                    (T[]) new Object[valuesLength + (int) (valuesLength * GROW_FACTOR)];
+            System.arraycopy(elements, 0, extendedArray, 0, valuesLength);
+            elements = extendedArray;
+        }
+    }
+
+    private void isIndexValid(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index " + index);
```
okuzan, line 119: including size in the message would facilitate debugging

okuzan (approved):
Well done!

src/main/java/core/basesyntax/ArrayList.java
RomanMenzheres: I forgot to add return statement. It's need to avoid redundant call of arraycopy.

PR #1148, opened 2023-09-20
Realized ArrayList and overrided all List functions. Created grow() f…

Trayvor:
…unc to make array bigger and toArray() func to make an array from ArrayList

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -4,114 +4,104 @@ import java.util.NoSuchElementException;
 
 public class ArrayList<T> implements List<T> {
     public static final double GROW_VALUE = 1.5;
-    private int maxSize = 10;
-    private int actualSize = 0;
-    private Object[] array = new Object[maxSize];
-    private Object[] tempArray;
+    private static final int DEFAULT_SIZE = 10;
+    private int size = 0;
+    private T[] array;
+    private T[] tempArray;
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_SIZE];
+        tempArray = (T[]) new Object[DEFAULT_SIZE];
+    }
 
     @Override
     public void add(T value) {
-        if (actualSize == maxSize) {
+        if (size == array.length) {
             grow();
         }
-        array[actualSize] = value;
-        actualSize++;
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-        if (index > actualSize || index < 0) {
-            throw new ArrayListIndexOutOfBoundsException("Index doesn't exist");
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index
+                    + " doesn't exist, array size is " + array.length);
         }
-        if (actualSize == maxSize) {
+        if (size == array.length) {
             grow();
         }
-        tempArray = new Object[maxSize];
-        System.arraycopy(array, 0, tempArray, 0, index);
+        tempArray = array.clone();
         tempArray[index] = value;
-        System.arraycopy(array, index, tempArray, index + 1, actualSize - index);
-        actualSize++;
-        array = tempArray;
+        System.arraycopy(array, index, tempArray, index + 1, size - index);
+        size++;
+        array = tempArray.clone();
     }
 
     @Override
     public void addAll(List<T> list) {
-        int requiredSize = maxSize - actualSize + list.size();
-        while (requiredSize > maxSize) {
-            grow();
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final double GROW_VALUE = 1.5;
+    private int maxSize = 10;
+    private int actualSize = 0;
```
```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+    private Object[] array = new Object[maxSize];
+    private Object[] tempArray;
```
okuzan, line 10: - temp array shouldn't be a field
- initialization should happen in the constructor, as mentioned in the checklist
- use type parameter instead of Object

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+
     @Override
     public void add(T value) {
-
+        if (actualSize == maxSize) {
+            grow();
+        }
+        array[actualSize] = value;
+        actualSize++;
```
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > actualSize || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index doesn't exist");
```
okuzan, line 24: include index and size in the message to simplify debugging

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        }
+        if (actualSize == maxSize) {
+            grow();
+        }
+        tempArray = new Object[maxSize];
+        System.arraycopy(array, 0, tempArray, 0, index);
```
okuzan, line 30: in each method of this class, use arraycopy only once, you can pass the same array as both src and dest

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        tempArray[index] = value;
+        System.arraycopy(array, index, tempArray, index + 1, actualSize - index);
+        actualSize++;
+        array = tempArray;
     }
 
     @Override
     public void addAll(List<T> list) {
```
okuzan, line 38: here you can just call add method in a for loop

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        System.arraycopy(toArray(list), 0, tempArray, actualSize, list.size());
+        actualSize = actualSize + list.size();
+        array = tempArray;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= actualSize || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index doesn't exist");
```
okuzan, line 53: include index and size . extract this checker to a private utility method to avoid code duplication

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if ((array[i] != null && array[i].equals(element))
+                    || (array[i] == null && element == null)) {
+                remove(i);
+                return element;
+            }
+        }
+        throw new NoSuchElementException("There is no such element in array");
```
okuzan, line 89: include element in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        array = tempArray;
+        return removedItem;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < array.length; i++) {
+            if ((array[i] != null && array[i].equals(element))
+                    || (array[i] == null && element == null)) {
```
okuzan, line 84: remove extra parentheses
```suggestion
            if ((array[i] != null && array[i].equals(element))
                    || (array[i] == element)) {
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Trayvor: why its better to use for loop instead of arraycopy?

Ivan95kos (approved):
Nice

PR #1151, opened 2023-09-26
Overrided the methods

Nataliia-Ladzoryshyn:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    private void checkIndex(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Excepted index is incorrect");
+        }
+    }
+
+    private void checkIndexEqualsSize(int index) {
+        if (index == size) {
+            throw new ArrayListIndexOutOfBoundsException("Excepted index is incorrect");
+        }
```
olekskov, line 118: we may easily remove this method
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+            }
+        }
+        return -1;
+    }
+
+    private void checkIndex(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Excepted index is incorrect");
+        }
+    }
```
olekskov, line 113: include check for == here
```suggestion
    private void checkIndex(int index) {
        if (index >= size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Excepted index is incorrect");
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        checkIndexEqualsSize(index);
```
olekskov, line 45: then remove usage
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
+        checkIndexEqualsSize(index);
```
olekskov, line 51:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        T oldValue = (T) elementData[index];
+        elementData[index] = value;
 
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        checkIndexEqualsSize(index);
```
olekskov, line 61:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+        if (size == elementData.length) {
+            expandCapacity();
+        }
+        elementData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index);
```
olekskov, line 26: and add here reuse of existing method to solve problem that forced you to create 2 methods for check
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            expandCapacity();
+        }
```
olekskov, line 20: Please move if statement into expandCapacity() method as we always need to check this condition before resize. Better to make public method simpler

olekskov (changes requested):
Please check comments above

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
 package core.basesyntax;
 
+import com.google.common.base.Objects;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
```
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
```
```suggestion
    private T[] elementData;
```
that's where we should use type parameter

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
```
okuzan, line 13: you will need a cast before initialization

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            if (Objects.equal(element, elements[i])) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Excepted index is incorrect");
```
okuzan, line 110: a good practice is to add an index and size to facilitate debugging

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void expandCapacity() {
+        if (size == elementData.length) {
+            int newCapacity = (int) (elementData.length * 1.5);
```
okuzan, line 91:
```suggestion
            int newCapacity = (int) (elementData.length * GROW_FACTOR);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        return oldElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+
+        int index = indexOf(element);
+        if (index == -1) {
+            throw new NoSuchElementException("Element  not found");
```
okuzan, line 74: add this element to the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
-        return null;
+        checkIndex(index);
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
+        checkIndex(index);
+        T oldValue = (T) elementData[index];
+        elementData[index] = value;
```
okuzan, line 53: remove this line

okuzan (changes requested):

kozhukhovsky (approved):

PR #1155, opened 2023-09-28
implemented ArrayList class

hehetenya:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(data[i], element)) {
```
kshuryhin, line 72: Please, try to find workaround and do not use class `Objects`

[Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/array-list.html#dont-use-class-objects)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= MINIMAL_ARRAY_INDEX;
+    }
+
+    private Object[] grow() {
+        int oldLength = data.length;
+        Object[] newArray = new Object[(int) (oldLength * GROW_COEFFICIENT)];
+        if (size >= 0) {
```
kshuryhin, line 92: Do we really need to check, if the `size >= 0`? It looks like it always is >= 0.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        data[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
kshuryhin, line 41: Probably it would be better to use _for-each_ here, because sometimes, `get(i)` operations could be really expensive, especially in case of `LinkedList`.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
-
+        checkIndexForGetOperation(index);
+        data[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexForGetOperation(index);
+        T removedObject = get(index);
```
kshuryhin, line 59: Probably we can get rid of `checkIndexForGetOperation` in this method, as it is already present in `get(index)`.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= MINIMAL_ARRAY_INDEX;
```
kshuryhin, line 86: Minor, but probably it will me sufficient enough to check if our `size == MINIMAL_ARRAY_INDEX`

kshuryhin (changes requested):
Great job! Some minor points to consider

src/main/java/core/basesyntax/ArrayList.java
hehetenya: it says that "foreach not applicable to type 'core.basesyntax.List<T>'". Or should there be java.util.List?

src/main/java/core/basesyntax/ArrayList.java
hehetenya: Is using Objects.equals() a bad practice or is it only for this task?

proxychang (approved):

proxychang (approved):
Good job!

bhdnchui (approved):
Good job!

PR #1156, opened 2023-09-29
wrote an implementation of custom ArrayList class and its overriding …

RhymeWriter:
…methods of List interface

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,132 @@
 package core.basesyntax;
 
-public class ArrayList<T> implements List<T> {
-    @Override
-    public void add(T value) {
+import java.util.Arrays;
+import java.util.NoSuchElementException;
 
+public class ArrayList<T> implements List<T> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] storage;
+    private int size;
+
+    public ArrayList() {
+        storage = (T[]) new Object[DEFAULT_CAPACITY];
     }
 
     @Override
-    public void add(T value, int index) {
+    public void add(T value) {
+        if (size == storage.length) {
+            grow();
+        }
+        storage[size] = value;
+        size++;
+    }
 
+    @Override
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        if ((index < 0) || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index! " + index);
+        }
+        if (size == storage.length) {
+            grow();
+        }
+        System.arraycopy(storage, index, storage, index + 1, size - index);
+        storage[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (list.size() > (storage.length - size)) {
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            storage[size] = list.get(i);
+            size++;
+        }
     }
 
     @Override
-    public T get(int index) {
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        return storage[index];
     }
 
     @Override
-    public void set(T value, int index) {
-
+    public void set(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        this.storage[index] = value;
     }
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        T valueToBeRemoved = storage[index];
+        if (size > 0) {
+            System.arraycopy(storage, 0, storage, 0, index);
+            System.arraycopy(storage, index + 1, storage, index, size - index - 1);
+            size--;
+        }
+        return valueToBeRemoved;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T value) throws NoSuchElementException {
+        containsValue(value);
+        T valueToBeRemoved = null;
+        if (size > 0) {
+            T[] resultStorage = (T[]) new Object[storage.length];
+            int resultIndex = 0;
+            for (int i = 0; i < size; i++) {
+                if ((value == null && storage[i] == null)
+                        || (value != null && value.equals(storage[i]))) {
+                    valueToBeRemoved = storage[i];
+                    size--;
+                } else {
+                    resultStorage[resultIndex++] = storage[i];
+                }
+            }
+            storage = resultStorage;
+        }
+        return valueToBeRemoved;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        storage = Arrays.copyOf(storage, (int) (storage.length * 1.5));
+    }
+
+    private void getIndexValidation(int index) throws ArrayListIndexOutOfBoundsException {
+        if ((index < 0) || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index! " + index);
+        }
+    }
+
+    private void containsValue(T value) throws NoSuchElementException {
+        if (value == null) {
+            for (T data : storage) {
+                if (data == null) {
+                    return;
+                }
+            }
+        } else {
+            for (T data : storage) {
+                if (value.equals(data)) {
+                    return;
+                }
+            }
+        }
+        throw new NoSuchElementException("There is no such value in the storage! " + value);
     }
 }
+
+


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
     }
 
     @Override
-    public void add(T value, int index) {
+    public void add(T value) {
+        if (size == storage.length) {
+            grow();
+        }
+        storage[size] = value;
+        size++;
```
```suggestion
        storage[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+                if (value.equals(data)) {
+                    return;
+                }
+            }
+        }
+        throw new NoSuchElementException("There is no such value in the storage! " + value);
     }
 }
+
+
```
okuzan, line 132: remove two lines

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+                if (data == null) {
+                    return;
+                }
+            }
+        } else {
+            for (T data : storage) {
+                if (value.equals(data)) {
+                    return;
+                }
+            }
```
okuzan, line 126: combine conditions - one for loop, inside of it one if with combined condition. But since method is used only once, better just move its logic to the place where it's called

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        storage = Arrays.copyOf(storage, (int) (storage.length * 1.5));
```
okuzan, line 105: we can only use systems.arraycopy to move elements

```suggestion
        storage = Arrays.copyOf(storage, (int) (storage.length * GROW_FACTOR));
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        T valueToBeRemoved = storage[index];
+        if (size > 0) {
+            System.arraycopy(storage, 0, storage, 0, index);
+            System.arraycopy(storage, index + 1, storage, index, size - index - 1);
```
okuzan, line 67: in each method. system.arraycopy should be used only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        getIndexValidation(index);
+        return storage[index];
     }
 
     @Override
-    public void set(T value, int index) {
-
+    public void set(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        this.storage[index] = value;
```
okuzan, line 58: get rid of redundant this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+                        || (value != null && value.equals(storage[i]))) {
+                    valueToBeRemoved = storage[i];
+                    size--;
+                } else {
+                    resultStorage[resultIndex++] = storage[i];
+                }
+            }
+            storage = resultStorage;
+        }
+        return valueToBeRemoved;
```
okuzan, line 91: method is overengineered. Remove exception from method declaration, it's unchecked. You don't need second storage and swap them. For loop -> if with equals condition -> pass index to remove by index method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        this.storage[index] = value;
     }
 
     @Override
-    public T remove(int index) {
-        return null;
+    public T remove(int index) throws ArrayListIndexOutOfBoundsException {
+        getIndexValidation(index);
+        T valueToBeRemoved = storage[index];
+        if (size > 0) {
```
okuzan, line 65: remove if, system arraycopy handles everything fine

okuzan (changes requested):
Good job overall

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    private void grow() {
+        T[] newStorage = (T[]) new Object[(int) (storage.length * GROW_FACTOR)];
+        System.arraycopy(storage, 0, newStorage, 0, size);
+        storage = newStorage;
+    }
+
+    private void getIndexValidation(int index) throws ArrayListIndexOutOfBoundsException {
+        if ((index < 0) || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index! " + index);
+        }
```
okuzan, line 109:
```suggestion
    private void checkIndex(int index) {
        if ((index < 0) || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Invalid index! " + index + " for size: " + size);
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public void addAll(List<T> list) {
-
+        while (list.size() > (storage.length - size)) {
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            storage[size] = list.get(i);
+            size++;
+        }
```
okuzan, line 45: just call add method in the loop, all the necessary logic is there, if add is implemented correctly, so will be this  method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    public void add(T value) {
+        if (size == storage.length) {
+            grow();
+        }
+        storage[size++] = value;
+    }
 
+    @Override
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
+        if ((index < 0) || index > size) {
```
okuzan, line 26:
```suggestion
        if (index < 0 || index > size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+                    System.arraycopy(storage, i + 1, storage, i, size - i);
+                    storage[size] = null;
+                    break;
+                }
+            }
+            if (valueToBeRemoved == null && value != null) {
+                throw new NoSuchElementException("There is no such value in the storage! " + value);
+            }
+        }
+        return valueToBeRemoved;
```
okuzan, line 87:
```suggestion
    for (int i = 0; i < size; i++) {
        if (list[i] != null && list[i].equals(element) || list[i] == element) {
            return remove(i);
        }
    }
    throw new NoSuchElementException("No such value: " + element);
```

okuzan (changes requested):
Good job on the improvements, however code can be further simplified

src/main/java/core/basesyntax/ArrayList.java
RhymeWriter: wow, this is a really cool punch, nice! thanks!

src/main/java/core/basesyntax/ArrayList.java
RhymeWriter: this realization realy easely to read, thanks

src/main/java/core/basesyntax/ArrayList.java
olekskov: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
-    public void add(T value, int index) {
+    public void add(T value) {
+        if (size == storage.length) {
+            grow();
+        }
+        storage[size++] = value;
+    }
 
+    @Override
+    public void add(T value, int index) throws ArrayListIndexOutOfBoundsException {
```
olekskov, line 25: ArrayListIndexOutOfBoundsException extends RuntimeException
there is no need to add it to method headers, please recheck all occurrences

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    public ArrayList() {
+        storage = (T[]) new Object[DEFAULT_CAPACITY];
     }
 
     @Override
-    public void add(T value, int index) {
+    public void add(T value) {
+        if (size == storage.length) {
+            grow();
+        }
```
olekskov, line 20: this condition is always checked before grow() call, please move it inside method

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
```
okuzan, line 6:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
okuzan: not fixed - method can be simplified, reuse logic from remove by index

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
-        return null;
+    public T remove(T value) {
+        T valueToBeRemoved;
+        for (int i = 0; i < size; i++) {
+            if (storage[i] != null && storage[i].equals(value) || storage[i] == value) {
+                valueToBeRemoved = storage[i];
+                size--;
+                System.arraycopy(storage, i + 1, storage, i, size - i);
+                storage[size] = null;
+                return valueToBeRemoved;
```
okuzan, line 70:
```suggestion
                T valueToBeRemoved = storage[i];
                System.arraycopy(storage, i + 1, storage, i, size - i);
                storage[--size] = null;
                return valueToBeRemoved;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        System.arraycopy(storage, index + 1, storage, index, size - index - 1);
+        size--;
+        return valueToBeRemoved;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T value) {
+        T valueToBeRemoved;
```
okuzan, line 63:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        for (int i = 0; i < size; i++) {
+            if (storage[i] != null && storage[i].equals(value) || storage[i] == value) {
+                valueToBeRemoved = storage[i];
+                size--;
+                System.arraycopy(storage, i + 1, storage, i, size - i);
+                storage[size] = null;
+                return valueToBeRemoved;
+            }
+
```
```suggestion
```

okuzan (changes requested):

PR #1157, opened 2023-09-30
overrided methods for class ArrayList

StVadym:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] letItGrow() {
+        return arrayList = Arrays.copyOf(arrayList, (int)Math. round(arrayList.length * 1.5));
```
Ivan95kos, line 135: use constants instead of numbers

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        arrayList[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index should be in the range 0 to "
+                    + size);
+        }
```
Ivan95kos, line 37: move this check to private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
 
+        } else {
+            System.arraycopy(arrayList, index, arrayList, index + 1, size);
+        }
+        arrayList[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
```
Ivan95kos, line 52: call the method `add`
like:
```
            add(list.get(i));
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index > size() - 1) {
+            throw new ArrayListIndexOutOfBoundsException("Index should be in the range 0 to "
+                    + size);
+        }
```
Ivan95kos, line 67: move this check to private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        return arrayList[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index > size() - 1) {
+            throw new ArrayListIndexOutOfBoundsException("Index should be in the range 0 to "
+                    + size);
+        }
```
Ivan95kos, line 76: the same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            System.arraycopy(arrayList, index, arrayList, index, 1);
+        } else {
+            System.arraycopy(arrayList, index + 1, arrayList, index, size - 1);
+        }
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
```
Ivan95kos, line 97: try to find the index and call the method `remove(int index)`

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
```
Ivan95kos, line 85: It is hard to read
rewrite like this:

```
    @Override
    public T remove(T element) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Elements do not exist" + element);
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        arrayList[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (arrayList.length <= size + list.size()) {
+            arrayList = letItGrow();
+        }
```
Ivan95kos, line 53: You have this logic in the `add` method
```suggestion
```

Ivan95kos (changes requested):
a few commnets

Ivan95kos (approved):
Nice

PR #1158, opened 2023-10-03
Created own ArrayList class

HovorukhaBohdan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_LENGTH];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        arrayIncrease(size + 1);
```
kshuryhin, line 17: I would say that is a bit confusing when we are calling arrayIncrease each time. Suggest to call it only, if internal array is full and make this condition explicit.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index is not exists");
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int temp = size;
+        for (int i = 0; i < list.size(); i++) {
+            arrayIncrease(size + 1);
```
kshuryhin, line 38: You are not correctly resizing the array for the `addAll` method. You should calculate the total required size first and then resize if necessary.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        int index = findIndex(element);
+        if (index != -1) {
+            System.arraycopy(data, index + 1, data, index, size - index);
+            size--;
+        } else {
+            throw new NoSuchElementException("Index is not exists");
```
kshuryhin, line 87: Error message here is a bit misleading, because we are trying to remove element, but no an index.

kshuryhin (changes requested):
Good job, but there are some changes needed to be done

Alexander1iubchenko (approved):
Everything seems ok to me, good job

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_LENGTH = 10;
+    private T[] data;
+    private int size;
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_LENGTH];
+        size = 0;
```
bhdnchui, line 12: I think it is redundant.
In java default int value is 0.
https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        int temp = size + list.size();
 
+        if (temp > data.length) {
+            arrayIncrease(temp + 1);
+        }
+
+        int fromIndex = size;
+        for (int i = 0; i < list.size(); i++) {
+            data[i + fromIndex] = list.get(i);
+            size++;
```
bhdnchui, line 51:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        for (int i = 0; i < data.length; i++) {
+            if (data[i] == type || data[i] != null && data[i].equals(type)) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private void arrayIncrease(int length) {
+        T[] copy = (T[]) new Object[length + length >> 1];
```
bhdnchui, line 128: image

bhdnchui (changes requested):
Good job! But a few comments need to be processed.

bhdnchui (approved):
Good job!

PR #1162, opened 2023-10-08
realized ArrayList class

fedorovychh:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+    private static final int NO_SUCH_ELEMENT_INDEX = -1;
+    private static final String NO_SUCH_ELEMENT_MESSAGE = "There is no such element in list.";
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "This index does not exist in list.";
+
+    private Object[] data;
+    private int size;
+
+    public ArrayList() {
+        data = new Object[DEFAULT_CAPACITY];
+        size = EMPTY_LIST_VALUE;
```
DevIgork, line 21: I don't know how necessary it is to initialize the primitive in the constructor, let the mentor decide

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int BIT_SHIFT_VALUE = 1;
+    private static final int ONE_ELEMENT_SIZE = 1;
+    private static final int EMPTY_LIST_VALUE = 0;
+    private static final int FIRST_ELEMENT_INDEX = 0;
+    private static final int NO_SUCH_ELEMENT_INDEX = -1;
+    private static final String NO_SUCH_ELEMENT_MESSAGE = "There is no such element in list.";
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "This index does not exist in list.";
+
+    private Object[] data;
```
DevIgork, line 16: you can create an array of generics

DevIgork (approved):
I saw only 2 things that can be changed, but it is up to the mentor to decide

src/main/java/core/basesyntax/ArrayList.java
proxychang: Yes, according to [this](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

proxychang:
> I saw only 2 things that can be changed, but it is up to the mentor to decide

thanks!

src/main/java/core/basesyntax/ArrayList.java
proxychang: I think it is not necessary because when will runtime be it will resolves to Object, so we can declare Object array instead of generics array

image

src/main/java/core/basesyntax/ArrayList.java
kshuryhin: @proxychang Well, I believe that is the case, when we are initialising our instance variables with some non-default values. Here, it is absolutely logical per `ArrayList` creation to have default(0) `size` value. Although it is not a mistake, that could be a bit confusing to see we are doing `size` non-default initialisation.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+    private int size;
+
+    public ArrayList() {
+        data = new Object[DEFAULT_CAPACITY];
+        size = EMPTY_LIST_VALUE;
+    }
+
+    private Object[] grow() {
+        data = data.length > EMPTY_LIST_VALUE
```
kshuryhin, line 25: Probably this check is a bit redundant, as our array length is always at minimum `DEFAULT_CAPACITY`, which is 10 in our case, what do you think?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+                ? Arrays.copyOf(data, data.length + (data.length >> BIT_SHIFT_VALUE))
+                : new Object[DEFAULT_CAPACITY];
+        return data;
+    }
+
+    private int getIndexByElement(T element) {
+        for (int i = 0; i < data.length; i++) {
```
kshuryhin, line 32: Probably here we can iterate only up to `size`, not the entire `data` array.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
-        return null;
+        checkGetByIndex(index);
+        return (T) data[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkGetByIndex(index);
+        data[index] = data[index] != null ? value : data[index];
```
kshuryhin, line 90: But what if we want to set non-null value to the position, where currently `null` is located?
Try to run such test

```
    @Test
    public void setNonNullValueAtNullPosition() {
        ArrayList<String> arrayList = new ArrayList<>();
       arrayList.add("5");
       arrayList.add(null);
       arrayList.set("10", 1);
       Assert.assertEquals("10", arrayList.get(1)); //false
       Assert.assertEquals(null, arrayList.get(1)); //true
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        }
+        size--;
+        return removedObject;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        checkElementByIndex(getIndexByElement(element));
+        int elementIndex = getIndexByElement(element)
```
kshuryhin, line 106: Here you are calling `gerIndexByElement` several times, consider to fix that.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+
+    private void checkAddByIndex(int index) {
+        if (index < EMPTY_LIST_VALUE || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE);
+        }
+    }
+
+    private void checkGetByIndex(int index) {
+        if (index < EMPTY_LIST_VALUE || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE);
```
kshuryhin, line 48: Here and for other error messages, we can make them more informative, like which indexes, which elements etc.

kshuryhin (changes requested):
Great job, but there are some points to consider!

OlexVoit (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+    }
+
+    private Object[] grow() {
+        data = Arrays.copyOf(data, data.length + (data.length >> BIT_SHIFT_VALUE));
+        return data;
+    }
+
+    private int getIndexByElement(T element) {
+        for (int i = 0; i < size; i++) {
+            if (data[i] == element || (data[i] != null && data[i].equals(element))) {
```
mnyshenko, line 30: don't forget to extract expression into a separate variable when you use it multiple times

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == FIRST_ELEMENT_INDEX;
     }
```
mnyshenko, line 120: It's always better to put public methods before private because when you open some class, you want to see its general logic, not some specific private methods that **serve** public methods

mnyshenko (approved):
Approved, but left a few minor comments to consider in future

PR #1166, opened 2023-10-09
Add implementation for the ArrayList.

MinoDentori:
1. Override methods from the List interface.
2.  Create methods checkIndex and checkIndexForAdd. They are used in add by index, set by index, and remove by index methods.
3. Create a grow method, which allows ArrayList to grow when needed (1.5X by default, or in a large specific number if default growing is not enough).
4. Support method findElement allows to find element by value, not index.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,152 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    public ArrayList(int capacity) {
+        elementData = new Object[capacity];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (isFull()) {
+            elementData = grow();
+        }
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        if (isFull()) {
+            elementData = grow();
+        }
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int numberOfValuesToAdd = list.size();
+        Object[] valuesToAdd = convertListToArray(list);
+        if (numberOfValuesToAdd == 0) {
+            return;
+        }
+        if (numberOfValuesToAdd > elementData.length - size) {
+            elementData = grow(size + numberOfValuesToAdd);
+            System.arraycopy(valuesToAdd, 0, elementData, size, numberOfValuesToAdd);
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,152 @@
...
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
kshuryhin, line 14: Probably we do not need explicitly initialise `size` here. Point from **Common Mistakes** was mostly related to cases, when we need to make some non-default initialisation.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,152 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int indexOfElement = findElement(element);
+        return remove(indexOfElement);
+    }
+
+    private void fastRemove(int index) {
```
kshuryhin, line 82: Probably, `internalRemove` will be more suitable here, what do you think? `fastRemove` makes me feel it is somehow optimised.

kshuryhin (changes requested):
Awesome! Just some minor comments

src/main/java/core/basesyntax/ArrayList.java
MinoDentori: After you told me, I felt the same way. In case we have the "remove" method, the "internalRemove" name will be the best neutral renaming variant. Fixed.

MinoDentori:
Fix comments

Vladlesn (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,150 @@
...
+        }
+    }
+
+    private Object[] grow() {
+        return grow(size + 1);
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
```
d1sam, line 118: Let's better use messages as constants.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,150 @@
...
+                    + " is out of bounds of array");
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
```
d1sam, line 125: Let's better use messages as constants.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,150 @@
...
+        return size == elementData.length;
+    }
+
+    private int findElement(T elementToFind) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elementToFind, elementData[i])) {
+                return i;
+            }
+        }
+        throw new NoSuchElementException("There is no " + elementToFind + " in the List");
```
d1sam, line 140: Let's better use messages as constants.

d1sam (approved):
Good job, but please be careful with String literals as messages.

MinoDentori:
Fixed comments

PR #1171, opened 2023-10-11
Create first version of my custom ArrayList

ponomvrenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 0;
+    private static final int DEFAULT_SIZE = 0;
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_CAPACITY];
+        size = DEFAULT_SIZE;
```
vadymhrnk, line 14: It's up to you but I would rather start with larger `DEFAULT_CAPACITY` so we will have less iterations of growing the initial array.
Also, initializing size as `private static final int DEFAULT_SIZE = 0;` is redundant. `int` by default is 0.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > array.length) {
+            int newCapacity = array.length + (array.length >> 1);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
```
vadymhrnk, line 22: I think this code would be redundant if size of initial array would be 10 for example.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-
+        ensureCapacity(size + 1);
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
```
vadymhrnk, line 36: For me, this error message is not informative enough

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        T oldValue = get(index);
+        int numMoved = size - index - 1;
+        if (numMoved > 0) {
+            System.arraycopy(array, index + 1, array, index, numMoved);
+        }
+        array[--size] = null;
```
vadymhrnk, line 71: I think you could just decrement size without assigning a null

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(array[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
vadymhrnk, line 89: You could simplify this block by combining two if statements in one

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void validateIndex(int index) {
```
vadymhrnk, line 103: This method is separated with `private void ensureCapacity(int minCapacity)` by overridden methods, I think you should connect them here, or on top.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
```
vadymhrnk, line 105: Same as other `ArrayListIndexOutOfBoundsException`, might be more informative

vadymhrnk (changes requested):

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: I don't think so. The logic of this method is removing element by index in ArrayList, after that we have to shift our copy array to fill empty index inside. After that we have to appoint last empty element by null👀

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: Okay👌🏻 It simplifies the logic and reduces redundancy

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: 1. Initial capacity: 
This can be beneficial for performance, especially in situations where the list is expected to grow significantly. 
2. Initializing:
I prefer explicitly initializing all fields to make the code more explicit and self-documenting.

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: Yes, I can agree with you, but for me it was necessary to understand the logic of this method. It is crucial for dynamically resizing the array when adding elements, reducing unnecessary resizing operations, and the growth factor that impacts performance and memory usage.

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: Agree 👍🏻

src/main/java/core/basesyntax/ArrayList.java
ponomvrenko: `validateIndex()` is primarily concerned with checking the validity of an index, while `ensureCapacity()` focuses on managing the array's capacity. These are distinct concerns, and it's often beneficial to keep them separate for clarity.

vadymhrnk:
Good job! 👍 I see that your homework status changed to mentor review, so it's probably better leave the status of the pull request how it shows right now.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 0;
```
bhdnchui, line 7: Let's make it 10, as in java ArrayList.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_CAPACITY];
+        size = DEFAULT_SIZE;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > array.length) {
+            int newCapacity = array.length + (array.length >> 1);
```
bhdnchui, line 19: Which newCapacity value will be if your array is 0 length?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            if (newCapacity < minCapacity) {
```
bhdnchui, line 20: Is it be false ever? We should not do anything in false case?

bhdnchui (changes requested):
Good job, but a few comments need to be processed.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(size + list.size());
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
d1sam, line 50:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

d1sam: you have ensureCapacity inside add method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + " out of bound!!"
```
d1sam, line 99: make message as a constant.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && array[i] == null)
+                    || (element != null && element.equals(array[i]))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
d1sam, line 84: make message as a constant.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        if (array.length == 0) {
+            int newCapacity = Math.max(DEFAULT_CAPACITY, minCapacity);
+            array = Arrays.copyOf(array, newCapacity);
+        } else if (minCapacity > array.length) {
+            int newCapacity = array.length + (array.length >> 1);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
+            array = Arrays.copyOf(array, newCapacity);
+        }
```
d1sam, line 27:
```suggestion
         if (minCapacity > array.length) {
            int newCapacity = (int) (array.length * 1.5);
            Object[] newData = new Object[newCapacity];
            System.arraycopy(array, 0, newData, 0, size);
            array = newData;
        }
```

d1sam (changes requested):
Good job, but there are some changes should be done.

d1sam (approved):
Well done.

PR #1176, opened 2023-10-11
implemented class ArrayList

YevheniiVlasenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int BASE_SIZE = 10;
+    private static final String INDEX_EXCEPTION_MESSAGE = "Provided index "
```
forzen123212, line 7: Just use IndexOutOfBounds exception in method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int BASE_SIZE = 10;
```
forzen123212, line 6: It would be better use BASE_CAPACITY name or rename size value to currentAmountofValues

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    public ArrayList() {
+        this.objects = new Object[BASE_SIZE];
+    }
+
+    private void checkFreeSpace(int neededSpace) {
+        if (neededSpace > objects.length) {
+            int newLength = (int)(objects.length * 1.5);
+            if (newLength < neededSpace) {
+                newLength = neededSpace;
+            }
```
forzen123212, line 21: Remove double if construction, just check current capacity and if you need to grow your array use your arraycopy with objectLength * 1.5 argument

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+            Object[] newArray = new Object[newLength];
+            System.arraycopy(objects, 0, newArray,0, objects.length);
+            objects = newArray;
+        }
+    }
+
+    private void sizeIncrease() {
```
forzen123212, line 28: You will use size++ only once, remove this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        checkFreeSpace(size + 1);
+        objects[size] = value;
+        sizeIncrease();
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_EXCEPTION_MESSAGE);
```
forzen123212, line 48: use validateindex

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        checkFreeSpace(size + 1);
+        objects[size] = value;
+        sizeIncrease();
```
forzen123212, line 42: checkFreeSpace(size + 1);
objects[size++] = value;

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0 || objects == null;
```
forzen123212, line 103: don't need to check null

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        T returnValue = (T) objects[index];
+        System.arraycopy(objects, index + 1, objects, index, size - index - 1);
+        objects[--size] = null;
+        return returnValue;
     }
```
forzen123212, line 84: You dont need to copy array if element that will be removed is last, think about realization

forzen123212 (changes requested):

src/main/java/core/basesyntax/ArrayList.java
YevheniiVlasenko: Validation conditions are different

src/main/java/core/basesyntax/ArrayList.java
YevheniiVlasenko: and if you are comparing your array to the list and array.length + list.size > array.length * 1.5?

src/main/java/core/basesyntax/ArrayList.java
forzen123212: we have class by condition, you have same message in this exception throw
if (index < 0 || index > upperBound) {
            throw new ArrayListIndexOutOfBoundsException(
                    String.format("Index %d out of bounds for size %d", index, currentSize));
                    
                    use this instead

forzen123212 (approved):
Good job!

src/main/java/core/basesyntax/ArrayList.java
YevheniiVlasenko: > we have class by condition, you have same message in this exception throw if (index < 0 || index > upperBound) { throw new ArrayListIndexOutOfBoundsException( String.format("Index %d out of bounds for size %d", index, currentSize));
> 
> ```
>                 use this instead 
> ```

I think my realization is much simpler and readable. And it also provides enough information on the exception case. Readability etc.)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-        return null;
+        validateIndex(index);
+        T returnValue = (T) objects[index];
+        if (index == size - 1) {
+            objects[--size] = null;
+            return returnValue;
+        }
+        System.arraycopy(objects, index + 1, objects, index, size - index - 1);
+        objects[--size] = null;
+        return returnValue;
```
mnyshenko, line 86: `objects[--size] = null;` 
statement is written twice here, let's think if we can try to simplify this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null ? objects[i] == null : element.equals(objects[i])) {
+                return remove(i);
+            }
```
mnyshenko, line 94: `objects[i]` also used twice here, let's extract to a separate variable

mnyshenko: You check if element variable holds null in a for loop, we can determine this outside of this loop only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+
+    private void checkFreeSpace(int neededSpace) {
+        if (neededSpace > objects.length) {
+            int newLength = (int) (objects.length * 1.5);
+            if (newLength < neededSpace) {
+                newLength = neededSpace;
+            }
+            Object[] newArray = new Object[newLength];
+            System.arraycopy(objects, 0, newArray, 0, objects.length);
+            objects = newArray;
```
mnyshenko, line 24: The name says that this method should only check if there is a free space in the array, it should not change the array somehow. Please, refactor either the method name or the extract this logic somewhere else

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        sizeIncrease();
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        checkFreeSpace(size + list.size());
+        for (int i = 0; i < list.size(); i++) {
+            objects[size] = list.get(i);
+            sizeIncrease();
```
mnyshenko, line 60: I guess it's better to increase the size only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+    private static final String INDEX_EXCEPTION_MESSAGE = "Provided index "
+            + "is < 0 or > list size. Please, provide valid index.";
+    private Object[] objects;
+    private int size;
+
+    public ArrayList() {
+        this.objects = new Object[BASE_CAPACITY];
+    }
+
+    private void checkFreeSpace(int neededSpace) {
```
mnyshenko, line 16: It's always better to put public methods before private because when you open some class, you want to see its general logic, not some specific private methods that **serve** public methods

mnyshenko (changes requested):
A few points to consider

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(objects[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
d1sam, line 97: simplify this without `if-else`

d1sam (changes requested):
Good job, but some changes should be done.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        objects[--size] = null;
+        return returnValue;
     }
 
-    @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((objects[i] == null && element == null) || (element != null && element.equals(
```
d1sam, line 85:
```suggestion
            if ((objects[i] == element) || (element != null && element.equals(
```

d1sam (approved):
Good job, but please pay attention to my comment.

PR #1188, opened 2023-10-16
implement custom ArrayList class

denysokhrymovych:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY_FOR_ARRAY = 10;
+    private static final String INDEX_ERROR_MESSAGE =
+            "The passed index is not allowed";
+    private static final String ELEMENT_NOT_FOUND_MESSAGE = "Element not found";
+    private Object[] elementData;
```
DPotapov93, line 10: Do not use class Objects, you have to use generic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY_FOR_ARRAY];
```
DPotapov93, line 14: casting an object array to a generic array

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
     @Override
     public void add(T value, int index) {
-
+        validateIndex(index, true);
+        if (size == elementData.length) {
+            resize();
+        }
+        if (size == index) {
+            add(value);
+        } else {
```
DPotapov93, line 34: Don’t complicate if-else construction

DPotapov93 (changes requested):
Good job! But consider some minor changes

src/main/java/core/basesyntax/ArrayList.java
denysokhrymovych: Thank you for your review, but I use Object which is not the same as Objects. I chose Object because it is used in the standard ArrayList

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
-
+        validateIndex(index, false);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index, false);
+        final T element = this.get(index);
```
bhdnchui, line 67: Just get element by index from array instead of `this.get(index);`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+        System.arraycopy(elementData, index + 1, elementData, index,
+                size - index - 1);
+        size--;
+        elementData[size] = null;
```
bhdnchui, line 71:
```suggestion
        elementData[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = findIndexOfElementToRemove(element);
+        System.arraycopy(elementData, index + 1, elementData, index,
+                size - index - 1);
+        size--;
+        elementData[size] = null;
+        return element;
```
bhdnchui, line 82: It is the same code as in ` public T remove(int index)` method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return this.size() == 0;
+    }
+
+    private void resize() {
+        Object[] oldArray = elementData;
+        int newCapacityForArray = oldArray.length + (oldArray.length >> 1);
+        elementData = new Object[newCapacityForArray];
```
bhdnchui, line 98:
```suggestion
        elementData = new Object[oldArray.length + (oldArray.length >> 1)];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    private Object[] toArray(List<T> list) {
+        int listSize = list.size();
+        Object[] result = new Object[listSize];
+        for (int i = 0; i < listSize; i++) {
+            result[i] = list.get(i);
+        }
+        return result;
+    }
+
+    private void validateIndex(int index, boolean isForAdd) {
```
bhdnchui, line 111: Separate it into two methods.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+        if (!isForAdd && (index < 0 || index >= size)) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+    }
+
+    private int findIndexOfElementToRemove(T element) {
+        for (int i = 0; i < size; i++) {
+            T currentValue = this.get(i);
```
bhdnchui, line 122: Just get element by index from array instead of this.get(i);

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,130 @@
...
+    private void validateIndex(int index, boolean isForAdd) {
+        if (isForAdd && (index < 0 || index > size)) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+        if (!isForAdd && (index < 0 || index >= size)) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_ERROR_MESSAGE);
+        }
+    }
+
+    private int findIndexOfElementToRemove(T element) {
```
bhdnchui, line 120:
```suggestion
    private int findIndexOfElement(T element) {
```
I think it would be better.

src/main/java/core/basesyntax/ArrayList.java
bhdnchui, line 10: I think we can rename it more shorter and better. For example: `elements`

mnyshenko (approved):
Approved, but left a comment to consider

PR #1190, opened 2023-10-17
Added jv-array-list solution

Stanislav-Viner:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final String OUT_OF_BOUNDS = "Index [%d] is out of bounds for list of size [%d]";
+    private static final String NO_SUCH_ELEMENT = "No such element [%h] in this list";
+    private static final String ILLEGAL_CAPACITY = "Illegal capacity: %d";
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
```suggestion
    private T[] elements;
```
that's exactly where we should use type parameter

okuzan (approved):
Good job!

PR #1200, opened 2023-10-27
implement `ArrayList`

Paradoxallist:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
```
sokimaaa, line 7: Use generic type here. 
```suggestion
    private T[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index " + index + " out of bounds for length" + elementData.length);
+        }
```
sokimaaa, line 28: extract into separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        while (size + list.size() >= elementData.length) {
+            grow();
+        }
```
sokimaaa, line 41: grow instantly to the needed length, otherwise, it increases execution time. 
```suggestion
        grow();
```

sokimaaa (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    }
+
+    private void checkToCorrectIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(messageIndexOutOfBoundsException(index));
+        }
+    }
+
+    private void grow(int minCapacity) {
+        int newCapacity = Math.max((int)(elementData.length * 1.5f), minCapacity);
```
Sanderfuck, line 102: Save value in constant as resize factor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            grow(size + 1);
```
Sanderfuck, line 17: Don`t sent size its redundant, just create method check capacity or something else

Sanderfuck (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Paradoxallist: https://github.com/mate-academy/jv-array-list/pull/1200#discussion_r1374809713
I believe that in this place you need to pass the minimum size to which you need to increase the array, otherwise it increases the execution time.

Sanderfuck: The main thing of resizing that it check size and grow size automatically, you don`t need manage it by yourself

Sanderfuck: If you want enhancement performance of addAll you can make it only in that method without reusing method add

Sanderfuck (changes requested):

Rommelua (approved):

PR #1202, opened 2023-10-30
realised class ArrayList

aksoli666:
create const: DEFAULT_CAPACITY,
GROW_FACTOR
implements: add() - 3, get(), set(),
remove() - 2, size(), isEmpty()
added the methods grow(), checkIndex()

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -46,7 +46,7 @@ public class ArrayList<T> implements List<T> {
     public T get(int index) {
         checkIndexInclusive(index);
 
-        return (T) values[index];
+        return values[index];
     }
 
     @Override
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -5,12 +5,11 @@ import java.util.NoSuchElementException;
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
     private static final double GROWTH_FACTOR = 1.5;
-    private Object[] values;
+    private T[] values;
     private int size;
-    private T oldValue;
 
     public ArrayList() {
-        values = new Object[DEFAULT_CAPACITY];
+        values = (T[]) new Object[DEFAULT_CAPACITY];
     }
 
     @Override
@@ -22,9 +21,7 @@ public class ArrayList<T> implements List<T> {
 
     @Override
     public void add(T value, int index) {
-        if (index < 0 || index > size) {
-            throw new ArrayListIndexOutOfBoundsException("Index out of bounds " + index);
-        }
+        checkIndexExclusive(index);
 
         grow();
 
@@ -39,8 +36,6 @@ public class ArrayList<T> implements List<T> {
 
     @Override
     public void addAll(List<T> list) {
-        grow();
-
         for (int i = 0; i < list.size(); i++) {
             T element = list.get(i);
             add(element);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private Object[] values;
```
sokimaaa, line 8: Use generic type here. 
```suggestion
    private T[] values;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        grow();
+        values[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds " + index);
+        }
```
sokimaaa, line 27: potentially repeating code, should be extracted into private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
+        grow();
 
+        for (int i = 0; i < list.size(); i++) {
+            T element = list.get(i);
+            add(element);
+        }
```
sokimaaa, line 47: Method Add() already has the grow(), so it's redundant. 
```suggestion
        for (int i = 0; i < list.size(); i++) {
            T element = list.get(i);
            add(element);
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        checkIndex(index);
+
+        return (T) values[index];
     }
 
     @Override
     public void set(T value, int index) {
+        checkIndex(index);
 
+        oldValue = (T)values[index];
```
sokimaaa, line 61: why do you need to save the previous value? 
it's not been asked by task, so to be removed.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+
+        oldValue = (T)values[index];
```
sokimaaa, line 69: why do you need to save the previous value?
it's not been asked by task, so to be removed.

sokimaaa, line 69: you should instantiate the local var for the purpose of returning the removed value.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,123 @@
...
+                index = i;
+                break;
+            }
+        }
+
+        if (index == -1) {
+            throw new NoSuchElementException("Element not found " + element);
+        }
+
+        return remove(index);
```
sokimaaa, line 96: The code could be simplified. 
When you reach a situation when `index=i`, we could instantly use `return + remove(index)` to leave the method. Otherwise, an exception will be thrown. 
```suggestion
        for (int i = 0; i < size; i++) {
            if ((element == null && values[i] == null)
                    || (element != null && element.equals(values[i]))) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element not found " + element);
```

sokimaaa (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexInclusive(index);
+
+        return (T) values[index];
```
Sanderfuck, line 49: No need cast (T)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexInclusive(index);
+
+        final T removedValue = (T)values[index];
+
+        for (int i = index; i < size - 1; i++) {
+            values[i] = values[i + 1];
+        }
```
Sanderfuck, line 67: Use System.arraycopy

Sanderfuck (changes requested):

src/main/java/core/basesyntax/ArrayList.java
kerrrusha: not fixed
image

kerrrusha (changes requested):

Sanderfuck (approved):

PR #1205, opened 2023-11-01
Hw jv array list

AdamMudrak:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        int nextElementIndex = index + 1;
+        while (nextElementIndex < size) {
+            values[index++] = values[nextElementIndex++];
+        }
+        values[index] = null;
+        size--;
+        return element;
+    }
+
+    private void throwNewIndexOutOfBoundsException(int index) {
```
kozhukhovsky, line 111:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
-
+        growIfArrayFull();
+        values[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        growIfArrayFull();
+        size++;
```
kozhukhovsky, line 24: It's better to have two variables. Size - the number of occupied cells.
Capacity - the number of all cells.
In fact, it is better to remake the decision based on this information.

kozhukhovsky (changes requested):

AdamMudrak:
Not sure if that's what you meant...

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private int maxListSize = 10;
```
Ivan95kos, line 7: make the `maxListSize` a constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+    private T[] values;
+    private int capacity;
+    private int size;
```
Ivan95kos, line 10: useless variable
the `size` variable is enough
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+            for (int currentIndex = capacity - 2; currentIndex > index; currentIndex--) {
+                moveElementsToTheRight(currentIndex);
+            }
+        }
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
```
Ivan95kos, line 40: You can do it simpler
```
            add(list.get(i));
        }
```

Ivan95kos (changes requested):
left a few comments

src/main/java/core/basesyntax/ArrayList.java
AdamMudrak: But if I make it a constant, how do I change it when the limit is reached?

src/main/java/core/basesyntax/ArrayList.java
AdamMudrak: I added this because Anton said there must be two variables for two different purposes. Or is it easier not to touch it on this stage?

kozhukhovsky (approved):
Соррі, щодо коментаря з сapacity та size. Щось сам затупив, та тебе запутав. Вибачаюсь. :)

AdamMudrak:
> Соррі, щодо коментаря з сapacity та size. Щось сам затупив, та тебе запутав. Вибачаюсь. :)
> 

Не парься, з усіма буває ;)

PR #1211, opened 2023-11-03
Overridden ArrayList methods

FkordJava:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
+    public void indexValidator(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
+        }
+    }
```
sokimaaa, line 19: this method is not expected to be used outside this class, so make private.
don't forget to move it at the end of class, because of private.
```suggestion
    private void indexValidator(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] elementData;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
sokimaaa, line 12: You dont need to set size=0, because java makes the same instead of you. 
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+
+    public Object[] grow(T[] elementData, int size) {
+        if ((size + 1) >= elementData.length) {
+            int newCapacity = elementData.length + (elementData.length >> 1);
+            T[] newElementData = (T[]) new Object[newCapacity];
+            System.arraycopy(elementData, 0, newElementData, 0, size);
+            elementData = newElementData;
+        }
+        return elementData;
+    }
```
sokimaaa, line 29: Think about proper access modifier for grow method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        this.size = 0;
+    }
+
+    public void indexValidator(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
+        }
+    }
+
+    public Object[] grow(T[] elementData, int size) {
```
sokimaaa, line 21: Method grow should be void, because it changes the main array, e.g. this.elementData. 
So, you can directly access this  `this.elementData`.
```suggestion
    public void grow(T[] elementData, int size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
+        }
```
sokimaaa, line 42: Potentially repeating code should be extracted into private method.
```suggestion
        if (index < 0 || index > size) {
            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
        }
```

sokimaaa (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    private void addIndexValidator(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is not valid");
+        }
+    }
+
+    private void grow(T[] elementData, int size) {
+        if ((size + 1) >= elementData.length) {
+            int newCapacity = elementData.length + (elementData.length >> 1);
```
kerrrusha, line 105: binary operators reduces code readability. consider creating `GROWTH_FACTOR = 1.5` constant
```suggestion
            int newCapacity = elementData.length * GROWTH_FACTOR;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    return remove(i);
+                }
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
kerrrusha, line 78: can be more simple
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == null ? elementData[i] == null : element.equals(elementData[i])) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element not found");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public void add(T value, int index) {
-
+        addIndexValidator(index);
+        if (size == elementData.length) {
+            grow(elementData, size);
+        }
+        for (int i = size; i > index; i--) {
+            elementData[i] = elementData[i - 1];
+        }
```
kerrrusha, line 29: image

kerrrusha (changes requested):

PR #1215, opened 2023-11-06
ArrayList

rafik2211:
Constructor (ArrayList()): The class has a constructor that initializes an array of objects with a default capacity of 10 and sets the size to 0.
ensureCapacity(int minCapacity) Method: This private method ensures that the internal array has sufficient capacity to accommodate new elements. It checks the minimum required capacity, calculates the new capacity as 1.5 times the old capacity, and adjusts it if necessary. The elements array is then resized using Arrays.copyOf().
add(T value) Method: This method adds an element to the end of the list. It checks the capacity and, if necessary, resizes the array before adding the element.
add(T value, int index) Method: This method inserts an element at a specific index in the list. It checks if the index is within a valid range, resizes the array if needed, and shifts elements to make space for the new one.
addAll(List<T> list) Method: This method adds all elements from another list to the end of the current list by iterating through the elements of the given list and calling the add(T value) method for each.
get(int index) Method: This method retrieves an element from the list at the specified index. It performs boundary checks to ensure the index is within a valid range.
set(T value, int index) Method: This method updates the value of an element at the specified index in the list. It also checks the index for validity.
remove(int index) Method: This method removes an element from the list at the specified index and returns the removed element. It adjusts the array to fill the gap left by the removed element.
remove(T element) Method: This method removes the first occurrence of the specified element from the list by iterating through the list and using the remove(int index) method when a match is found.
size() Method: This method returns the current size (number of elements) of the list.
isEmpty() Method: This method checks whether the list is empty by verifying if the size is 0.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,27 +1,15 @@
 package core.basesyntax;
 
-import java.util.Arrays;
 import java.util.NoSuchElementException;
 
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
-    private Object[] elements;
+    private static final double GROW_FACTOR = 1.5;
+    private T[] elements;
     private int size;
 
     public ArrayList() {
-        this.elements = new Object[DEFAULT_CAPACITY];
-        this.size = 0;
-    }
-
-    private void ensureCapacity(int minCapacity) {
-        int oldCapacity = elements.length;
-        if (minCapacity > oldCapacity) {
-            int newCapacity = (int) (oldCapacity * 1.5);
-            if (newCapacity < minCapacity) {
-                newCapacity = minCapacity;
-            }
-            elements = Arrays.copyOf(elements, newCapacity);
-        }
+        this.elements = (T[]) new Object[DEFAULT_CAPACITY];
     }
 
     @Override
@@ -33,14 +21,12 @@ public class ArrayList<T> implements List<T> {
     @Override
     public void add(T value, int index) {
         if (index < 0 || index > size) {
-            throw new ArrayListIndexOutOfBoundsException("Incorrect index " + index);
+            throw new ArrayListIndexOutOfBoundsException("Index out of range " + index
+                    + ", size is " + size);
         }
 
         ensureCapacity(size + 1);
-        for (int i = size; i > index; i--) {
-            elements[i] = elements[i - 1];
-        }
-
+        System.arraycopy(elements, index, elements, index + 1, size - index);
         elements[index] = value;
         size++;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
okuzan, line 16: place ur private methods after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
```suggestion
    private T[] elements;
```
that's exactly where the type parameter should be used :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
okuzan, line 13: we need this to distinguish between a field and an argument that have the same name, which is not the case here. Also, 0 is default for int

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        int oldCapacity = elements.length;
+        if (minCapacity > oldCapacity) {
+            int newCapacity = (int) (oldCapacity * 1.5);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
```
okuzan, line 23: we can only use System.arraycopy to work with arrays

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        }
+
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of range " + index);
+        }
```
okuzan, line 68: duplicate, extract to a private utility method, include size in the message too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        size--;
+        return removed;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && elements[i] == null)
+                    || (element != null && element.equals(elements[i]))) {
```
okuzan, line 93: redundant parentheses , 
```suggestion
            if ((element ==  elements[i])
                    || (element != null && element.equals(elements[i]))) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of range " + index);
+        }
+
+        final T removed = (T) elements[index];
+        for (int i = index; i < size - 1; i++) {
+            elements[i] = elements[i + 1];
+        }
```
okuzan, line 82: use system arraycopy to move elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+
+        elements[size - 1] = null;
+        size--;
```
```suggestion
        elements[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Incorrect index " + index);
+        }
+
+        ensureCapacity(size + 1);
+        for (int i = size; i > index; i--) {
+            elements[i] = elements[i - 1];
+        }
```
okuzan, line 42: same here

olekskov (approved):

PR #1229, opened 2023-11-24
Jv hw array list

Nonzeq:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static int CAPACITY = 10;
```
sokimaaa, line 6: Shouldnt be static
```suggestion
    private static int CAPACITY = 10;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,145 @@
...
+    private int elementsCount;
+    private T[] elements;
+
+    @SuppressWarnings("unchecked")
```
sokimaaa, line 10: never use suppress warning

sokimaaa (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Nonzeq: why it Shouldn't be static?

sokimaaa: static applies to all instances of class.
So if you change static in one class, it will affect others.

Make it or constant, e.g. private static final or plain field depends on context

sokimaaa (approved):

src/main/java/core/basesyntax/ArrayList.java
Nonzeq: ok i'll mark it. thx

PR #1237, opened 2023-12-08
implemented methods from List interface

petro2061:
In this task implements methods from interface List for collection ArrayList. Added special methods for check index and resize inside array.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        changeElementsSize();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        changeElementsSize();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
```
okuzan, line 13: redundant `this`s
0 is default for int

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
```suggestion
    private T[] elements;
```
here we should use our type parameter

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index must be a positive number "
+                    + "and less than the number of elements in the collection: "
+                    + index);
+        }
+    }
+
+    private void changeElementsSize() {
+        if (size == elements.length) {
+            int newArraySize = (int) (elements.length * 1.5);
```
okuzan, line 98:
```suggestion
            int newArraySize = (int) (elements.length * GROW_FACTOR);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+            Object[] newDefaultArray = new Object [newArraySize];
+            System.arraycopy(elements, 0, newDefaultArray,0, size);
```
okuzan, line 100:
```suggestion
            System.arraycopy(elements, 0, newDefaultArray, 0, size);
```
use IDEA autoformatter Ctrl Alt L

okuzan (changes requested):
Good job! Minor fixes left

okuzan (approved):
😎💪

PR #1242, opened 2023-12-13
Implementation of methods from the List<T> interface to the ArrayList<T> class

nazariykolesnikoww:


Rommelua (changes requested):
Good job

Rommelua (approved):
You missed my point a bit)
Approved

PR #1251, opened 2023-12-26
Created my own ArrayList

RCherniavskiy:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    private static final int DEFAULT_VALUE = 10;
+    private static final double MORE_SIZE = 1.5;
+
+    private Object[] array;
+
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_VALUE];
+        size = 0;
```
olekskov, line 15: 0 is default value for int, this line makes no difference
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    }
+
+    private void ensureCapacity() {
```
olekskov, line 18: private methods should be placed after public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        if (index != size) {
+            checkIndex(index);
+        }
+        ensureCapacity();
+        size++;
+        ensureCapacity();
 
+        for (int i = size - 1; i > index; i--) {
+            array[i] = array[i - 1];
+        }
```
olekskov, line 53: Please replace with predefined method according to checklist:

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index != size) {
+            checkIndex(index);
+        }
+        ensureCapacity();
+        size++;
+        ensureCapacity();
```
olekskov, line 49: Please rework with single call of ensureCapacity();

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity();
+        for (int i = 0; i < list.size(); i++) {
+            array[size++] = list.get(i);
+            ensureCapacity();
+        }
```
olekskov, line 63: just call your add() method in loop, it already has all required checks and logic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_VALUE = 10;
+    private static final double MORE_SIZE = 1.5;
+
+    private Object[] array;
```
olekskov, line 9:
```suggestion
    private T[] array;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+
+    private int size;
+
+    public ArrayList() {
+        array = new Object[DEFAULT_VALUE];
```
olekskov, line 14: in such way you won't need to cast objects to T in methods
```suggestion
        array = (T[]) new Object[DEFAULT_VALUE];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        final T removedElement = (T) array[index];
+        for (int i = 0; i < tempArray.length; i++) {
+            tempArray[i] = array[index + 1 + i];
+        }
+        for (int i = index; i < size - 1; i++) {
+            array[i] = array[i + 1];
+        }
+        for (int i = 0; i < tempArray.length; i++) {
+            array[index + i] = tempArray[i];
+        }
```
olekskov, line 91: simplify with single call of System.arraycopy()

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+            for (int i = 0; i < size; i++) {
+                newArray[i] = array[i];
+            }
+            array = newArray;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 99: would be a good idea to include the size of the collection too! If you will have a message invalid index: 5 , current size: 4  
you will immediately know you are off by one position - makes debugging easier

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (size == array.length) {
+            int extension = (int) (array.length * MORE_SIZE);
+            T [] newArray = (T[]) new Object[extension];
+            for (int i = 0; i < size; i++) {
+                newArray[i] = array[i];
+            }
```
okuzan, line 92: use system.arraycopy to move elements around

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Object[] tempArray = new Object[size - index - 1];
+        final T removedElement = array[index];
+        System.arraycopy(array, index + 1, tempArray, 0, tempArray.length);
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
+        System.arraycopy(tempArray, 0, array, index, tempArray.length);
```
okuzan, line 61: one system arraycopy is enough for each method, hint: you can use the same array as both src and dest

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_VALUE];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        array[size] = value;
+        size++;
```
okuzan, line 21:
```suggestion
        array[size++] = value;
```
not a mistake but FYI

okuzan (changes requested):
Good job so far!

okuzan (approved):
Great job!

PR #1253, opened 2024-01-02
Solution

suprun-anna:
created methods for ArrayList (add, remove, size, get, set, isEmpty, addAll)

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int CAPACITY = 10;
+    private static final int START_SIZE = 0;
+    private static final double GROWTH = 1.5;
+
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[CAPACITY];
+        size = START_SIZE;
+    }
+
     @Override
     public void add(T value) {
-
+        growIfArrayFull();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    private static final int START_SIZE = 0;
+    private static final double GROWTH = 1.5;
+
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = new Object[CAPACITY];
+        size = START_SIZE;
+    }
```
olekskov, line 16: 1) 0 is default int value, not need to explicitly set it
2) Better to use parametrized type for array in order to avoid type casting in code
```suggestion
    private static final int CAPACITY = 10;
    private static final double GROWTH = 1.5;

    private T[] elementData;
    private int size;

    public ArrayList() {
        elementData = (T[]) new Object[CAPACITY];
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " out of bounds");
+        }
+        growIfArrayFull();
+        System.arraycopy(elementData, index, elementData, index + 1, size - index);
+        size++;
+        elementData[index] = value;
```
olekskov, line 33: Better to slighlty reoder commands to reuse existing methods, it makes code more simple to understand
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkForIndexOutOfBounds(index);
        growIfArrayFull();
        System.arraycopy(elementData, index, elementData, index + 1, size - index);
        size++;
        elementData[index] = value;
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #1258, opened 2024-01-15
added solution to the task

yanna-stepanova:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+            values = grow(values.length);
+        }
+        T[] tempArray = (T[]) new Object[values.length];
+        for (int i = 0; i < index; i++) {
+            tempArray[i] = values[i];
+        }
+        tempArray[index] = backEmptyForNull(value);
+        for (int i = index + 1; i < newSize; i++) {
+            tempArray[i] = values[i - 1];
+        }
```
olekskov, line 40: Use approach according to checklist:

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,141 @@
...
+        T[] newValues = (T[]) new Object[newCapacity];
+        for (int i = 0; i < values.length; i++) {
+            newValues[i] = values[i];
+        }
+        return newValues;
+    }
+
+    private T backEmptyForNull(T value) {
+        return ((value == null) ? (T) EMPTY_ELEMENTDATA : value);
+    }
```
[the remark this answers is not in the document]
yanna-stepanova: Without this substitute value with null isn't written to my array list. It skips it. And method "get" throws NullPointException, when it makes "equals" with 'null'.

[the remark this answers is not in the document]
kozhukhovsky: add check for null value before equals in "get" method

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        T foundValue = null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && get(i) == null)
+                    || ((element != null && get(i) != null) && get(i).equals(element))) {
```
```suggestion
            if (element == values[i] || element != null && element.equals(values[i])) {
```
we can do it simpler, as in Objects.equals impl

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        size--;
+        return oldValue;
     }
 
     @Override
-    public T remove(T element) {
-        return null;
+    public T remove(T element) throws NoSuchElementException {
+        T foundValue = null;
+        int index = -1;
```
okuzan, line 82:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((element == null && get(i) == null)
+                    || ((element != null && get(i) != null) && get(i).equals(element))) {
+                index = i;
+                foundValue = remove(index);
+                break;
+            }
+        }
+        checkElement(foundValue, index);
+        return foundValue;
```
```suggestion
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element couldn't be found: " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] grow(int oldCapacity) {
+        int newCapacity = oldCapacity + oldCapacity / 2;
```
okuzan, line 106: we always call grow with the same argument, we can remove it. Let's make a clearly defined GROW_FACTOR constant
```suggestion
    private T[] grow() {
        int newCapacity = (int) (GROW_FACTOR * values.length);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        T[] newValues = (T[]) new Object[newCapacity];
+        System.arraycopy(values, 0, newValues, 0, size);
+        return newValues;
+    }
+
+    private void checkBoundsIndex(int index) {
+        if ((index >= size && size != 0) || index < 0) {
```
okuzan, line 113:
```suggestion
        if (index < 0 || index >= size) {
```

okuzan: write a separate check for add, it will be easier to read, right in the method, since it's only used once

okuzan (changes requested):
You can simplify a lot of your codebase 👇

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        checkBoundsIndex(index);
+        return values[index];
     }
 
     @Override
-    public void set(T value, int index) {
-
+    public void set(T value, int index) throws ArrayListIndexOutOfBoundsException {
```
okuzan, line 53: remove declarations for unchecked exceptions everywhere

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    public ArrayList() {
+        values = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= values.length) {
+            values = grow();
+        }
```
```suggestion
        growIfNeeded();
```
let's move this logic to the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        values[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == 0 && index == size) {
+            add(value);
+        } else {
+            checkBoundsIndex((index == size && size != 0) ? (index - 1) : index);
```
okuzan, line 28: follow IDEA's suggestions  - size != 0 is always true when reached, remove it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        }
+        values[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == 0 && index == size) {
+            add(value);
+        } else {
```
okuzan, line 27: let's call return after add - this way we won't need to wrap the rest of the logic in `else` - will be more readable, it's a [known pattern](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)

okuzan (changes requested):
Good job, much cleaner code

okuzan (approved):
Well done!

PR #1263, opened 2024-01-20
Created `ArrayList` class which implements the `List` interface and o…

Ovikt0r:
…verrided the methods of the interface.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int BIT_DIVIDER = 1;
+    private static final int SOURCE_POSITION = 0;
+    private static final int DESTINATION_POSITION = 0;
+    private static final int SHIFT_INDEX_NUMBER = 1;
+    private static final int SIZE_INITIALIZER = 0;
+    private int size;
+    private Object[] elements;
+
+    public ArrayList() {
+        size = SIZE_INITIALIZER;
+        elements = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfArrayIsFull();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndex(index, index > size());
+        growIfArrayIsFull();
+        Object[] elements = this.elements;
+        System.arraycopy(elements, index, elements, index + SHIFT_INDEX_NUMBER, size - index);
+        elements[index] = value;
+        size++;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int BIT_DIVIDER = 1;
+    private static final int SOURCE_POSITION = 0;
+    private static final int DESTINATION_POSITION = 0;
+    private static final int SHIFT_INDEX_NUMBER = 1;
+    private static final int SIZE_INITIALIZER = 0;
+    private int size;
+    private Object[] elements;
```
Ivan95kos, line 13: You need to use the parametrize type
```suggestion
    private T [] elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index, index >= size());
+        growIfArrayIsFull();
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
```
Ivan95kos, line 59: The method is overloaded, try to do it simpler
like:
```
    @Override
    public T remove(int index) {
        checkIndex(index);
        T removedElement = elements[index];
        size--;
        return removedElement;
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        System.arraycopy(elements, index + SHIFT_INDEX_NUMBER,
+                newElements, index, newElements.length - index);
+        T removedItem = (T) elements[index];
+        elements = newElements;
+        size--;
+        return removedItem;
     }
 
     @Override
     public T remove(T element) {
```
Ivan95kos, line 72: the same:
```
    @Override
    public T remove(T element) {
            if ((element == null && elements[i] == null)
                    || (element != null && element.equals(elements[i]))) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element not found" + element);
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        growForAddList(list.size());
+        for (int i = 0; i < list.size(); i++) {
+            elements[size + i] = list.get(i);
+        }
+        size += list.size();
```
Ivan95kos, line 42: use the `add` method
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

Ivan95kos (changes requested):
left a few comments

kozhukhovsky (approved):

PR #1272, opened 2024-01-25
Created own ArrayList

moskalenkosergii:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] DEFAULTCAPACITY_EMPTY_ELEMENTDATA = {};
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = (T[]) DEFAULTCAPACITY_EMPTY_ELEMENTDATA;
+    }
```
olekskov, line 13: There is no need to create zero-sized array in the beginning
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
    private T[] elementData;
    private int size;

    public ArrayList() {
        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+            T[] clonableObject = (T[]) new Object[newCapacity];
+
+            if (oldCapacity > 0 || elementData != DEFAULTCAPACITY_EMPTY_ELEMENTDATA) {
+                for (int i = 0; i < elementData.length; i++) {
+                    clonableObject[i] = elementData[i];
+                }
+                elementData = clonableObject;
+            } else {
+                elementData = (T[]) new Object[Math.max(DEFAULT_CAPACITY, minCapacity)];
+            }
```
olekskov, line 125: as a result we may slightly simplify resize method as out array cannot have 0 size

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The specified index is"
+                    + " larger than the array size");
+        }
```
olekskov, line 50: Please extract repeated logic into separate method (check lines 56-59 and 24-27)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+                remove(i);
+                elementFound = true;
+                break;
+            }
+        }
+
+        if (!elementFound) {
+            throw new NoSuchElementException("Element with such data was not found");
+        }
+        return element;
```
olekskov, line 98: 1) there is null-safe comparison pattern, no need of if/else for it
2) we may use return from if statement to avoid additional if and variable
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == elementData[i] || element != null && element.equals(elementData[i])) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element with such data was not found");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("The specified index is"
+                    + " larger than the array size");
+        }
+        resizeIfNeeded();
 
+        for (int i = size; i > index; i--) {
+            elementData[i] = elementData[i - 1];
+        }
```
olekskov, line 32: checklist item: https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#use-systemarraycopy-to-move-your-array-elements

Please recheck same occurences in other methods

olekskov (changes requested):

kozhukhovsky (approved):

PR #1273, opened 2024-01-26
HW ArrayList

Cliffcamer:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+
+    private Object[] array;
```
okuzan, line 11: That's where we should use class type parameter T

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.array = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
okuzan, line 16:
```suggestion
        array = new Object[DEFAULT_CAPACITY];
```
0 is a default value for int
let's not use `this` where we don't need to - https://stackoverflow.com/a/2411283

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
+        array[--size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(element, array[i])) {
```
okuzan, line 68: image
[Checklist](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (size == array.length) {
+            int newCapacity = (int) (array.length * GROWTH_FACTOR);
+            array = Arrays.copyOf(array, newCapacity);
```
okuzan, line 88: https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 94: let's include size in the message too to further facilitate debugging

okuzan (changes requested):
Good job! Clean code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+            array = newArray;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(index + " not within the size.");
```
okuzan, line 99:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Invalid index " + index + " for size " + size);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    }
+
+    private void ensureCapacity() {
+        if (size == array.length) {
+            int newCapacity = (int) (array.length * GROWTH_FACTOR);
+            Object[] newArray = new Object[newCapacity];
+            for (int i = 0; i < size; i++) {
+                newArray[i] = array[i];
+            }
+            array = newArray;
```
okuzan, line 90: use system.arraycopy to move elements

src/main/java/core/basesyntax/ArrayList.java
okuzan: not fixed, use T, you will need to add some casting too (in constructor)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        System.arraycopy(array, index + 1, array, index, size - index - 1);
+        array[--size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == null && array[i] == null)
```
okuzan, line 65:
```suggestion
            if ((element == array[i])
```

okuzan (changes requested):

Rommelua (approved):

PR #1277, opened 2024-01-30
implemented custom ArrayList class with required methods

DanilKi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list != null) {
+            for (int i = 0; i < list.size(); i++) {
+                add(list.get(i));
+            }
+        }
```
okuzan, line 42:
```suggestion
            for (int i = 0; i < list.size(); i++) {
                add(list.get(i));
            }
```
not consistent with real arrayList behaviour, it should just throw NPE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     @Override
+    @SuppressWarnings("unchecked")
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedElement = (T) elementData[index];
+        int numberOfElementsToMove = size - index - 1;
+        if (numberOfElementsToMove > 0) {
+            System.arraycopy(elementData, index + 1, elementData, index, numberOfElementsToMove);
+        }
```
```suggestion
          System.arraycopy(elementData, index + 1, elementData, index, numberOfElementsToMove);
```
it won't do anything in that case anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        if (element != null) {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    return remove(i);
+                }
+            }
+        } else {
```
okuzan, line 77: remove external if-else, combine two bodies into one, you need to write a correct equals condition in internal `if` - look at Objects.equals code for reference

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            increaseCapacity();
+        }
```
okuzan, line 19: move if check inside of the method

okuzan (changes requested):
Good job!

olekskov (approved):

PR #1278, opened 2024-01-31
The task has done bu the "README" instructions

NekitCool1:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static int DEFAULT_SIZE = 10;
+    private int size;
+    private T[] elements = (T[]) new Object[10];
```
kerrrusha, line 9: initialize this in the constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    private void ensureCapacity(int minCapacity) {
```
kerrrusha, line 11: 1. private methods should be located after public methods
2. this name better explains what does this method do
3. we dont need any parameters for that
```suggestion
        private void growArray() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        int currentCapacity = elements.length;
+        if (minCapacity > currentCapacity) {
+            int newCapacity = currentCapacity + (currentCapacity >> 1);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
+            elements = Arrays.copyOf(elements, newCapacity);
+        }
```
kerrrusha, line 19: it's better to avoid unnecessary use of bitwise operations. use constants instead
```suggestion
        int newCapacity = (int) (elementData.length * RESIZE_FACTOR);
        if (size == elementData.length) {
        ... do resize
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
 
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "The size of the list less than your index");
+        }
```
kerrrusha, line 34: extract this to the method (violates SRP and DRY principles)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            ensureCapacity(size + 1);
+            elements[size++] = list.get(i);
```
kerrrusha, line 45: extract this to the method (violates SRP and DRY principles)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
kerrrusha, line 53: extract this to the method (violates SRP and DRY principles)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
kerrrusha, line 61: extract this to the method (violates SRP and DRY principles)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
kerrrusha, line 69: extract this to the method (violates SRP and DRY principles)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    removedElement = elements[i];
+                    return remove(i);
+                }
+            }
+        }
```
kerrrusha, line 95:
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == null ? elementData[i] == null : element.equals(elementData[i])) {
                ...do necessary logic...
            }
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        if (removedElement == null) {
+            throw new NoSuchElementException("No such element");
+        }
+        return remove(-1);
```
kerrrusha, line 99: if program runs here, it means that no element was removed, so just throw exception here
```suggestion
        throw new NoSuchElementException("No such element: " + element);
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static int DEFAULT_SIZE;
+    private static int RESIZE_FACTOR;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        DEFAULT_SIZE = 10;
+        RESIZE_FACTOR = 5;
```
okuzan, line 15: only non-static fields of the class should be initialized in a constructor, make these final, init at declaration
resize factor 5 implies you are growing it x5 times

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return size == 0;
+    }
+
+    private void growArray() {
+        int currentCapacity = elements.length;
+        if (size + 1 > currentCapacity) {
+            int newCapacity = currentCapacity + (int) (currentCapacity * RESIZE_FACTOR);
+            if (newCapacity < size + 1) {
+                newCapacity = size + 1;
+            }
```
okuzan, line 96: use system.arraycopy to move elements around, as mentioned in the checklist

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growArray() {
+        int currentCapacity = elements.length;
+        if (size + 1 > currentCapacity) {
```
```suggestion
        if (size >= currentCapacity) {
```
when tempted to use +1 in comparisons, you can usually fix it with adding/removing equals

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        exceptionCalling(index);
+        T oldValue = elements[index];
+        int numMoved = size - index - 1;
+        if (numMoved > 0) {
+            System.arraycopy(elements, index + 1, elements, index, numMoved);
+        }
```
okuzan, line 64: this if wrap is unnecessary, nothing will happen if it isn't true anyway

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            growArray();
+            elements[size++] = list.get(i);
```
okuzan, line 41: you are repeating the logic of `add`
just call add here

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private final int defaultSize;
+    private final double resizeFactor;
+    private int size;
+    private T[] elements;
+
+    public ArrayList() {
+        defaultSize = 10;
+        resizeFactor = 0.5;
```
kerrrusha, line 14: according to this comment it was supposed to do this changes:
image

```suggestion
    private static final int DEFAULT_SIZE = 10;
    private static final double RESIZE_FACTOR = 0.5;
    private int size;
    private T[] elements;

    public ArrayList() {
```

src/main/java/core/basesyntax/ArrayList.java
kerrrusha: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "The size of the list less than your index");
+        }
```
kerrrusha, line 29:
```suggestion
        checkIndexForAdd(index);
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        int currentCapacity = elements.length;
+        if (size >= currentCapacity) {
+            int newCapacity = currentCapacity + (int) (currentCapacity * RESIZE_FACTOR);
+            T[] array = (T[])new Object[newCapacity];
+            System.arraycopy(elements, 0, array, 0, currentCapacity);
+            elements = array;
+        }
+    }
+
+    private void exceptionChecking(int index) {
```
kerrrusha, line 91:
```suggestion
    private void checkIndex(int index) {
```

kerrrusha (approved):

PR #1279, opened 2024-02-01
task solution

trokhim03:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
+        T removeElement = array[index];
+        int startIndex = index;
+        int elementsToMove = size - index - 1;
+        System.arraycopy(array, startIndex + 1, array, startIndex, elementsToMove);
+        size--;
+        return removeElement;
     }
 
     @Override
     public T remove(T element) {
```
okuzan, line 71: i believe you can implement it a more concise way :) looks like code duplication
remove external if-else for starters
reuse the existing method with this exact logic
image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (size == array.length) {
+            array = grow();
+        }
+        if (index - 1 > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Object index cannot be "
+                    + "larger than the list size");
+        }
```
okuzan, line 33:
```suggestion
        if (index > size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Object index cannot be "
                    + "larger than the list size");
        }
```

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,126 @@
...
+    public ArrayList() {
+        array = (T[])new Object[DEFAULT_CAPACITY];
+
+    }
+
     @Override
     public void add(T value) {
+        if (size == array.length) {
+            array = grow();
+        }
```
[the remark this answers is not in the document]
kerrrusha: not fixed. if you need any help - feel free to ask mentors at the Mate platform chat

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        array[size++] = value;
 
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == array.length) {
+            array = grow();
+        }
```
kerrrusha, line 28: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+
+    private boolean checkBound(int index) {
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index can't be negative :" + index);
+        }
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Bad input index: " + index
+                    + "for size: " + size);
+        }
+        return true;
```
kerrrusha, line 111: you never use the return value, moreover, if some check method returns boolean, it's not check method, it's validate method
```suggestion
    private void checkBound(int index) {
        if (index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Index can't be negative :" + index);
        }
        if (index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Bad input index: " + index
                    + "for size: " + size);
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (size == array.length) {
+            array = grow();
+        }
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Object index cannot be "
+                    + "larger than the list size");
+        }
```
kerrrusha, line 32: should be separated
```suggestion
        checkBoundForAdd(index);
```

kerrrusha (changes requested):

kerrrusha (approved):

PR #1280, opened 2024-02-01
Develop

oKukHtina:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return (size == 0);
     }
+
+    private void checkIndexPosition(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("");
```
olekskov, line 89: Please add meaningful message to all thrown exceptions

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+            T currentElement = values[i];
+            if (currentElement == element
+                    || currentElement != null
+                    && currentElement.equals(element)) {
+                incomeElement = element;
+                deleteElementInArray(i);
+                return incomeElement;
+            }
+        }
+        throw new NoSuchElementException("");
```
olekskov, line 74: Here you can reuse more top-level method
```suggestion
        for (int i = 0; i < values.length; i++) {
            if (values[i] == element || values[i] != null && values[i].equals(element)) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("..some message..");
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
-        return null;
+        T incomeElement;
+        for (int i = 0; i < values.length; i++) {
+            T currentElement = values[i];
+            if (compareElementsFromArray(currentElement, element)) {
+                incomeElement = element;
+                deleteElementInArray(i);
+                return incomeElement;
+            }
+        }
```
olekskov, line 71: Please reuse remove(int index) method here as was shown on previous review

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        values = newArray;
+    }
+
+    private void deleteElementInArray(int index) {
+        System.arraycopy(values, index + 1, values, index,
+                values.length - index - 1);
+        size--;
+    }
+
+    private boolean compareElementsFromArray(T currentElement, T element) {
```
olekskov, line 105: methods that return boolean are usually more readable with is/are in name
```suggestion
    private boolean areEqualElements(T currentElement, T element) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        int newLength = currentLength + (currentLength >> 1);
+        T[] newArray = (T[]) new Object[newLength];
+        System.arraycopy(values, 0, newArray, 0, values.length);
+        values = newArray;
+    }
+
+    private void deleteElementInArray(int index) {
+        System.arraycopy(values, index + 1, values, index,
+                values.length - index - 1);
+        size--;
```
olekskov, line 102: after refactoring according to the first comment it would make sense to move this logic to remove(int index) method

olekskov (changes requested):
Good job, minor comments left

olekskov (changes requested):
`

olekskov (approved):
Good job

PR #1283, opened 2024-02-06
implemented custom ArrayList

ghost:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double DEFAULT_CAPACITY_MULTIPLIER = 1.5;
+
+    private T[] container;
+    private int size;
+
+    public ArrayList() {
+        container = (T[])new Object[DEFAULT_CAPACITY];
+        size = 0;
```
kerrrusha, line 14: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        container[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index value while adding");
+        }
```
kerrrusha, line 28: extract to the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+
+    public ArrayList() {
+        container = (T[])new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        growIfNecessary(1);
```
kerrrusha, line 19:
```suggestion
        growArray();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index value");
+        }
+    }
+
+    private void growIfNecessary(int elementsToBeAdded) {
+        while (size + elementsToBeAdded > container.length) {
+            grow();
+        }
+    }
```
kerrrusha, line 105: we dont need this
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        T []copy = (T[])new Object[(int)(container.length * DEFAULT_CAPACITY_MULTIPLIER)];
+        for (int i = 0; i < container.length; i++) {
+            copy[i] = container[i];
+        }
+        container = copy;
```
kerrrusha, line 92: rewrite this logic in next way:
1. calculate new capacity based on current array length and RESIZE_FACTOR
2. if array is full - resize it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double DEFAULT_CAPACITY_MULTIPLIER = 1.5;
+
+    private T[] container;
+    private int size;
+
+    public ArrayList() {
+        container = (T[])new Object[DEFAULT_CAPACITY];
```
kerrrusha, line 13:
```suggestion
        container = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        growIfNecessary(list.size());
+        for (int i = 0; i < list.size(); i++) {
+            container[i + size] = list.get(i);
+        }
+        size += list.size();
```
kerrrusha, line 39: just iterate through elements and call `add` method on them

koval-kn (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+    public ArrayList() {
+        container = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growArray(size + 1);
+        container[size] = value;
+        size++;
```
```suggestion
        container[size++] = value;
```
not a mistake but FYI

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        for (int i = 0; i < size; i++) {
+            if (elementsAreEqual(element, container[i])) {
+                index = i;
+                break;
+            }
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("Can't find element while removing");
+        }
+        return remove(index);
```
okuzan, line 69: rewrite this without using index local var, second if and break
in such  cases, cleaner code would be returning inside of loop, throwing outside of loop (without if, because if that code is getting executed, it means we didn't find the match)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+
+    private void growArray(int minCapacity) {
+        if (minCapacity <= container.length) {
+            return;
+        }
+        T []copy = (T[]) new Object[(int)(container.length * RESIZE_FACTOR)];
+        for (int i = 0; i < container.length; i++) {
+            copy[i] = container[i];
+        }
+        container = copy;
```
okuzan, line 90: use system.arraycopy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+
+    private void validateIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index value");
+        }
+    }
+
+    private void validateIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index value while adding");
```
okuzan, line 101: include index and size in the messag

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        T[] copy = (T[])new Object[container.length];
+
+        for (int i = 0; i < index; i++) {
+            copy[i] = container[i];
+        }
+        copy[index] = value;
+        for (int i = index; i < size; i++) {
+            copy[i + 1] = container[i];
+        }
+        container = copy;
```
okuzan, line 115: use system arraycopy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+    }
+
+    private void validateIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index value while adding");
+        }
+    }
+
+    private void insertElement(T value, int index) {
+        T[] copy = (T[])new Object[container.length];
```
okuzan, line 106:
```suggestion
        T[] copy = (T[]) new Object[container.length];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            copy[i + 1] = container[i];
+        }
+        container = copy;
+        size++;
+    }
+
+    private void removeElement(int index) {
+        for (int i = index; i < size - 1; i++) {
+            container[i] = container[i + 1];
+        }
```
okuzan, line 122: system arraycopy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            return true;
+        }
+        if (el1 == null || el2 == null) {
+            return false;
+        }
+        return el1.equals(el2);
     }
 }
+
+
```
okuzan, line 137: you have three empty lines at the end of this file, you should always have exactly one

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
+    private int size;
+
+    public ArrayList() {
+        container = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growArray(size + 1);
```
olekskov, line 18: You call method only with size+1 agrument. Better just include this value to method logic
```suggestion
        growIfRequired();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,122 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void growArray(int minCapacity) {
+        if (minCapacity <= container.length) {
```
olekskov, line 77:
```suggestion
        if (size < container.length) {
```

olekskov (approved):
Good job

PR #1284, opened 2024-02-07
jv-array-list

OleksiiBashkirov:
created ArrayList
added checkCapacity method

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_INDEX = 1.5;
+
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        }
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
```
olekskov, line 61: Please extract repeated logic to separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
olekskov, line 82: Please compare null and not null objects in one if
        // your logic
    }
}

olekskov (changes requested):

OleksiiBashkirov:
extracted repetitive logic into the checkIndex method
also created checkIndexForAdd method 
modified public T remove(T element) method with a single if

olekskov (approved):

PR #1285, opened 2024-02-07
array-list-resolved

margosalii:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (arrayList[i] == element || (arrayList[i] != null && arrayList[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("There is no " + element + " to remove");
+    }
+
+    public void checkIndexOutOfBoundsException(int index) {
```
olekskov, line 81:
```suggestion
    public void checkIndexOutOfBounds(int index) {
```

olekskov (approved):
Good job, approved.
No need to add Exception word to methods, we use it only for classes names

PR #1286, opened 2024-02-08
method was implemented

Artem-Kushniruk:


pom.xml
```diff
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>17</jdk.version>
+        <jdk.version>11</jdk.version>
```
okuzan, line 12: why change?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementList.length) {
+            elementList = grow();
+        }
+        elementList[size] = value;
+        size++;
```
```suggestion
        elementList[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
+    private Object[] elementList;
+    private int size;
+
+    public ArrayList() {
+        this.elementList = EMPTY_ELEMENT_DATA;
```
```suggestion
        elementList = EMPTY_ELEMENT_DATA;
```
https://stackoverflow.com/a/2411283

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
+    private Object[] elementList;
```
okuzan, line 9: You have class type parameter, why use object

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+    public ArrayList() {
+        this.elementList = EMPTY_ELEMENT_DATA;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementList.length) {
+            elementList = grow();
+        }
```
```suggestion
        growIfNeeded();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+        size += listAsArray.length;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index > size - 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can not get element, "
+                    + "index is out of bounds.");
+        }
```
okuzan, line 54:
```suggestion
        if (index >= size || index < 0) {
            throw new ArrayListIndexOutOfBoundsException("Can not get element, "
                    + "index is out of bounds.");
        }
```
make this a private utility method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
-        return null;
+        if (index > size - 1 || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can not remove element, "
+                    + "index is out of bounds.");
+        }
+        int newSize = size - 1;
+        final Object oldValue = elementList[index];
+        if (newSize > index) {
+            System.arraycopy(elementList, index + 1, elementList, index, newSize - index);
+        }
```
okuzan, line 77: if is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if ((element == elementList[i])
+                    || (element != null && element.equals(elementList[i]))) {
+                index = i;
+                break;
+            }
```
okuzan, line 91: call remove by index here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+        }
+        if (index == -1) {
+            throw new NoSuchElementException("Element not found");
+        }
```
okuzan, line 95: throw exception unconditionally here
Returning early is usually a [preferred approach](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement) in SE

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,135 @@
...
+        Object[] elementData = new Object[size];
+        System.arraycopy(elementList, 0, elementData, 0, size);
+        return elementData;
+    }
+
+    private Object[] grow(int minCapacity) {
+        int oldCapacity = elementList.length;
+        if (elementList != EMPTY_ELEMENT_DATA) {
+            int minGrowth = minCapacity - oldCapacity;
+            int defaultGrowth = oldCapacity >> 1;
```
okuzan, line 122: avoid bitwise operations, gains are insignificant, readability is more important, create a GROW_FACTOR constant

okuzan (changes requested):

pom.xml
Artem-Kushniruk: Because Maven can not build the project with default(SDK openjdk-21 and jdk version 17), so i need to change SDK to 11th version and version of jdk in POM.xml to 11th respectively. Only with such settings maven successfully builds the project

src/main/java/core/basesyntax/ArrayList.java
Artem-Kushniruk: As i've read through the internet, JAVA doesn't support generic arrays.Anyway i've some examples like https://www.baeldung.com/java-generic-array, but for me using Object is better for code readability, also i noticed that ArrayList "default" method also uses Object.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int GROW_FACTOR = 2;
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
+    private Object[] elementList;
```
okuzan, line 10: Use T instead of object here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((element == elementList[i])
+                    || (element != null && element.equals(elementList[i]))) {
+                Object oldValue = elementList[i];
+                remove(i);
+                return (T) oldValue;
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
okuzan, line 78: include element in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        if (size == elementList.length) {
+            return grow(size + 1);
+        }
+        return elementList;
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Can not remove element, "
+                    + "index is out of bounds.");
```
okuzan, line 122: include index and size in the message

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
okuzan: we can't create generic array using new, but we can declare it, assign array of objects to it and cast

pom.xml
okuzan: please adjust JAVA_HOME, and settings on Project Settings to work with Java 17, it may be a problem later on in the course

okuzan (approved):
Well done

PR #1287, opened 2024-02-09
"task is done"

MarynaBodina:
let me know if something's wrong.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,98 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double RESIZE_FACTOR = 1.5;
+
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        this.array = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            resize();
+        }
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        this.array = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == array.length) {
+            resize();
+        }
```
olekskov, line 21: You call resize method only in combination with if block. It's better to put this if inside method
```suggestion
        resizeIfRerquired();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+        if (size == array.length) {
+            resize();
+        }
```
olekskov, line 32: better to reuse existing methods, it will make code more easy to read
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double RESIZE_FACTOR = 1.5;
+
+    private Object[] array;
+    private int size;
+
+    public ArrayList() {
+        this.array = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
```
olekskov, line 15: 1) Better to parametrize array field properly in order to not type-cast in methods code
2) 0 is the default int value, no need to explicitly set it
```suggestion
    private T[] array;
    private int size;

    public ArrayList() {
        this.array = (T[]) new Object[DEFAULT_CAPACITY];
    }
```

olekskov (changes requested):
Good job, just some style comments left

PR #1288, opened 2024-02-09
Created a custom ArrayList class

IgorGunder:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
kerrrusha, line 12: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+                    return i;
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
```
kerrrusha, line 125: can be simplified into one loop. think about how

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = indexOf(element);
+        if (index != -1) {
+            return remove(index);
+        } else {
+            throw new NoSuchElementException("Element not found: " + element);
+        }
```
kerrrusha, line 99:
```suggestion
        } 
        throw new NoSuchElementException("Element not found: " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= elements.length) {
+            T[] newElements = (T[]) new Object[elements.length * 2];
```
kerrrusha, line 18: 2 is a magic value

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+            System.arraycopy(elements, 0, newElements, 0, elements.length);
+            elements = newElements;
+        }
```
kerrrusha, line 21: refactor to the separate method
```suggestion
        growArrayIfRequred();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
```
kerrrusha, line 30: refactor to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        size += listSize;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of bounds" + index);
+        }
```
kerrrusha, line 67: refactor to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid array index: " + index);
+        }
```
kerrrusha, line 83: refactor to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        return elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
```
kerrrusha, line 75: refactor to the separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+            elements = newElements;
+        }
 
+        for (int i = 0; i < listSize; i++) {
+            if (size + i >= elements.length) {
+                throw new ArrayListIndexOutOfBoundsException("Invalid array range: " + list);
+            }
+            elements[size + i] = list.get(i);
+        }
+        size += listSize;
```
kerrrusha, line 60: dont invent problems and difficulties in your code :)
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        if (size == elements.length) {
+            T[] newElements = (T[]) new Object[elements.length * 2];
+            System.arraycopy(elements, 0, newElements, 0, elements.length);
+            elements = newElements;
+        }
```
kerrrusha, line 35:
```suggestion
        growArrayIfRequred();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        for (int i = size - 1; i >= index; i--) {
+            elements[i + 1] = elements[i];
+        }
```
kerrrusha, line 38: not implemented:

kerrrusha (changes requested):

src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
```diff
@@ -5,3 +5,4 @@ public ArrayListIndexOutOfBoundsException(String message) {
         super(message);
     }
 }
+
```
okuzan, line 8: when you commit anything, always review the files, to not commit something inadvertently, like you did here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                return i;
+            }
+        }
+
+        return -1;
+    }
+
+    public void checkIndexBounds(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 102: include size in the message too, if you will get exception message like `invalid index 4, size 3` you will immediately know you are off by one, without debugging

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+
+    public int indexOf(T element) {
+        for (int i = 0; i < size; i++) {
+            if (element == null) {
+                if (elements[i] == null) {
+                    return i;
+                }
+            } else if (element.equals(elements[i])) {
+                return i;
+            }
```
okuzan, line 94: look up code for Objects.equals, and write only one correct comparison condition
`a == b || a != null && a.equals(b)`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexBounds(index);
+        T removedElement = elements[index];
+        for (int i = index; i < size - 1; i++) {
+            elements[i] = elements[i + 1];
+        }
```
okuzan, line 62: use system.arraycopy to move elements around instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        growArrayIfRequred();
+        T[] newElements = (T[]) new Object[elements.length + EXPANSION_FACTOR1];
+        System.arraycopy(elements, 0, newElements, 0, index);
+        System.arraycopy(elements, index, newElements, index + EXPANSION_FACTOR1, size - index);
```
okuzan, line 31: only one system. arraycopy call is needed in each method, you don't need that copy, pass the same array as src and dest

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growArrayIfRequred();
+        elements[size] = value;
+        size++;
```
```suggestion
        elements[size++] = value;
```
not a mistake but nice to do in one line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int EXPANSION_FACTOR1 = 1;
+    private static final int EXPANSION_FACTOR2 = 2;
```
```suggestion
    private static final int EXPANSION_FACTOR = 2;
```

src/main/java/core/basesyntax/ArrayList.java
okuzan, line 94: used only once, consider to moving to remove instead of having a separate method
also, all your utility methods should be private, not public

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
IgorGunder: Підскажіть будь ласка як за допомогою лише одного System.arraycopy можна скопіювати елементи з масиву elements до масиву newElements до індексу index і в ньому ж скопіювати елементи з масиву elements після індексу index до масиву newElements із зсувом на одну позицію вліво.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            if (element == null && elements[i] == null
+                    || element != null && element.equals(elements[i])) {
+                index = i;
+                break;
+            }
+        }
+        if (index != -OFFSET) {
+            return remove(index);
+        }
+        throw new NoSuchElementException("Element not found: " + element);
```
Rommelua, line 74:
```suggestion
        for (int i = 0; i < size; i++) {
            if (element == elements[i] 
                    || element != null && element.equals(elements[i])) {
                return remove(i);
            }
        }
        throw new NoSuchElementException("Element not found: " + element);
```

Rommelua (approved):

PR #1289, opened 2024-02-09
Worked with ArrayList

ghost:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY];
```
olekskov, line 13: Better to parametrize array properly to avoid type-casting in methods
```suggestion
    private T[] elementData;
    private int size;

    public ArrayList() {
        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elementData.length) {
```
olekskov, line 98: you always call this method with size+1 argument, so it's possible to work without it
```suggestion
    private void ensureCapacity() {
        if (size >= elementData.length) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+        }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
```
olekskov, line 45: Please extract repeating logic to separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elementData.length) {
+            int newCapacity = (int) (elementData.length * GROWTH_FACTOR);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
```
olekskov, line 102: Excessive check. We add elements 1-by-1 so it's impossible to resize less than required
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
olekskov, line 83: combine into one loop
```
    if(element == elementData[i] || element !=null && element.equals(elementData[i]) {
        ...your logic...
    }
}
```

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
```diff
@@ -4,4 +4,7 @@ public class ArrayListIndexOutOfBoundsException extends RuntimeException {
     public ArrayListIndexOutOfBoundsException(String message) {
         super(message);
     }
+
+    public ArrayListIndexOutOfBoundsException(int index) {
+    }
```
okuzan, line 9: undo this change, what's the point of passing an arg and ignoring it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null ? elements[i] == null : element.equals(elements[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
okuzan, line 83: include element in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+        }
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(index);
+        }
```
okuzan, line 58: not fixed, should be a separate private utility method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+    }
+
+    private void ensureCapacity() {
```
okuzan, line 15: private methods should go after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
okuzan, line 8: To leverage the full benefits of Java's type system and ensure type safety, it's recommended to use the generic type parameter T rather than a non-specific Object array for the internal storage of ArrayList. This approach enables the compiler to check for type compatibility at compile time, preventing potential runtime errors and making the code more maintainable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+    }
+
+    private void ensureCapacity() {
+        if (size == elements.length) {
+            int newCapacity = elements.length + (elements.length >> 1);
```
okuzan, line 17: create a separate GROW_FACTOR constant for better readability than bitwise operations can offer

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+            elements = Arrays.copyOf(elements, newCapacity);
```
okuzan, line 18: we can't use Arrays.copyOf in this task, use System.arraycopy instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(index);
+        }
+        T removedElement = (T) elements[index];
+        int numToMove = size - index - 1;
+        if (numToMove > 0) {
+            System.arraycopy(elements, index + 1, elements, index, numToMove);
+        }
```
okuzan, line 71: this `if` is redundant, doesn't change the flow of the program

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,108 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
Elena-Bruyako, line 8:
```suggestion
    private T[] elements;
```

src/main/java/core/basesyntax/ArrayList.java
Elena-Bruyako: Not fixed

src/main/java/core/basesyntax/ArrayList.java
Elena-Bruyako: Not fixed

src/test/java/core/basesyntax/ArrayListTest.java
```diff
@@ -14,8 +14,10 @@ public void addElementToArrayList() {
         arrayList.add(null);
         arrayList.add("for");
         arrayList.add("Mate");
-        Assert.assertEquals("Test failed! Size of array should be " + 4 + "but it is "
-                + arrayList.size(), 4, arrayList.size());
+        Assert.assertEquals(
```
Elena-Bruyako, line 17: Why did you change this class?

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/ArrayListTest.java
Vasul-Pankiv, line 1: do not change test classes (problem could be in tab symbol) try to revert your changes

Vasul-Pankiv (approved):

PR #1293, opened 2024-02-13
solve the task

esidarapp:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
```suggestion
    private T[] elements;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
okuzan, line 12: should we use `this` here?  https://stackoverflow.com/a/2411283
0 is default for int, this line is useless

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        int oldCapacity = elements.length;
+        if (minCapacity > oldCapacity) {
+            int newCapacity = (oldCapacity * 3 / 2) + 1;
```
okuzan, line 18: create a clearly defined GROW_FACTOR constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            Object[] newElements = new Object[newCapacity];
+            System.arraycopy(elements, 0, newElements, 0, size);
+            elements = newElements;
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity(size + 1);
```
okuzan, line 27: ensureCapacity doesn't need to accept any arguments

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+                    return i;
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return i;
+                }
+            }
+        }
```
okuzan, line 108: merge into one loop
hint: look up code of Objects.equals

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index" + index);
+        }
```
okuzan, line 53: - add size to make debugging easier
- duplicated logic, extract to  a private method. Each condition apart from one in `add` is the same

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
kozhukhovsky: not fixed

kozhukhovsky (approved):

PR #1296, opened 2024-02-15
done

SukhachovS:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            throw new IllegalArgumentException("Array`s capacity cannot be <= 0");
+        } else {
+            array = (T[]) new Object[capacity];
+            this.size = 0;
+        }
+    }
+
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
okuzan, line 22: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        checkCapacity();
+        array[size] = value;
+        size++;
```
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private T[] array;
+
+    public ArrayList(int capacity) {
+        if (capacity <= 0) {
+            throw new IllegalArgumentException("Array`s capacity cannot be <= 0");
+        } else {
```
okuzan, line 14: we don't need else after returning/throwing

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            array = (T[]) new Object[capacity];
+            this.size = 0;
```
okuzan, line 16: remove this line, but remember [where](stackoverflow.com/a/2411283) we should use `this`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        }
+    }
+
+    private void checkCapacity() {
+        if (size == array.length) {
+            int updatedCapacity = (int) (array.length * 1.5);
+            T[] updatedArray = (T[]) new Object[updatedCapacity];
+            for (int i = 0; i < array.length; i++) {
+                updatedArray[i] = array[i];
+            }
```
okuzan, line 111: use system.arraycopy for moving elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index cannot be less then 0");
+        }
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index cannot be >= size");
+        }
+    }
+
+    private void checkCapacity() {
+        if (size == array.length) {
+            int updatedCapacity = (int) (array.length * 1.5);
```
okuzan, line 107: GROW_FACTOR constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index cannot be less then 0");
+        }
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index cannot be >= size");
```
okuzan, line 101: include size in the message, to make debugging easier for users of this class

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        }
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(array[i], element)) {
```
okuzan, line 79: can we use objects in this hw?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        T removedElement = (T) array[index];
+        for (int i = index; i < size - 1; i++) {
+            array[i] = array[i + 1];
+        }
```
okuzan, line 71: system.arraycopy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index cannot be < 0 or > size");
+        }
+        checkCapacity();
+        for (int i = size; i > index; i--) {
+            array[i] = array[i - 1];
+        }
```
okuzan, line 41: same

okuzan (changes requested):
Good job overall, but please always read through the common mistakes checklist before creating a PR/ requesting a review

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Index cannot be < 0 or > size");
+        }
+        checkCapacity();
+        if (size - index >= 0) {
+            System.arraycopy(array, index, array, index + 1, size - index);
+        }
```
okuzan, line 37: this if is redundant, system.arraycopy won't do anything otherwise

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        T removedElement = (T) array[index];
+        if (index < size - 1) {
+            System.arraycopy(array, index + 1, array, index, size - index - 1);
+        }
```
okuzan, line 67: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((array[i] == null && element == null)
+                    || (array[i] != null && array[i].equals(element))) {
+                return remove(i);
```
okuzan, line 77: redundant parentheses, although you can keep them for the sake of better readability
condition can be simplified
```suggestion
            if (array[i] == element
                    || array[i] != null && array[i].equals(element)) {
                return remove(i);
```

okuzan (approved):
well done

PR #1297, opened 2024-02-15
Solved jv-array-list

Shimeriyu:
Recrated ArrayList implementation, all tests passed

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,127 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int DEFAULT_MIN_INCREMENT = 1;
+    private int size;
+    private T[] elementData;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        validateAddIndex(index);
+        if (index == size) {
+            add(value);
+            return;
+        }
+        ensureCapacity();
+        System.arraycopy(elementData, index, elementData, index + 1, size);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(list.size());
+        for (int i = 0; i < list.size(); i++) {
+            /*add() is not used to avoid calling ensureCapacity() in every iteration*/
+            elementData[size] = list.get(i);
+            size++;
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        validateGetRemoveSetIndex(index);
+        return elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
+        validateGetRemoveSetIndex(index);
+        elementData[index] = value;
 
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateGetRemoveSetIndex(index);
+        T removedElement = elementData[index];
+
+        if (index == size - 1) {
+            elementData[index] = null;
+            size--;
+            return removedElement;
+        }
+        System.arraycopy(elementData, index + 1, elementData, index, size);
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] != null && elementData[i].equals(element)
+                    || elementData[i] == element) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found in list");
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow(int neededCapacity) {
+        neededCapacity += elementData.length;
+        int newCapacity = elementData.length;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(list.size());
+        for (int i = 0; i < list.size(); i++) {
+            /*add() is not used to avoid calling ensureCapacity() in every iteration*/
```
nowickimj, line 40: most of the times the execution of ensureCapacity will end with size check - it's better to reuse it than maintaining the redundant code.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        }
+    }
+
+    private void validateAddIndex(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private void validateGetRemoveSetIndex(int index) {
```
nowickimj, line 122: I think `validateIndex` will be enough - you're considering the generic case & the add case.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
     @Override
     public void add(T value, int index) {
-
+        validateAddIndex(index);
+        if (index == size) {
+            add(value);
+            return;
+        }
+        ensureCapacity();
+        System.arraycopy(elementData, index, elementData, index + 1, size);
```
nowickimj, line 31: Incorrect last element - you should copy only the subset of elements, not the whole array.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
-        return null;
+        validateGetRemoveSetIndex(index);
+        T removedElement = elementData[index];
+
+        if (index == size - 1) {
+            elementData[index] = null;
+            size--;
+            return removedElement;
+        }
+        System.arraycopy(elementData, index + 1, elementData, index, size);
```
nowickimj, line 69: Last argument is also invalid.

nowickimj (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
 
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        T removedElement = elementData[index];
+
+        if (index == size - 1) {
```
nowickimj, line 61: The `index == size - 1` case can be removed - arraycopy call is enough here.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+            elementData[index] = null;
+            size--;
+            return removedElement;
+        }
+        System.arraycopy(elementData, index + 1, elementData, index, size - index);
```
nowickimj, line 66: Invalid copy length, you can omit the removed item as well:
```java
elementData[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
nowickimj, line 61: The index == size - 1 case can be removed - arraycopy call is enough here.

nowickimj (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+
+    private void validateAddIndex(int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private void validateIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
```
okuzan, line 101: let's include size and index to the exception message to make debugging for potential users of this class easier, it's a good practice to include relevant details to make exceptions as informative as possible

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow() {
+        int newCapacity = (int)(elementData.length * GROW_RATE);
```
okuzan, line 83:
```suggestion
        int newCapacity = (int) (elementData.length * GROW_RATE);
```
make sure to autoformat your code in IDEA

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        elementData = Arrays.copyOf(elementData, newCapacity);
```
okuzan, line 84: [checklist issue](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#use-systemarraycopy-to-move-your-array-elements) - use System.arraycopy() instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elementData[i] != null && elementData[i].equals(element)
+                    || elementData[i] == element) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found in list");
```
okuzan, line 69: add `element` to the message

okuzan (changes requested):
Good job! Some code style changes suggested

Rommelua (approved):

PR #1298, opened 2024-02-16
done

ChabVlad:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int MAX_CAPACITY = 10;
+    private Object[] array = new Object[MAX_CAPACITY];
```
olekskov, line 7: 1. Better to parametrize array in the beginning to avoid type-casts in methods
2. It's not max capacity, our array can be much longer after resizes
```suggestion
    private T[] array = (T) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+    private int size;
+
+    public void grow() {
```
olekskov, line 10: It's our internal logic, no need to expose it.
Please also move it after public methods after modificator is changed
```suggestion
    private void grow() {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+        Object[] newArray = new Object[size + size / 2];
+        for (int i = 0; i < size; i++) {
+            newArray[i] = array[i];
+        }
```
olekskov, line 14: https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#use-systemarraycopy-to-move-your-array-elements

In all places where you copy elements use relevant predefined method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        array[size] = value;
+        size++;
+        if (size == array.length) {
+            grow();
+        }
```
olekskov, line 24: 1. Please perform check before operation, not after. Probaly it was last insert by user. Why then we need to extend array?
2. you call grow() method only inside if block. Please include this check inside method
```suggestion
        growIfRequired();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(
+            "Array list index out of bounds. " + index);
+        }
```
olekskov, line 63: Please extract repeated logic to separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+            grow();
+        }
+        for (int i = 0; i < list.size(); i++) {
+            if (array.length == size) {
+                grow();
+            }
+            array[size] = list.get(i);
+            size++;
 
+        }
```
olekskov, line 55: just reuse add(T value) method in loop. It will determine required amount of resizes

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+    }
+
+    private void arrayListIndexOutOfBoundsExceptionMethodIfSizeMore(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Array list index out of bounds. " + index);
+        }
+    }
+
+    private void arrayListIndexOutOfBoundsExceptionMethodIfSizeMoreOrEqual(int index) {
```
olekskov, line 29: private method should be placed after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
-
+        array[size] = value;
+        size++;
+        growIfRequared();
     }
 
     @Override
     public void add(T value, int index) {
-
+        arrayListIndexOutOfBoundsExceptionMethodIfSizeMore(index);
```
olekskov, line 45:
```suggestion
        checkIndexValid(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     public T remove(T element) {
-        return null;
+        T removedValue;
+        for (int i = 0; i <= size - 1; i++) {
+            if ((array[i] != null && array[i].equals(element))
+                    || (element == null && array[i] == null)) {
+                removedValue = (T) array[i];
+                for (int j = i; j < size - 1; j++) {
+                    array[j] = array[j + 1];
+                }
```
olekskov, line 94: Please use System.arraycopy() method in all places where you shifting elements.
This mothod may be use also for copying within one array

olekskov (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        growIfRequared();
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        growIfRequared();
```
```suggestion
```
addAll is functionally equivalent to calling add in a loop, nothing else is needed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        array[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexDataIfSizeEqualOrMore(index);
+        T[] newArray = (T[]) new Object[array.length];
+        System.arraycopy(array, 0, newArray, 0, index);
+        System.arraycopy(array, index + 1, newArray, index,size - index - 1);
```
okuzan, line 51: as Oleksandr said, we can copy array into itself with system.arraycopy - only one call is needed in each method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        return removedValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T removedValue;
+        for (int i = 0; i <= size - 1; i++) {
+            if ((array[i] != null && array[i].equals(element))
+                    || (element == null && array[i] == null)) {
```
okuzan, line 63:
```suggestion
            if (array[i] != null && array[i].equals(element)
                    || element == array[i]) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                removedValue = (T) array[i];
+                System.arraycopy(array, i + 1, array, i, size - i - 1);
+                array[size - 1] = null;
+                size--;
```
```suggestion
                array[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+                return removedValue;
+            }
+        }
+        throw new NoSuchElementException(
+        "Element not found. " + element);
```
```suggestion
        throw new NoSuchElementException("Element not found. " + element);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
     @Override
     public boolean isEmpty() {
+        if (size() == 0) {
+            return true;
+        }
         return false;
     }
+
+    private void grow() {
+        T[] newArray = (T[]) new Object[size + size / 2];
```
okuzan, line 89: create a GROW_FACTOR constant and use it here, for the sake of readability

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        if (size == array.length) {
+            grow();
+        }
+    }
+
+    private void checkIndexData(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Array list index out of bounds. " + index);
+        }
```
okuzan, line 104: including size would be good too, for debugging purposes

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] array = (T[]) new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
-
+        array[size] = value;
+        size++;
+        growIfRequared();
```
okuzan, line 14:
```suggestion
        growIfRequared();
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
         return false;
     }
+
+    private void grow() {
+        T[] newArray = (T[]) new Object[size + size / 2];
+        System.arraycopy(array, 0, newArray, 0, size);
+        array = newArray;
+    }
+
+    private void growIfRequared() {
```
okuzan, line 94:
```suggestion
    private void growIfRequired() {
```
misspelled, IDEA highlights such things with green, be careful, pay attention to warnings

okuzan (changes requested):

Vasul-Pankiv (approved):

PR #1303, opened 2024-02-20
Jv al

ReamFOX:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        values[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        indexValidation(index);
+        T removedElement = values[index];
+        System.arraycopy(values, index + 1, values, index, size - index - 1);
+        values[--size] = null;
```
k0sm0naft, line 61: You anyway checking about size, and user can't get this `value` after u change `size`
```suggestion
        size--;
```

k0sm0naft (changes requested):
Good job, but needed small improvements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        values = (T[]) new Object[DEFAULT_SIZE];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == values.length) {
+            grow();
+        }
```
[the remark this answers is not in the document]
ReamFOX: I changed this because I had a different logic in addAll and forgot that I had placed the if outside the method. But I agree.

src/main/java/core/basesyntax/ArrayList.java
ReamFOX: Yes, you right, i didn`t notice that, thanks.

k0sm0naft (approved):
Great!

PR #1304, opened 2024-02-21
main version

EduarfF:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
```
slizko1, line 4: Don`t use Objects or any util class. That said in checklist.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index is out of range");
+        }
+        if (size == elementData.length) {
+            resize();
+        }
+        for (int i = size; i > index; i--) {
+            elementData[i] = elementData[i - 1];
+        }
+        elementData[index] = value;
+        ++size;
```
slizko1, line 38: Use System.arraycopy() to move your array elements. It said in checklist

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            elementData[size] = list.get(i);
+            size++;
+        }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
```
slizko1, line 58: Extract condition to method because it's repeats 3 times

slizko1 (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        }
+        elementData[size] = value;
+        ++size;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of range");
```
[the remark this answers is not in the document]
karma-o: RuntimeException has a constructor that you can use in your custom one - `super(message, exception)`
we do not have an original exception here to pass there, so you can leave it as it is, just remember this info for the future use

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+    }
+
+    private void resize() {
+        int newSize = size + (size >> 1);
+        Object[] newArr = new Object[newSize];
+        System.arraycopy(elementData, 0, newArr, 0, size);
+        elementData = newArr;
+    }
+
+    private void isIndexInvalid(int index) {
```
karma-o, line 100: method name implies that we should return boolean, so let's do that, or rename it to `checkIndex()`

src/test/java/core/basesyntax/ArrayListTest.java
```diff
@@ -37,6 +37,7 @@ public void addElementToArrayListByIndex() {
         Assert.assertEquals("Test", arrayList.get(0));
         Assert.assertEquals("Test failed! Can't correctly add element by index " + 1,
                 "Academy", arrayList.get(1));
+
```
karma-o, line 40: Let's not change tests

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+
 public class ArrayList<T> implements List<T> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] elementData;
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
karma-o, line 13: primitive class fields have default values
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size;
+    private Object[] elementData;
```
karma-o, line 9: we should use generic types for the fields this way we won't need to cast it to `T` on returns
```suggestion
    private T[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+
+    public ArrayList() {
+        elementData = new Object[DEFAULT_CAPACITY];
```
karma-o, line 12:
```suggestion
        elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        elementData[size] = value;
+        ++size;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index is out of range");
+        }
```
karma-o, line 29: reuse `isIndexValid()` instead

karma-o (changes requested):

src/main/java/core/basesyntax/ArrayList.java
EduarfF: I can't because conditions are different

PR #1306, opened 2024-02-21
Interface List was implemented in ArrayList class

slizko1:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,169 @@
...
+
     @Override
     public void add(T value) {
-
+        if (size == dataArray.length) {
+            grow(dataArray);
+            addOnNextFreeSell(value);
+        } else {
+            addOnNextFreeSell(value);
+        }
```
oksana-divkit, line 32: A small suggestion. We will call `addOnNextFreeSell` in any case, so we can eliminate duplication
```suggestion
        if (size == dataArray.length) {
            grow(dataArray);
        }
        addOnNextFreeSell(value);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,169 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void isDataArrayFull() {
```
oksana-divkit, line 102: Looking at the name `isDataArrayFull`, I assume that this method will return a boolean. It will be better to name as `growIfFull` or something else with a verb describing a process inside.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,169 @@
...
+        if (dataArray.length == size) {
+            grow(dataArray);
+        }
+    }
+
+    private void grow(Object[] fullArray) {
+        dataArray = new Object[fullArray.length + fullArray.length / 2];
```
oksana-divkit, line 109: The number `2` here is a little magical. Maybe it is possible to create a constant with coefficient?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,169 @@
...
+        isDataArrayFull();
+        size++;
+    }
+
+    private void addOnNextFreeSell(T value) {
+        addSize();
+        dataArray[size - 1] = value;
+    }
+
+    private void addByIndex(T value, int index) {
```
oksana-divkit, line 128: `isDataArrayFull()` called here twice (on 129th row and inside `addSize()`)
It's a good sign that a method does too much.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,169 @@
...
+    }
+
+    private T removeElement(int index) {
+        isIndexValid(index);
+        Object objectToDelete = dataArray[index];
+        if (size - 1 == index) {
+            size--;
+            dataArray[index] = null;
+            return (T) objectToDelete;
+        }
```
oksana-divkit, line 162: I'm not sure about this part. Is it necessary here? Below we have the same logic and code besides coping the array;

oksana-divkit (approved):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+            grow(dataArray);
+        }
+    }
+
+    private void grow(Object[] fullArray) {
+        dataArray = new Object[fullArray.length + fullArray.length / 2];
+        System.arraycopy(fullArray, 0, dataArray, 0, fullArray.length);
+    }
+
+    private void growTo(Object[] fullArray, List<T> list) {
```
karma-o, line 109: we do need to pass dataArray as a parameter here, use it from a field reference
```suggestion
    private void growTo(List<T> list) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+        dataArray = new Object[size + list.size()];
+        System.arraycopy(fullArray, 0, dataArray, 0, fullArray.length);
+    }
+
+    private void addSize() {
+        size++;
+    }
+
+    private void addOnNextFreeSell(T value) {
```
karma-o, line 118:
```suggestion
    private void addOnNextFreeCell(T value) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+        growIfFull();
+        addSize();
+        dataArray[size - 1] = value;
```
karma-o, line 121: why -1? just increment size after you have added an element
```suggestion
        dataArray[size] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+            this.dataArray = new Object[capacity];
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        growIfFull();
+        addOnNextFreeSell(value);
     }
```
karma-o, line 29: as Oksana mentioned, you are calling `growIfFull()` 2 times, here, and in the `addOnNextFreeCell()`, so we need to remove it from here, but then what we have left here, is just a call to `addOnNextFreeCell()`, which also not looking good, so we can just remove that private `addOnNextFreeCell()` method, and move its whole logic to the `add(T value)`, and call this public `add(T value)` method in other places instead of `addOnNextFreeCell()`, which would make total sense if you think about it, because we do not always need a common logic to be a private method, we can freely use public methods like in that example

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+    private boolean isIndexExist(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
+        } else {
+            return true;
+        }
+    }
+
+    private void indexValidation(int index) {
+        if (index >= 0 && index <= size) {
```
karma-o, line 140: invert this if to throw exception inside it's body and avoid redundant return

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+        dataArray[index] = value;
+        addSize();
+    }
+
+    private boolean isIndexExist(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
+        } else {
+            return true;
+        }
```
karma-o, line 136:
```suggestion
        }
        return true;
```

karma-o, line 136: overall I do not recommend to mix-up the boolean based validation and exception throwing validation, it's better to pick one and either make it void and name it like `checkIndex()` or make it return boolean and name it `isIndexValid()` and not throw and exception in it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+        indexValidation(index);
+        Object objectToDelete = dataArray[index];
+        if (size - 1 == index) {
+            size--;
+            dataArray[index] = null;
+            return (T) objectToDelete;
+        }
+        size--;
+        System.arraycopy(dataArray, 0, dataArray, 0, index);
+        System.arraycopy(dataArray, index + 1, dataArray, index, size - index);
```
karma-o, line 161: I do not think we need 2 `arraycopy` calls here, not sure, but please double-check it

src/main/java/core/basesyntax/ArrayList.java
karma-o: agree, most likely it is redundant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
     public void set(T value, int index) {
-
+        if (isIndexExist(index)) {
+            dataArray[index] = value;
+        }
     }
 
     @Override
     public T remove(int index) {
+        if (isIndexExist(index)) {
```
karma-o, line 72: this is why I do not recommend to mix up the boolean and exception validation approaches `if` here is essentially redundant because there will never be a situation when it is false, because it will just throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+            return (T) dataArray[index];
+        }
         return null;
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (isIndexExist(index)) {
```
karma-o, line 65: same here

karma-o (changes requested):
Left some comments

src/main/java/core/basesyntax/ArrayList.java
slizko1: In first arraycopy i copy values before index in next i copy values after index and pste them from index position

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final Object[] EMPTY_DATAARRAY = {};
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] dataArray;
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,165 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.dataArray = new Object[DEFAULT_CAPACITY];
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        growIfFull();
+        dataArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        indexValidation(index);
+        if (index == size && size != dataArray.length) {
```
andrii-hoienko, line 25: You make an index validation and then next line you make another index validation, maybe it makes sense to unite this logic, or overload method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+            add(value);
+        } else {
+            addByIndex(value, index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list.isEmpty()) {
```
andrii-hoienko, line 34: This is redundant, you can iterate and if list is empty 0 iteration will be executed, and it is ok

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+            return;
+        }
+        if (list.size() + size > dataArray.length) {
+            growTo(list);
```
andrii-hoienko, line 38: No reason to grow here, we can grow in method 'add'

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        }
+        for (int i = 0; i < list.size(); i++) {
 
+            this.add(list.get(i));
```
andrii-hoienko, line 42:
```suggestion
            add(list.get(i));
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+
+    private void growIfFull() {
+        if (dataArray.length == size) {
+            grow();
+        }
+    }
+
+    private void grow() {
+        Object[] fullArray = dataArray;
+        dataArray = new Object[fullArray.length + fullArray.length / 2];
```
andrii-hoienko, line 92: It is required to grow in 1.5 times, so i suggest
dataArray = new Object[fullArray.length* GROWTH_FACTOR];

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+            return;
+        }
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
+        } else {
+            return;
+        }
```
andrii-hoienko, line 122:
```suggestion
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
+        } else {
+            return;
+        }
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
```
andrii-hoienko, line 119:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        return (element == null && listElement == null)
+                || (element != null && element.equals(listElement));
+    }
+
+    private T removeElement(int index) {
+        indexValidation(index);
+        size--;
+        System.arraycopy(dataArray, 0, dataArray, 0, index);
+        Object objectToDelete = dataArray[index];
+        System.arraycopy(dataArray, index + 1, dataArray, index, size - index);
```
andrii-hoienko, line 135: I believe we can make  System.arraycopy only once

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        dataArray[index] = value;
+        size++;
+    }
+
+    private void isIndexExist(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("This index don't exist");
+        } else {
+            return;
+        }
```
andrii-hoienko, line 114:
```suggestion
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+    }
+
+    private void addByIndex(T value, int index) {
+        growIfFull();
+        System.arraycopy(dataArray, index, dataArray, index + 1, size - index);
+        dataArray[index] = value;
+        size++;
+    }
+
+    private void isIndexExist(int index) {
```
andrii-hoienko, line 109: Methods 'isIndexExist' and 'indexValidation' could be collapsed into one by adding additional boolean parameter, because look like their purpose is the same

andrii-hoienko (changes requested):

karma-o (changes requested):

PR #1312, opened 2024-02-24
added the necessary methods

AnyaRatusnyak:


makaroshyna (changes requested):
GJ! Move index checks to the new method and it will be a lot better)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
```
karma-o, line 14: * primitive fields have default value
* `this` is redundant since there are no other variables with similar name 
```suggestion
    public ArrayList() {
        elementData = (T[]) new Object[DEFAULT_CAPACITY];
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        int oldCapacity = elementData.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        return elementData = Arrays.copyOf(elementData, newCapacity);
     }
 
     @Override
-    public void add(T value, int index) {
+    public void add(T value) {
+        if (size == elementData.length) {
+            elementData = grow();
```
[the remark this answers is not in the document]
karma-o: agree, let's just call a method here and move the new `elementData` assignment to the grow method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        }
     }
 
     @Override
-    public T get(int index) {
-        return null;
+    public T get(int index) throws ArrayListIndexOutOfBoundsException {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("the index is invalid");
+        }
```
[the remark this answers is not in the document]
karma-o: agree, we should move all index checks to a separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("the index is invalid");
```
karma-o, line 48: we should also add index that we have received to the message and maybe even a current size of the list

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("the index is invalid");
+        }
+        final T removedElement = elementData[index];
+        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+        elementData[size - 1] = null;
+        size--;
```
karma-o, line 70:
```suggestion
        elementData[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] grow() {
+        int oldCapacity = elementData.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        return elementData = Arrays.copyOf(elementData, newCapacity);
```
karma-o, line 98: `Arrays` class is not allowed to use in this task, check common mistakes

karma-o (changes requested):

andrii-hoienko (approved):
Good job!

PR #1313, opened 2024-02-25
All assumptions met, code passed tests

TylTejlor:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        ensureCapacity(size + 1);
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
```
kerrrusha, line 26: violating SRP principle, should be extracted to separate method
```suggestion
        checkIndexForAdd(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index");
+        }
```
kerrrusha, line 60: violating SRP principle and dublicates code, should be extracted to separate method
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
kerrrusha, line 81: can be simplified into 1 loop by changing `if` condition

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            if (element.equals(elements[i])) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = elements.length * 2;
```
kerrrusha, line 106: and make 1.5 as constant
```suggestion
            int newCapacity = elements.length * 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    private int size;
+
+    public ArrayList() {
+        elements = (T[])new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity(size + 1);
```
nowickimj, line 18: Since every `ensureCapacity` call involves `size + 1` argument and `size` itself is a field, you don't need to pass this argument at all - you can calculate it within the method.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        T removedElement = (T) elements[index];
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[--size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
```
nowickimj, line 61: You've implemented `indexOf(T element)` method - why not use it here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int indexOf(T element) {
+        for (int i = 0; i < size; i++) {
+            if (element.equals(elements[i])) {
```
nowickimj, line 82: Null check for `element` required here.

nowickimj (changes requested):
Few comments

nowickimj (approved):
Great job!

PR #1314, opened 2024-02-25
 implemented methods add,remove,addAll,get,getSize,isEmpty

DmytroHadiuchko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= 0;
+    }
+
+    private void newSizeForArray() {
+        int oldCapacity = array.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
+        array = Arrays.copyOf(array, newCapacity);
```
VolandevlodD, line 105: Don't use Objects, Arrays, or any other util class.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= 0;
```
VolandevlodD, line 99: Probably, the size will always be >= 0;
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        array[size - 1] = null;
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(array[i], element)) {
```
VolandevlodD, line 85: Don't use Objects, Arrays, or any other util class.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + " , Size: " + size);
+        }
+        if (size == array.length) {
+            newSizeForArray();
+        }
+        for (int i = size; i > index; i--) {
+            array[i] = array[i - 1];
+        }
```
VolandevlodD, line 37: It`s reccomended to use System.arraycopy() to move your array elements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        array[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() > array.length) {
+            newSizeForArray();
```
VolandevlodD, line 45: What if list.size exceeds  newSizeForArray?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        for (int i = 0; i < list.size();i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + " , Size: " + size);
```
VolandevlodD, line 55: It's a good practice not to copy-paste the same literals; it's better to use constants.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+
+    private Object[] array;
```
VolandevlodD, line 10: The name 'array' sounds too general; let's try something like 'elementData'.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + " , Size: " + size);
+        }
+        final T removedElement = (T) array[index];
+        for (int i = index; i < size - 1; i++) {
+            array[i] = array[i + 1];
+        }
+        array[size - 1] = null;
+        size--;
```
VolandevlodD, line 78: It's recommended in the checklist to use System.arraycopy() for moving array elements.

VolandevlodD (changes requested):
**It looks good. I've left some comments below with the requested changes due to the checklist.**

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return (T) elementData[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_FOR_MESSAGE
+                    + index + SIZE_FOR_MESSAGE + size);
+        }
```
VolandevlodD, line 65: Since you're reusing this code, you can put it in a separate method like in the example provided. Then, you can simply call the method instead of copying and pasting the same code.
```
private void isIndexLegal(int index) {
        if (!(index > -1 && index < size)) {
            throw new ArrayListIndexOutOfBoundsException(String.format(INDEX_ERR_MSG, index,size));
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        elementData[size - 1] = null;
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null && elementData[i] == null
```
VolandevlodD, line 85:
```suggestion
            if (element ==  elementData[i]
```
If they both null this expression will return true

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+                    || elementData[i] != null && elementData[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
VolandevlodD, line 90: Some message could be added.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void newSizeForArray(int capacity) {
+        int oldCapacity = elementData.length;
+        int newCapacity = oldCapacity + (capacity >> 1);
+        T[] newElementData = (T[]) new Object[newCapacity];
+        System.arraycopy(elementData,0,newElementData,0,size);
+        elementData = newElementData;
```
VolandevlodD, line 108: You don't need to accept a parameter capacity here, since you already have access to the field size. You can simply use it inside the private method.

VolandevlodD (changes requested):
It looks good, but I would recommend some improvements in the code.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final String INDEX_FOR_MESSAGE = "Index: ";
+    private static final String SIZE_FOR_MESSAGE = ", Size:";
```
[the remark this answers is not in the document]
karma-o: It is fine to use formats too, IMHO constants in exception messages actually reduce readability, and I would avoid them in that case, but that is totally up to you 😉

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+
+    private Object[] elementData;
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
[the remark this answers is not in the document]
karma-o: agree

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final String INDEX_FOR_MESSAGE = "Index: ";
+    private static final String SIZE_FOR_MESSAGE = ", Size:";
+
+    private Object[] elementData;
```
karma-o, line 10: saves us from casting the values on returns
```suggestion
    private T[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    private int size;
+
+    public ArrayList() {
+        this.elementData = new Object[DEFAULT_CAPACITY];
```
karma-o, line 14: for `T[]` field to work fine + redundant `this`
```suggestion
        elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        this.size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length || size > elementData.length) {
+            newSizeForArray(size);
```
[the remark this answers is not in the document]
karma-o: agree, no need to pass it as a parameter, and we can name it like `grow()`

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        System.arraycopy(elementData,index, elementData,index + 1,size - index);
+        elementData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        if (size + list.size() > elementData.length || size > elementData.length) {
+            newSizeForArray(size);
```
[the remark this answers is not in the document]
karma-o: a good suggestion to avoid multiple `grow()` calls

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_FOR_MESSAGE
+                    + index + SIZE_FOR_MESSAGE + size);
+        }
```
karma-o, line 32: index checks are duplicated across all methods, extract them to a private one

src/main/java/core/basesyntax/ArrayList.java
karma-o: should be added 😉

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_FOR_MESSAGE
+                    + index + SIZE_FOR_MESSAGE + size);
+        }
+        final T removedElement = (T) elementData[index];
+        System.arraycopy(elementData,index + 1, elementData,index,size - index - 1);
+        elementData[size - 1] = null;
+        size--;
```
karma-o, line 78:
```suggestion
        elementData[--size] = null;
```

src/main/java/core/basesyntax/ArrayList.java
karma-o: or actually, make use of it by recieving a new size here from `addAll()` method to avoid multiple array extensions if we receive a big list there

karma-o (changes requested):

PR #1315, opened 2024-02-25
Implement array list.

makaroshyna:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        }
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index " + index + " is out of bounds!");
```
nklimovych, line 27: You can put exception message into constant to provide DRY rule

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    }
+
+    private void grow() {
+        int newCapacity = (int) (elements.length * GROWS_COEFFICIENT);
+        T[] biggerArray = (T[]) new Object[newCapacity];
+        System.arraycopy(elements, 0, biggerArray, 0, elements.length);
+        elements = biggerArray;
+    }
+
+    private boolean isCapacityReached() {
```
nklimovych, line 92: This method is redundant because you use it only twice, but it has a fairly negative impact on code readability

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return size == elements.length;
+    }
+
+    private void checkIndex(int index) {
```
nklimovych, line 96:
```suggestion
    private void validateIndex(int index) {
```
I may be wrong, but it seems to me that this name will be more informative

nklimovych (changes requested):
Looks good, but there are a few suggestions

src/main/java/core/basesyntax/ArrayList.java
makaroshyna: I don't think DRY is relevant here, as I only have 1 exception in a different method. So there is no repetetive code here 🙃

src/main/java/core/basesyntax/ArrayList.java
makaroshyna: There is checkIndex in Objects class that throws IndexOutOfBoundsException and that's quite similar to mine method.

PR #1316, opened 2024-02-25
Implement custom ArrayList class (jv-array-list)

VolandevlodD:
Created a custom ArrayList class implementing the List interface. 
Overrides required methods to mimic java.util.ArrayList behavior, including default capacity and dynamic array growth.
Throws custom ArrayListIndexOutOfBoundsException for invalid index passed to methods.
Throws NoSuchElementException in remove(T element) if the element is not present.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,146 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int ARRAY_BEGINNING = 0;
+
+    private int size;
+    private Object[] elementStore;
+
+    ArrayList() {
+        size = 0;
+        elementStore = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfArrayFull();
+        elementStore[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (indexIsLegal(index) || index == size) {
+            size++;
+            growIfArrayFull();
+            System.arraycopy(elementStore, index, elementStore,
+                    nextIndex(index), calculateElementsToCopy(index));
+            set(value, index);
+        } else {
+            throw createErrorWithIndex(index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
+    }
 
+    private int calculateElementsToCopy(int index) {
+        return size - index;
+    }
+
+    private ArrayListIndexOutOfBoundsException createErrorWithIndex(int index) {
+        return new ArrayListIndexOutOfBoundsException("Index: [" + index
+                + "] is out of bounds for list size [" + size + "]!");
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (indexIsLegal(index)) {
+            return (T) elementStore[index];
+        } else {
+            throw createErrorWithIndex(index);
+        }
     }
 
-    @Override
-    public void set(T value, int index) {
+    private void growIfArrayFull() {
+        if (isFull()) {
+            int newCapacity = size + size / 2;
+            Object[] oldStore = new Object[size];
+            System.arraycopy(elementStore, ARRAY_BEGINNING, oldStore, ARRAY_BEGINNING, size);
+            elementStore = new Object[newCapacity];
+            System.arraycopy(oldStore, ARRAY_BEGINNING, elementStore, ARRAY_BEGINNING, size);
+        }
+    }
 
+    private int nextIndex(int index) {
+        return index + 1;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T oldValue;
+        if (indexIsLegal(index)) {
+            oldValue = (T) elementStore[index];
+            if (isFull()) {
+                elementStore[index] = null;
+                size--;
+                return oldValue;
+            }
+            System.arraycopy(elementStore, nextIndex(index),
+                    elementStore, index, calculateElementsToCopy(index));
+            elementStore[size] = null;
+            size--;
+            return oldValue;
+        } else {
+            throw createErrorWithIndex(index);
+        }
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((elementStore[i] == element)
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+        return size - index;
+    }
+
+    private ArrayListIndexOutOfBoundsException createErrorWithIndex(int index) {
+        return new ArrayListIndexOutOfBoundsException("Index: [" + index
+                + "] is out of bounds for list size [" + size + "]!");
+    }
+
+    private void growIfArrayFull() {
+        if (isFull()) {
```
fedaktaras, line 127:
```suggestion
    private void growIfArrayFull() {
        if (size == elementStore.length) {
```
method with one line. Not necessary.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+            int newCapacity = size + size / 2;
```
fedaktaras, line 128: As I remember I should grow 1.5 times. Make it constant.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+            Object[] oldStore = new Object[size];
+            System.arraycopy(elementStore, ARRAY_BEGINNING, oldStore, ARRAY_BEGINNING, size);
+            elementStore = (T[]) new Object[newCapacity];
+            System.arraycopy(oldStore, ARRAY_BEGINNING, elementStore, ARRAY_BEGINNING, size);
```
fedaktaras, line 132: can be done with one System.arraycopy. You don't need to split here.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+        }
+    }
+
+    private boolean indexIsLegal(int index) {
```
fedaktaras, line 136:
```suggestion

    private boolean isLegalIndex(int index) {
```
methods that return boolean should start with is etc. Like question sentence "Is it valid index?"

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+            throw createErrorWithIndex(index);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
fedaktaras, line 42: if you add to list with size 5 list with size 40 it will grow 4 times. You may calculate what new capacity you need and make it once. growToCapacity(this.size + list.size) 
Also each time it will check is it full

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
+    private static final int ARRAY_BEGINNING = 0;
+    private static final int EMPTY_SIZE = 0;
+
+    private int size;
+    private T[] elementStore;
+
+    ArrayList() {
+        size = EMPTY_SIZE;
+        elementStore = (T[]) new Object[DEFAULT_CAPACITY];
+    }
```
fedaktaras, line 16:
```suggestion

    private int size;
    private T[] elementStore;

    ArrayList() {
        elementStore = (T[]) new Object[DEFAULT_CAPACITY];
    }
```
it is always 0 when toy create it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,147 @@
...
-    public T remove(int index) {
-        return null;
+    public T remove(T element) {
+        for (int i = 0; i < size; i++) {
+            if ((elementStore[i] == element)
+                    || (elementStore[i] != null && elementStore[i].equals(element))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
```
fedaktaras, line 82: Some message may be added.

fedaktaras (changes requested):
It might be good to form error message inside method. I believe I will have to do the same.
I think System.arraycopy may be used a bit simpler way. but it just my opinion.

src/main/java/core/basesyntax/ArrayList.java
VolandevlodD: This expression is reused several times in the code, so I think it's good not to copy-paste it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,156 @@
...
+    }
+
+    private int calculateElementsToCopy(int index) {
+        return size - index;
+    }
+
+    private ArrayListIndexOutOfBoundsException createErrorWithIndex(int index) {
+        return new ArrayListIndexOutOfBoundsException("Index: [" + index
+                + "] is out of bounds for list size [" + size + "]!");
+    }
```
karma-o, line 125: usually, there is no need to build the exception in a separate method, because in my opinion it makes it hard to read.
What I would do instead is make another method named `checkIndex()` which would perform a check if it is legal and throw an exception if it is not, then called it at the beginning of each method

Because what you have right now is all of your "normal logic flow" which is like a flow that we assume when no problems are encountered during the method execution is inside the if statement, which we want to avoid, because if we have 50 parameters that we need to validate - our normal flow would be even deeper inside those if statements.
So always try to keep the next order:
```java
public void method(String param1, String param2) {
    if (/*param1 is not valid*/) {
        // throw ex
    }
    checkParam2(param2);
    // business logic
}

private void checkParam2(String param2) {
    if (/*param2 is not valid*/) {
        // throw ex
    }
}
```
So please check all your methods and try to adapt them to use this pattern
More about this pattern: https://medium.com/swlh/return-early-pattern-3d18a41bba8

karma-o (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int ARRAY_BEGINNING = 0;
+    private static final int GROWTH_FACTOR = 2;
+    private static final String INDEX_ERR_MSG_PREFIX = "Index: [";
+    private static final String INDEX_ERR_MSG_MIDDLE = "] is out of bounds for list size [";
+    private static final String INDEX_ERR_MSG_SUFFIX = "]!";
```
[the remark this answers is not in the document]
VolandevlodD: Nice metod, thank you!

PR #1317, opened 2024-02-26
Implemented List interface in ArrayList class.

nalforov:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity >= dataArray.length) {
+            grow(minCapacity);
+        }
+    }
+
+    private void grow(int minCapacity) {
+        int oldCapacity = dataArray.length;
+        int newCapacity = oldCapacity + (oldCapacity >> 1);
```
okuzan, line 107: create a clearly defined GROW_FACTOR constant instead

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,113 @@
...
+        dataArray[size - 1] = null;
+        size--;
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(dataArray[i], element)) {
```
okuzan, line 66: [checklist issue ](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class)

src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
```diff
@@ -1,7 +1,7 @@
 package core.basesyntax;
 
 public class ArrayListIndexOutOfBoundsException extends RuntimeException {
-    public ArrayListIndexOutOfBoundsException(String message) {
-        super(message);
+    public ArrayListIndexOutOfBoundsException(int index, int size) {
+        super("Index: " + index + "; size: " + size);
```
okuzan, line 5: You shouldn't make changes to this class, why limit yourself to only one exception message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_FACTOR = 1.5;
+    private int size;
+    private T[] dataArray;
+
+    public ArrayList() {
+        this.dataArray = (T[]) new Object[DEFAULT_CAPACITY];
```
```suggestion
        dataArray = (T[]) new Object[DEFAULT_CAPACITY];
```
where do we need [`this`](https://stackoverflow.com/a/2411283)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+        }
+    }
+
+    private void ensureCapacity() {
+        if (size == dataArray.length) {
+            grow(size + 1);
+        }
+    }
+
+    private void grow(int minCapacity) {
```
okuzan, line 100: if we always pass the same argument, we don't need to pass argument

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+        int oldCapacity = dataArray.length;
+        int newCapacity = (int) (oldCapacity * GROW_FACTOR);
+        if (newCapacity - minCapacity < 0) {
+            newCapacity = minCapacity;
+        }
```
```suggestion
```
do we really need it?

src/main/java/core/basesyntax/ArrayList.java
okuzan, line 100: Move grow to ensure capacity, enough to only have one private method for this part of logic

okuzan (changes requested):

kozhukhovsky (approved):

PR #1318, opened 2024-02-26
Added implementation of all methods in ArrayList class

nklimovych:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE + index);
+        }
+        if (size == elementData.length) {
+            elementData = grow();
```
JohnGrey17, line 38: Screenshot (126)
i guess we can remove that if block and make that condition in grow() method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elementData[i], element)) {
```
JohnGrey17, line 82: in check list there is information that we cannot use Object methods
Screenshot (127)

JohnGrey17 (changes requested):

PR #1319, opened 2024-02-26
Created own ArrayList class which implements the List interface

ValentynSharshon:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private Object[] elementData;
```
Recdt, line 8: Why you didn't use T[] element data? Why is it Object[]?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    }
+
+    public ArrayList() {
+        this(DEFAULT_CAPACITY);
+    }
+
     @Override
     public void add(T value) {
-
+        increaseDataArrayIfFull();
```
Recdt, line 24: Won't calling the function load the stack a little more each time, maybe it's better to do an additional check?

Recdt (approved):

src/main/java/core/basesyntax/ArrayList.java
ValentynSharshon: The default implementation of AarayList uses an array of objects as an item store. Perhaps it can be explained by the fact that we don't work directly with this storage, and it makes no sense to use generics there, but I could be wrong.

src/main/java/core/basesyntax/ArrayList.java
ValentynSharshon: This function is not used recursively, it used just a few times. I don't think that calling it has much effect on the overall performance of the application.

PR #1320, opened 2024-02-26
Created my array list

Recdt:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int NOT_EXISTS = -1;
+    private static final int UNIT = 1;
```
Recdt, line 8: I don't know how to name that, so I named it like that.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (internalArray[i] == null ? element == null : internalArray[i].equals(element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("No such element in ArrayList!!!");
```
oksana-divkit, line 90: `!!!` is too emotional, don't you think? 😄

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            if (size == capacity) {
+                growIfArrayFull();
+            }
+            System.arraycopy(internalArray, index, internalArray,
+                    index + UNIT, size - index - UNIT);
+            internalArray[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Unavailable index "
+                    + "for adding element!!!");
+        }
```
oksana-divkit, line 43: I think, this part can be simplified a bit. And if you create a method for exception, the code becomes even clearer.
```suggestion
        if (size == index) {
            add(value);
            return;
        } 
        if (!isIndexAcceptable(index)) {
            throw new ArrayListIndexOutOfBoundsException("Unavailable index "
                    + "for adding element!!!");
        }
        size++;
        if (size == capacity) {
            growIfArrayFull();
        }
        System.arraycopy(internalArray, index, internalArray,
                index + UNIT, size - index - UNIT);
        internalArray[index] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
     @Override
     public void add(T value) {
-
+        if (size < capacity) {
+            internalArray[size] = value;
+            size++;
+        } else {
+            growIfArrayFull();
+            add(value);
+        }
```
oksana-divkit, line 25: I think, this part can be simplified a bit, since in any case we will add this element:
```suggestion
        if (size >= capacity) {
            growIfArrayFull();
        }
        internalArray[size] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Can't get out of bound index!!!");
+        }
+        return internalArray[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (!isIndexAcceptable(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get out of bound index!!!");
```
oksana-divkit, line 64: It would be nice to show the index and current size of the array in the message.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-
+        if (size == index) {
+            add(value);
+        } else if (isIndexAcceptable(index)) {
+            size++;
+            if (size == capacity) {
+                growIfArrayFull();
+            }
+            System.arraycopy(internalArray, index, internalArray,
+                    index + UNIT, size - index - UNIT);
```
oksana-divkit, line 38: From `checklist.md`:
> Resize the array in a separate method.

oksana-divkit (changes requested):
Wow, very cool solution! However, due to some violations from `checklist.md`, I am forced to reject it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (!isIndexAcceptable(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Can't get out of bound index!!!");
+        }
```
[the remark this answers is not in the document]
Recdt: This structure in the code returns different errors each time depending on the location of the error in the code. That is to say, it will be incorrect, in my opinion

src/main/java/core/basesyntax/ArrayList.java
Recdt: image
Is a growIfArrayFull() method is not a separate method for resizing?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+        if (isIndexAcceptable(index)) {
+            T temporary = internalArray[index];
+
+            System.arraycopy(internalArray, index + UNIT, internalArray,
+                    index, size - index - UNIT);
+            size--;
+            return temporary;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Can't remove non-existent element");
+        }
```
andrii-hoienko, line 81:
```suggestion
        if (isIndexAcceptable(index)) {
            T temporary = internalArray[index];

            System.arraycopy(internalArray, index + UNIT, internalArray,
                    index, size - index - UNIT);
            size--;
            return temporary;
        }
        throw new ArrayListIndexOutOfBoundsException("Can't remove non-existent element");
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
+            growIfArrayFull();
+        }
+        internalArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (size == index) {
```
andrii-hoienko, line 29: This method could be simplified I suggest to create one method that will check if index is ok instead of 'size == index' and !isIndexAcceptable(index)

andrii-hoienko (changes requested):

PR #1325, opened 2024-02-27
Implemented the ArrayList class

ghost:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void expandListIfNeeded(int neededSpace) {
+        if (neededSpace > objects.length) {
+            int newLength = (int) (objects.length * 1.5);
```
ValentynSharshon, line 87: Maybe we should put 1.5 in a constant, something like this GROWTH_CAPACITY?

ValentynSharshon (approved):
In my opinion, you have a great implementation of ArrayList.

src/main/java/core/basesyntax/ArrayList.java
VlasenkoYevhenii: You are definitely right!

yevheniisynytsia (changes requested):
Great job! But let's optimise it a bit

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+
+    private void expandListIfNeeded(int neededSpace) {
+        if (neededSpace > objects.length) {
+            int newLength = (int) (objects.length * GROW_FACTOR);
+            if (newLength < neededSpace) {
+                newLength = neededSpace;
+            }
+            Object[] newArray = new Object[newLength];
+            System.arraycopy(objects, 0, newArray, 0, objects.length);
+            objects = newArray;
```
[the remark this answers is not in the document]
VlasenkoYevhenii: I can't get what you mean. There is only one invocation of System.arraycopy() in the add() method and only if the array needs to be expanded.
There optionally can be two invocations in the public void add(T value, int index), but it's just a conditional action. I can't imagine any way to change it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+        }
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException(INDEX_EXCEPTION_MESSAGE);
```
karma-o, line 100: you are using the same message for different conditions, and 1 of them is wrong + we should include that index itself in the message
And my personal opinion: making exception message a constant - actually reduces readability, but this one is up to you, partially because of situations like this, where we have 1 condition and a message just does not correspond to it, and we can not see that because it is a constant

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
karma-o: Better to ask questions in the chat on the platform because we rarely see them here.
I am also not sure what Yevhenii meant, I am going to approve, but let's ask him in the chat what is this about

karma-o (approved):
Approved, but please ask Yevhenii in the chat what was his comment about, and check another one of mine

PR #1326, opened 2024-02-27
Jv array list solution

vikkushnir:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private static final String NO_SUCH_ELEMENT_EXCEPTION_MESSAGE = "Element not found in the list";
+    private T[] elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
-
+        growIfArrayFull();
+        elementData[size] = value;
+        size++;
```
Nikname2303, line 19: you can do that in 1 row

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        System.arraycopy(elementData, DEFAULT_INDEX, newArray, DEFAULT_INDEX, size);
+        elementData = newArray;
+    }
+
+    private void growIfArrayFull() {
+        if (size == elementData.length) {
+            int newCapacity = (int) (elementData.length * GROWTH_FACTOR);
+            resizeArray(newCapacity);
+        }
+    }
```
Nikname2303, line 94: you can do it in one method

Nikname2303 (approved):
Good job))) only some proposes for simplify

src/main/java/core/basesyntax/ArrayList.java
vikkushnir: I'm not sure, but perhaps on QnA, the mentor told us better to use size++ under the line, need to reask

src/main/java/core/basesyntax/ArrayList.java
vikkushnir: Thank you, will think in this place

PR #1327, opened 2024-02-27
First solution

Nikname2303:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_SIZE = 10;
+    private static final double CAPACITY_INCREASE_MODIFICATOR = 1.5;
+    private Object[] defaultList;
```
kaiiiseeel, line 8: Our class is parameterized, so maybe we should declare defaultList as T[] array not the Object[]?
`private Object[] defaultList;`

kaiiiseeel (approved):
LGFM! Left a comment

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void increaseCapacity() {
```
andrii-hoienko, line 87: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+            int newCapacity = (int) (defaultList.length * CAPACITY_INCREASE_MODIFICATOR);
+            Object[] newObjectArr = new Object[newCapacity];
+            System.arraycopy(defaultList, 0, newObjectArr, 0, size);
+            defaultList = newObjectArr;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index not found");
```
andrii-hoienko, line 98:
```suggestion
            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        increaseCapacity();
+        System.arraycopy(defaultList, index, defaultList, index + 1, size - index);
+        defaultList[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
andrii-hoienko, line 41: foreach would be better here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public void add(T value) {
-
+        increaseCapacity();
+        defaultList[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
```
andrii-hoienko, line 30: It is possible to unite this code with 'checkIndex' method to avoid duplication, maybe 'checkindex' should take an additional argument or be overloaded

src/main/java/core/basesyntax/ArrayList.java
andrii-hoienko, line 8:
```suggestion
    private Object[] items; //Actually it is not list, it is array, and why default? What if capacity is not default it is still default list?
```

andrii-hoienko (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Nikname2303: discussed on QnA

andrii-hoienko (approved):
Good job!

PR #1328, opened 2024-02-27
implemented simplified version of ArrayList

kaiiiseeel:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_BY = 1.5;
```
DmytroHadiuchko, line 7: need to add an empty line after the static variable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public boolean isFull() {
```
DmytroHadiuchko, line 100: I think that this method has to be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size == elementData.length;
+    }
+
+    public boolean isCorrectIndex(int index) {
```
DmytroHadiuchko, line 104: I think that this method has to be private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        if (index >= 0 && index < size) {
+            return true;
+        }
+        throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
DmytroHadiuchko, line 108: I'm not sure, but maybe we need to make "Invalid index" a constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        for (int i = 0; i < elementData.length; i++) {
+            if ((element == elementData[i])
+                    || (element != null && element.equals(elementData[i]))) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException();
+    }
+
+    public void grow() {
```
DmytroHadiuchko, line 83: I think that this method has to be private

DmytroHadiuchko (changes requested):

PR #1329, opened 2024-02-27
Added fields -2 constants , array of T and size to calculate size of …

azzilian:
…list

Methods for calculate capacity and 2 for checking index in 2 operations

other methods implementeds

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,103 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_MULTIPLIER = 1.5;
+
+    private T[] elementsData;
+    private int size;
+
+    public ArrayList() {
+        this.elementsData = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
+
+    private void indexExistence(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds: " + index);
+        }
+    }
+
+    private void resizeCheck(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index for add operation: "
+                    + index);
+        }
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elementsData.length) {
+            int newCapacity = (int) (elementsData.length * CAPACITY_MULTIPLIER);
+            elementsData = Arrays.copyOf(elementsData, Math.max(newCapacity, minCapacity));
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity(size + 1);
+        elementsData[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        resizeCheck(index);
+        ensureCapacity(size + 1);
+        System.arraycopy(elementsData, index, elementsData, index + 1, size - index);
+        elementsData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(size + list.size());
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double CAPACITY_MULTIPLIER = 1.5;
+    private static final int DEFAULT_CAPACITY = 10;
+
```
rel1c-hub, line 9: No need in empty line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    }
+
+    private void resizeCheck(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index for add operation: "
+                    + index);
+        }
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
rel1c-hub, line 31: Are you creating a new variable name just for one method?
And it's unnecessary, you can call class-level variables.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        ensureCapacity(size + 1);
+        System.arraycopy(elementsData, index, elementsData, index + 1, size - index);
+        elementsData[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        ensureCapacity(size + list.size());
```
rel1c-hub, line 55: You don't need to call this method; it's included in the add() method that you are calling in iteration.

rel1c-hub (approved):
Overall, it is an exemplary implementation, with some minor changes, and you are good to go. 
Well done, mate!

src/main/java/core/basesyntax/ArrayList.java
karma-o: actually, static fields are usually separated from the usual ones 😉

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private static final double CAPACITY_MULTIPLIER = 1.5;
+    private static final int DEFAULT_CAPACITY = 10;
+
+    private T[] elementsData;
+    private int size;
+
+    public ArrayList() {
+        this.elementsData = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
+    }
```
karma-o, line 16: redundant `this` bacause there are no other variables with such a name in scope and `size` has a default value as it is a primitive field
```suggestion
    public ArrayList() {
        elementsData = (T[]) new Object[DEFAULT_CAPACITY];
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+
+    private void indexExistence(int index) {
```
karma-o, line 18: all private methods go after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index for add operation: "
+                    + index);
+        }
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elementsData.length) {
+            int newCapacity = (int) (elementsData.length * CAPACITY_MULTIPLIER);
+            elementsData = Arrays.copyOf(elementsData, Math.max(newCapacity, minCapacity));
```
karma-o, line 34: Arrays are not allowed to use in this task, check common mistakes

src/main/java/core/basesyntax/ArrayList.java
karma-o: it is fine to call it here, because we check for the whole list size at once, but the fact that it is also being called in the `add(T value)`.. I think I would rather add a custom logic of adding a whole list instead of reusing `add(T value)` because of this, and this should make it more efficient, but overall this is not a big deal if you were to leave it as it is

src/main/java/core/basesyntax/ArrayList.java
azzilian: Not sure, kinda burn with it few times(

PR #1336, opened 2024-03-02
Task completed

apgroupua:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        growArray();
+        System.arraycopy(array, index, array, index + 1, size - index);
+        array[index] = value;
+        size++;
```
olekskov, line 29: I would suggest reusage of existing methods
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        growArray();
        System.arraycopy(array, index, array, index + 1, size - index);
        array[index] = value;
        size++;
```

olekskov (approved):
Nice job, approved.
Comment left for consideration only

PR #1346, opened 2024-03-12
implemented ArrayList program logic

kriznychenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        arrayList = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == arrayList.length) {
+            resize(size);
+        }
```
okuzan, line 19: move if inside of resize, because they are only ever used in conjunction

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
+                    + " out of ArrayListSize");
+        }
+
+        if (size == arrayList.length) {
+            resize(arrayList.length * 2);
```
okuzan, line 31: magic numbers should be constants with descriptive names

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        }
+
+        for (int i = size; i > index; i--) {
+            arrayList[i] = arrayList[i - 1];
+        }
```
okuzan, line 36: use system.arraylist to move elements

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        arrayList[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        int arrayPlusListSize = size + list.size();
+        if (arrayPlusListSize > arrayList.length) {
+            resize(arrayPlusListSize);
+        }
```
okuzan, line 47: redundant 
addAll is functionally equivalent to adding in a loop
resizing logic is in add already

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
+                    + " out of ArrayListSize");
+        }
```
okuzan, line 59: repeated logic, you should create a private method for this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        System.arraycopy(arrayList, index + 1, arrayList, index, numMoved);
+        size--;
+        arrayList[size] = null;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
```
okuzan, line 88: rewrite this method without using such index approach. Return inside of a loop directly, throw exception without any checking after the loop. If we exited the loop, it means nothing is found. This is more common approach in such situations. Add exception message, include element in it

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+        Object[] newArray = new Object[size + (size / 2)];
+
+        System.arraycopy(arrayList, 0, newArray, 0, arrayList.length);
+
+        arrayList = newArray;
+    }
+
+    public static boolean equals(Object a, Object b) {
+        return (a == b) || (a != null && a.equals(b));
+    }
```
okuzan, line 123: access modifier issue. Should they be exposed or hidden?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+            return true;
+        }
         return false;
     }
+
+    public void resize(int size) {
+        Object[] newArray = new Object[size + (size / 2)];
+
+        System.arraycopy(arrayList, 0, newArray, 0, arrayList.length);
+
```
okuzan, line 117: remove redundant empty lines

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
     }
+
+    public void resize(int size) {
+        Object[] newArray = new Object[size + (size / 2)];
```
okuzan, line 114: create a GROW_CONSTANT, it's not readable as is

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] arrayList;
+    private int size;
+
+    public ArrayList() {
+        arrayList = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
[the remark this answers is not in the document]
kriznychenko: I don`t understand it. I`d tried to equale size to arrayList.length, but it doesn`t work correctly.

kriznychenko:
Sorry, but I can`t find a correct solution

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
okuzan: 0 is default for int, you made a right choice removing this line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int GROW_CONSTANT = 2;
```
```suggestion
    private static final int GROW_CONSTANT = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private T[] arrayList;
```
okuzan, line 8: arraylist field of array list - can be a bit confusing, change to smth like values, elements etc

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
+                    + " out of ArrayListSize");
+        }
+    }
+
+    private void resize(int size) {
+        if (size == arrayList.length) {
+            T[] newArray = (T[]) new Object[size + (size / GROW_CONSTANT)];
```
okuzan, line 98: resize shouldn't accept any arguments
```suggestion
    private void resize() {
        if (size == arrayList.length) {
            T[] newArray = (T[]) new Object[(int) (arrayList.length * GROW_CONSTANT)];
```

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        int numMoved = size - index - 1;
+        if (index > size || index < 0 || numMoved < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
+                    + " out of ArrayListSize");
+        }
```
Rommelua, line 58: use checkIndex()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index
+                    + " out of ArrayListSize");
+        }
+    }
+
+    private void resize() {
+        if (size == values.length) {
+            T[] newArray = (T[]) new Object[(int) (values.length * GROW_CONSTANT)];
```
Rommelua, line 98: image

Rommelua (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_CONSTANT = 2.0;
```
Rommelua, line 7: image

Rommelua (approved):

PR #1350, opened 2024-03-26
[Java core/ArrayList/ArrayList] solution (v1)

ShastkivRuslan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_SIZE = 10;
+    public static final double ADD_SIE = 1.5;
```
kerrrusha, line 7: better name
```suggestion
    public static final double RESIZE_FACTOR = 1.5;
```

kerrrusha, line 7: they are used only in this class, so should be `private`
```suggestion
    private static final int DEFAULT_SIZE = 10;
    private static final double ADD_SIE = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+    }
+
+    public void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Wrong index: " + index
+                    + ", size: " + size);
+        }
+    }
+
+    private void arrayCapacityControl() {
```
kerrrusha, line 89: method name should sound like an action
```suggestion
    private void controlCapacity() {
```

kerrrusha (changes requested):
overall is very good

kerrrusha (changes requested):
.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,96 @@
...
+    public ArrayList() {
+        mainData = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        controlCapacity();
+        mainData[size] = value;
+        size++;
```
Rommelua, line 19:
```suggestion
        mainData[size++] = value;
```

Rommelua (approved):

PR #1352, opened 2024-03-30
MateAcademy/Task/jv-array-list - Bartosz Wójcik solution

banderan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] entryArray;
```
okuzan, line 8: replace Object, that's what you have class type parameter here for

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    private int size;
+
+    public ArrayList() {
+        entryArray = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
okuzan, line 13: 0 is default for int primitive

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    }
+
+    private void extendArrayLength() {
```
okuzan, line 16: private methods (class members in general) should go after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        int oldArrayLength = entryArray.length;
+        int newArrayLength = oldArrayLength + (oldArrayLength >> 1);
```
okuzan, line 18: better create a clearly defined GROW_FACTOR (1.5) constant, bitwise operations are less readable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        entryArray = Arrays.copyOf(entryArray, newArrayLength);
+    }
+
     @Override
     public void add(T value) {
-
+        if (entryArray.length <= size) {
+            extendArrayLength();
+        }
```
okuzan, line 26: if you only call method in `if`, consider moving `if` inside of the method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("index is too big");
+        }
```
okuzan, line 55: code duplication, extract to a private utility method, include index and size in the message for debugging purposes

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (entryArray[i] == null || entryArray[i].equals(element)) {
+                System.arraycopy(entryArray, i + 1, entryArray, i, size - i - 1);
+                entryArray[size - 1] = null;
+                size--;
+                return element;
+            }
+        }
+        throw new NoSuchElementException("No Such Element in array");
```
okuzan, line 92: include element in the message, we should strive to make them as informative as possible

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+
+    public ArrayList() {
+        entryArray = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private void extendArrayLength() {
+        int oldArrayLength = entryArray.length;
+        int newArrayLength = oldArrayLength + (oldArrayLength >> 1);
+        entryArray = Arrays.copyOf(entryArray, newArrayLength);
```
okuzan, line 19: don't use Arrays util class - [checklist ](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class) issue

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (entryArray[i] == null || entryArray[i].equals(element)) {
+                System.arraycopy(entryArray, i + 1, entryArray, i, size - i - 1);
+                entryArray[size - 1] = null;
+                size--;
+                return element;
```
okuzan, line 89: code duplication, you can reuse your method of removing by index

okuzan (changes requested):
Here are some suggestions to improve your solution 👇

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void extendArrayLength() {
+        if (entryArray.length <= size) {
+            int oldArrayLength = entryArray.length;
+            int growFactor = oldArrayLength + (oldArrayLength >> 1); // extend array length by x1.5
```
nowickimj, line 82: Missed previous comment here - grow/resize factor is constant (1.5) and therefore should be defined as a constant.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+            T[] newArray = (T[]) new Object[growFactor];
+            System.arraycopy(entryArray, 0, newArray, 0, size);
+            entryArray = newArray;
+        }
+    }
+
+    private void arrayException(int index) {
```
nowickimj, line 89: this method name doesn't say anything about its behavior nor what's it intended for - if you want to check index here, please name it properly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("index is out of bounds: index-" + index);
+        }
+    }
+
+    private void arrayCopyMethod(int i) {
```
nowickimj, line 95: this is not a proper name for method as well - you don't need a "method" in it since it's obvious.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        extendArrayLength();
 
+        entryArray[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
```
nowickimj, line 20: You've created a method for that, why not reuse it?

nowickimj (changes requested):
Please make sure that you applied ALL suggested fixes before re-requesting a review. Otherwise, it will be immediately rejected without checking.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     public void set(T value, int index) {
-
+        checkIndexValidity(index);
+        entryArray[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T element;
```
nowickimj, line 51: No need to split the declaration from assignment - make both in single line

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROWTH_FACTOR = 1.5;
+    private T[] entryArray = (T[]) new Object[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void add(T value) {
+        extendArrayLength();
 
+        entryArray[size] = value;
```
nowickimj, line 15: Unnecessary new line left above

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void extendArrayLength() {
+        if (entryArray.length <= size) {
+            int oldArrayLength = entryArray.length;
+            int growFactor = (int) (oldArrayLength * GROWTH_FACTOR); // extend array length by x1.5
```
nowickimj, line 83: 1. Variable name seems to be wrong - shouldn't this variable be called `newLength`?
2. no need to store `oldArrayLength` as variable, call `length` immediately here

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+        }
+    }
+
+    private void checkIndexValidity(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayListIndexOutOfBoundsException("index is out of bounds: index-" + index);
+        }
+    }
+
+    private void arrayCopy(int i) {
```
nowickimj, line 96: This name seems wrong:
1. According to method naming convention, it should be an imperative verb: e.g. "doSomething()" - just like the `checkIndexValidity` above
2. Name doesn't really says what the method is supposed to do.
3. Name mentions "copying" whereas you also modify `size` value, which could be considered a side effect.

src/test/java/core/basesyntax/ArrayListTest.java
nowickimj, line 1: You are not supposed to modify this class - please revert these changes.

nowickimj (changes requested):

PR #1354, opened 2024-04-02
added class ArrayList with implements List interface

mstefanchuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+                }
+            }
+            throw new NoSuchElementException("Element not found");
+        }
+        for (int i = 0; i < size; i++) {
+            if (element.equals(elements[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
Vasul-Pankiv, line 73: use if-else and throw exception only once after if-else

Vasul-Pankiv (changes requested):
Great Job!

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elements[i])) {
+                    return remove(i);
+                }
+            }
+        }
```
olekskov, line 72: I'd recommend not to split logic for null

olekskov (approved):
Good job

PR #1355, opened 2024-04-02
Task performance jv-array-list

ruslanPan:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+    public ArrayList() {
+        this.elementData = (T[]) new Object[ARRAY_DEFAULT];
+    }
+
     @Override
     public void add(T value) {
-
+        if (arraySize == elementData.length) {
+            increaseElementData();
+        }
```
Vasul-Pankiv, line 19: move this check into inceraseElementData()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
+        elementData[arraySize] = value;
+        arraySize++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > arraySize) {
+            throw new ArrayListIndexOutOfBoundsException("Not correct index: " + index);
+        }
```
Vasul-Pankiv, line 28: extract it to separate method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
 
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= arraySize) {
+            throw new ArrayListIndexOutOfBoundsException("Not correct index: " + index);
+        }
```
Vasul-Pankiv, line 50: same here, you use this check 3 times, extract it to separate method and reuse

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,97 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return arraySize == 0;
+    }
+
+    private void increaseElementData() {
+        int newElementCapacity = elementData.length + (elementData.length >> 1);
```
Vasul-Pankiv, line 94: it is cool that you know how to use bit-wise operation, but it makes code harder to read, so better to use more understandable "/" divide operator

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+    }
+
+    private void increaseElementData() {
+        if (arraySize == elementData.length) {
+            int newElementCapacity = elementData.length + (elementData.length / 2);
+            elementData = Arrays.copyOf(elementData,newElementCapacity);
+        }
+    }
+
+    private void correctIndex(int index) {
```
olekskov, line 88: Method names contain verbs that reflect what we are doing.
We are not correcting index in any way, we are checking that it's valid so should be checkIndex or validateIndex
```suggestion
    private void checkIndex(int index) {
```

olekskov (approved):
Good job

PR #1369, opened 2024-04-17
First solution

Wojtas1996PL:


src/test/java/core/basesyntax/ArrayListTest.java
nowickimj, line 1: You are not supposed to modify test cases - please revert these changes.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            grow();
+            elements[size] = list.get(i);
+            size++;
+        }
```
nowickimj, line 44: 1. Use `forEach` loop - it looks more clear.
2. You have `add()` method - why not use it?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("There is no such index");
+        }
```
nowickimj, line 51: This code repeats 3 times - why not extract it to some private method? How about `checkIndex(int index)`?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void grow() {
+        if (size >= elements.length) {
+            int newSize = (size * 3) / 2;
```
nowickimj, line 115: Make resize factor a constant (`1.5`)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
+        if (!isExecuted) {
+            throw new NoSuchElementException("There is no such element");
+        }
+        T removedElement = elements[index];
+        for (int i = index; i < size; i++) {
+            grow();
+            elements[i] = elements[i + 1];
+        }
+        size--;
+        return removedElement;
```
nowickimj, line 100: Try to think how you can reuse the existing code:
1. Check for element index
2. if found, call `remove(index)`
3. if not, throw an exception

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,121 @@
...
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("There is no such index");
+        }
+        T removedElement = elements[index];
+        for (int i = index; i < size; i++) {
+            grow();
+            elements[i] = elements[i + 1];
+        }
```
nowickimj, line 72: Same here - use `System.arrayCopy()`

nowickimj (changes requested):
Please revert changes to test class - you're not supposed to modify it.

Besides that, few comments left.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
nowickimj, line 37: A hint: if you want to iterate over ALL elements in array/collection (without using any specific starting/ending index), please use `forEach` loop.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        boolean isExecuted = false;
+        T removedElement = (T) new Object();
+        for (int i = 0; i < size; i++) {
+            if ((elements[i] != null && elements[i].equals(element))
+                    || elements[i] == null && element == null) {
```
nowickimj, line 67: `Objects.equals()` gives you null-safe comparison.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+                isExecuted = true;
+                removedElement = elements[i];
+                remove(i);
+                break;
```
nowickimj, line 71: `return removedElement` from the loop - no need to store `isExecuted` flag.
If you didn't return from loop, you can throw an exception right after the loop.

nowickimj (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Wojtas1996PL: I tried to use forEach but I am getting this information: "Foreach not applicable to type 'core.basesyntax.List<T>'"

src/main/java/core/basesyntax/ArrayList.java
Wojtas1996PL: But the checklist says "Don't use Objects, Arrays, or any other util class.". So can I use Objects.equals?

nowickimj: Good call, didn't see that this requirement was added.
Let's stick with your implementation.

src/main/java/core/basesyntax/ArrayList.java
nowickimj: My bad - it will work only for `Iterable`s.
Let's keep it that way.

nowickimj (changes requested):
Sorry for late response, but I didn't see your review request - once I submit a review, the PR is no longer visible for me so if further explanations are needed, don't hesitate to let me know on chat :)

I also did 2 mistakes during the review - my apologies for the trouble.

Please rework the `remove()` method to remove `break` usage according to my comment.

nowickimj (approved):

PR #1370, opened 2024-04-18
add hw jv-array-list solution

oleksandr-saulko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        this.elements = (T[]) new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
Elena-Bruyako, line 14: int type = 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
Elena-Bruyako, line 89: You can modify your logic a little. 
First, let's name this method - resize(), it's more appropriate.
Second, you don't actually need to pass some arguments, because the main purpose is to check if array is full like elements.length == size. 
Third, ArrayList increases to 1.5, not 2.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
 
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Elena-Bruyako, line 45: A lot of duplication, think how to manage it

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
oleksandr-saulko: Done

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
     public void set(T value, int index) {
-
+        indexPresent(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Elena-Bruyako, line 57: I think you have method for this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+        resize(size + 1);
```
Elena-Bruyako, line 28: Actually you don't need to provide any of arguments here. 
In method resize you can just check that `elements.length == size` and do the logic

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
oleksandr-saulko: Oh yeah) Done

src/main/java/core/basesyntax/ArrayList.java
oleksandr-saulko: Done, thx )

olekskov (approved):
Good job

PR #1371, opened 2024-04-18
Created ArrayList class which implements the List interface.

Alona-Martynova:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexCheck(int index) {
```
Elena-Bruyako, line 101: Minor comment, it's better to start the name of method from verb - checkIndex

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] arrayList;
+    private int size;
+    private int capacity;
+
     @Override
-    public void add(T value) {
+    protected Object clone() throws CloneNotSupportedException {
```
Elena-Bruyako, line 13: Be careful with clone method.
Do we need to resize here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+        return super.clone();
+    }
 
+    public void growCapacity() {
```
Elena-Bruyako, line 17:
```suggestion
    private void resize() {
```
Also, private method should be placed after all public methods
Think how to modify this logic according to comment above?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
```
Elena-Bruyako, line 7: Do we need to initialise the array with default capacity?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    private T[] arrayList;
```
Elena-Bruyako, line 8:
```suggestion
    private T[] elements;
```
or values

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+    public void add(T value) {
+        if (size == capacity) {
+            growCapacity();
+        }
+        arrayList[size++] = value;
+    }
 
+    @Override
+    public void add(T value, int index) {
+        if (index < 0 || index > size) {
```
Elena-Bruyako, line 38: This method could be simplified 

```
checkIndex(index);
resize();
System.arraycopy(elements, index, elements,index + 1, size - index);
elements[index] = value;
size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+            System.arraycopy(arrayList, index, arrayList, index + 1, size - index);
+            arrayList[index] = value;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
```
Elena-Bruyako, line 55: Do we need to resize here?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,106 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index);
+        return arrayList[index];
```
Elena-Bruyako, line 63:
```suggestion
        return (T) arrayList[index];
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Alona-Martynova: Please, explain why type casting is necessary? As I understand, the element has type T.

src/main/java/core/basesyntax/ArrayList.java
Alona-Martynova: The method 'checkIndex' has a different implementation.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+    }
+
+    private void resize() {
+        if (size == capacity) {
+            T[] initialArray = elements;
+            elements = (T[]) new Object[capacity + capacity / 2];
```
okuzan, line 99: better create explicit GROW_FACTOR constant

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 92: would be nice to include size here too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elements[i], element)) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("The element not found");
```
okuzan, line 77: include element here to make exception more debugging-friendly

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
+        elements[--size] = null;
+        return arrayElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(elements[i], element)) {
```
okuzan, line 73: [checklist issue](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+        resize();
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        resize();
```
okuzan, line 43:
```suggestion
```
not fixed
resizing happens in add() method anyway
it's functionally equivalent to calling add multiple times

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
...
+    public void add(T value) {
+        resize();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Index out of range: " + index);
```
okuzan, line 33: include size here too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,104 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elements;
+    private int size;
+    private int capacity;
```
okuzan, line 10: it's quite useless as a field, we can call .length on elements when we need it, not worth to keep it

okuzan (changes requested):
Good job!

src/main/java/core/basesyntax/ArrayList.java
Alona-Martynova: After removing the variable, it's not completely clear how to increment the size array.

Elena-Bruyako: In method resize you should calculate new capacity `int newCapacity = (int) (elements.length * CAPACITY_INDEX);` and create newArray with this newCapacity, then just copy old array into new array

src/main/java/core/basesyntax/ArrayList.java
Elena-Bruyako: I know, I gave you an example of code. You always can create private method to check if index fits

src/main/java/core/basesyntax/ArrayList.java
Elena-Bruyako: You are right, my mistake, didn't notice that you create T[] elements.

Elena-Bruyako (changes requested):
Discussed some comments

src/main/java/core/basesyntax/ArrayList.java
Alona-Martynova: > it's quite useless as a field, we can call .length on elements when we need it, not worth to keep it

I mean, that in different code parts the field (the array size) changes. For example, it increases: 
       `resize();
        elements[index] = value;
        size++;`
decreases:
`public T remove(int index) {
        checkIndex(index);
        T arrayElement = elements[index];
        elements[--size] = null;
        return arrayElement;
    }`.
.length is fixed.

> In method resize you should calculate new capacity `int newCapacity = (int) (elements.length * CAPACITY_INDEX);` and create newArray with this newCapacity, then just copy old array into new array

Also, in each method, this field is used differently. Need to create multiple methods (create and copy new arrays) and is that better than one field?

Elena-Bruyako: Let's discussed it in chat on the platform?

Elena-Bruyako (changes requested):
Let's discuss your question on chat on the platform

src/main/java/core/basesyntax/ArrayList.java
Alona-Martynova: > it's quite useless as a field, we can call .length on elements when we need it, not worth to keep it

Sorry, I thought that the useless field was the 'size' field.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,114 @@
...
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
-    public void add(T value) {
+    protected Object clone() throws CloneNotSupportedException {
```
Elena-Bruyako, line 16: Why do you need this method?

Elena-Bruyako (approved):
Approve, but clone method is useless

PR #1375, opened 2024-04-24
jv-array-list task solution

SvNadorozhnyi:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void grow(int minCapacity) {
+        if (minCapacity >= values.length) {
```
Elena-Bruyako, line 80: You actually don't need to provide minCapacity here, you can just check 
```suggestion
        if (values.length == size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,93 @@
...
+            int newCapacity = (int) (size + (size * 1.5));
```
Elena-Bruyako, line 81:
```suggestion
            int newCapacity = (int) (elements.length * CAPACITY_INDEX);
```
make 1.5 constant

Elena-Bruyako (changes requested):

PR #1388, opened 2024-05-18
Implementation of ArrayList

RomanChygryn:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Elena-Bruyako, line 13:
```suggestion
```
int type = 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
Elena-Bruyako, line 16: private methods should be placed after all public and protected methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Elena-Bruyako, line 49: Create separate method to avoid duplication of code

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+
+    public ArrayList() {
+        elements = new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
+        if (minCapacity > elements.length) {
+            int newCapacity = Math.max(DEFAULT_CAPACITY, elements.length + (elements.length >> 1));
+            elements = Arrays.copyOf(elements, newCapacity);
```
Elena-Bruyako, line 19: [Don't use Objects, Arrays, or any other util class.](https://github.com/mate-academy/jv-array-list/blob/master/checklist.md#dont-use-objects-arrays-or-any-other-util-class)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+            System.arraycopy(elements, index + 1, elements, index, numMoved);
+        }
+        elements[--size] = null;
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int index = 0; index < size; index++) {
```
Elena-Bruyako, line 76: It's better to create 
```
private int index(T element) {
        int index = -1;
            if (elements[i] == element
                    || elements[i] != null && elements[i].equals(element)) {
                index = i;
            }
        }
        return index;
    }
```

then

```suggestion
        int index = index(element);
        if (index == -1) {
            throw new NoSuchElementException("There is no such element in the list, " + element);
        }
        return remove(index);
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
     public void set(T value, int index) {
 
+        checkIndexBounds(index, size);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T oldValue = get(index);
```
Elena-Bruyako, line 53:
```suggestion
        T removedElement = (T) elements[index];
        System.arraycopy(elements, index + 1, elements, index, size - index - 1);
        size--;
        return removedElement;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int minCapacity) {
```
Elena-Bruyako, line 81:
```suggestion
    private void resize() {
        if (elements.length == size) {
            int newCapacity = (int) (elements.length * CAPACITY_INDEX);
            Object[] newArray = new Object[newCapacity];
            System.arraycopy(elements, 0, newArray, 0, size);
            elements = newArray;
        }
    }
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double CAPACITY_INDEX = 1.5;
+    private Object[] elements;
```
```suggestion
    private T[] elements;
```
Using T as the type parameter for the elements array leverages the class's type parameter, T, to define the type of objects that the collection will store. This change enhances type safety across the class, ensuring that all elements in the collection are of the same type as specified by T.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        }
+    }
+
+    private void checkIndexBounds(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private int index(T element) {
```
okuzan, line 97:
```suggestion
    private int indexOf(T element) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,107 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexBounds(index, size);
+        return (T) elements[index];
```
okuzan, line 42:
```suggestion
        return elements[index];
```

okuzan (changes requested):
You're almost there!

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

src/main/java/core/basesyntax/ArrayList.java
RomanChygryn: changed

okuzan (approved):
Well done!

PR #1391, opened 2024-05-23
added solution of task

apanyutin:


MentorMA (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
     public T remove(int index) {
-        return null;
+        checkIndexForGet(index);
+        int startPointForCopySecondPart = index + 1;
+        T[] secondArrayPart = getArrayCopy(array, startPointForCopySecondPart, array.length);
+        T value = array[index];
+
+        for (int i = index; i < secondArrayPart.length; i++) {
+            array[i] = secondArrayPart[i - index];
+        }
```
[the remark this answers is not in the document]
apanyutin: I think code will be less readable

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,142 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
+        checkEnoughArraySize(size + list.size());
 
+        for (int i = 0; i < list.size(); i++) {
+            array[i + size] = list.get(i);
+        }
+        size = size + list.size();
```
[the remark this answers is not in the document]
apanyutin: in this case method add has a redundant operations which complicate the algorithm.

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
Vasul-Pankiv: I don't think so, you made the method more complicated, instead of using just one copy method, you are getting a second part and saving it. For me, it's less readable than just an arrayCopyMethod

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+
+    private void checkEnoughArraySize(int inputSize) {
+        if (inputSize >= array.length) {
+            int lagerSize = inputSize + (inputSize >> 1);
+            T[] intermediateArray = (T[]) new Object[lagerSize];
+
+            System.arraycopy(array, 0, intermediateArray, 0, array.length);
+            array = intermediateArray;
+        }
+    }
```
Vasul-Pankiv, line 99: So if you name the method "check", should you do some action with an array?
Please change the name or the logic because it doesn't match

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            if (array[i] == value || (array[i] != null && array[i].equals(value))) {
+                index = i;
+                isValueExist = true;
+                break;
+            }
+        }
+        if (!isValueExist) {
+            throw new NoSuchElementException("don't find such element");
+        }
+        return index;
```
Vasul-Pankiv, line 135:
```suggestion
    private int getIndexByValue(T value) {
       for (int i = 0; i < size; i++) {
            if (array[i] == value || (array[i] != null && array[i].equals(value))) {
                return i;
            }
        }
        throw new NoSuchElementException("don't find such element");
```

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+    public ArrayList() {
+        array = (T[]) new Object[DEFAULT_CAPASITY];
+    }
+
     @Override
     public void add(T value) {
-
+        increaseArraySizeIfNecessary(size);
+        array[size] = value;
+        size++;
```
```suggestion
        array[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
     public void add(T value, int index) {
+        checkIndexForAdd(index);
+        increaseArraySizeIfNecessary(size);
 
+        T[] secondArrayPart = getArrayCopy(array, index, array.length);
+        array[index] = value;
+
+        int startIndexForCopyArray = index + NUMBER_ONE;
+        System.arraycopy(secondArrayPart, 0, array, startIndexForCopyArray,
+                secondArrayPart.length - startIndexForCopyArray);
```
okuzan, line 32: simplify, you only need one arraycopy per method, you can copy array into itself

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        size = size + list.size();
+        increaseArraySizeIfNecessary(size);
 
+        int startIndexForListElements = size - list.size();
+        for (int i = 0; i < list.size(); i++) {
+            array[startIndexForListElements + i] = list.get(i);
+        }
```
okuzan, line 44: call add in a loop
maybe multiple resizes will be needed - it will be handled in add

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        if (inputSize >= array.length) {
+            int lagerSize = inputSize + (inputSize >> 1);
+            T[] intermediateArray = (T[]) new Object[lagerSize];
+
+            System.arraycopy(array, 0, intermediateArray, 0, array.length);
+            array = intermediateArray;
+        }
+    }
+
+    private T[] getArrayCopy(T[] inputArray, int startIndex, int endIndex) {
```
okuzan, line 97: you don't need this method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+
+    private void checkIndexForGet(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("index not correct");
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("index not correct");
```
okuzan, line 113: include index and size in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,125 @@
...
+        }
+    }
+
+    private int getIndexByValue(T value) {
+        for (int i = 0; i < size; i++) {
+            if (array[i] == value || (array[i] != null && array[i].equals(value))) {
+                return i;
+            }
+        }
+        throw new NoSuchElementException("don't find such element");
```
okuzan, line 123: include element in the message

okuzan (changes requested):

sarakhmen (approved):
lgtm

PR #1393, opened 2024-05-23
solution

mgsorigo:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,138 @@
...
+        if (minCapacity > elements.length) {
+            int newCapacity = (int) (elements.length * GROWTH_FACTOR);
+            if (newCapacity < minCapacity) {
+                newCapacity = minCapacity;
+            }
+            T[] newElements = (T[]) new Object[newCapacity];
+            copyElements(elements, newElements, size);
+            elements = newElements;
+        }
+    }
```
MateuszRostek, line 107: Right now you copy elements and modify the array underneath on each call of this method, but you should only do it when it's really needed, so you can do it more like this and adjust your code a bit:

image

MateuszRostek (approved):
Really good job! There are some things that you should change according to my comments, but overall your solution is really good, so I will approve it!

PR #1397, opened 2024-06-04
There were overrided all the method of List interface. In addition to…

odeineka:
… it were created checkIndex() and CheckIndexForAdd() witch throw ArrayListIndexOutOfBoundsException. To make the code clear, there were created helper methods

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,116 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private T[] elements;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public ArrayList() {
+        elements = (T[]) new Object[INITIAL_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        elements[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        ensureCapacity();
+        for (int i = size; i > index; i--) {
+            elements[i] = elements[i - 1];
+        }
+        elements[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedElement = elements[index];
+        for (int i = index; i < size - 1; i++) {
+            elements[i] = elements[i + 1 ];
+        }
+        elements[size - 1] = null;
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null) {
+                if (elements[i] == null) {
+                    return remove(i);
+                }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        ensureCapacity();
+        for (int i = size; i > index; i--) {
+            elements[i] = elements[i - 1];
+        }
```
MentorMateAcademy, line 28: image

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,116 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T removedElement = elements[index];
+        for (int i = index; i < size - 1; i++) {
+            elements[i] = elements[i + 1 ];
+        }
```
MentorMateAcademy, line 58: image

MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int SIZE_MULTIPLIER = 2;
```
Elena-Bruyako, line 7:
```suggestion
    private static final int SIZE_MULTIPLIER = 1.5;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
+        if (size == elements.length) {
+            resize();
```
Elena-Bruyako, line 87:
```suggestion
            T[] newElements = (T[]) new Object[elements.length * SIZE_MULTIPLIER];
            System.arraycopy(elements, 0, newElements, 0, elements.length);
            elements = newElements;
```
Don't need to create separate method for resize

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,109 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity() {
```
Elena-Bruyako, line 85:
```suggestion
    private void resize() {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
+        for (int i = 0; i < size; i++) {
+            if (element == null) {
+                if (elements[i] == null) {
+                    return remove(i);
+                }
+            } else {
+                if (element.equals(elements[i])) {
+                    return remove(i);
+                }
+            }
```
okuzan, line 70: replace if-else with simple if, because they have the same inner body (return remove(I))

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -59,14 +59,11 @@ public T remove(int index) {
...
-            } else {
-                if (element.equals(elements[i])) {
-                    return remove(i);
-                }
+            if (element == null && elements[i] == null) {
+                return remove(i);
+            }
+            if (element != null && element.equals(elements[i])) {
+                return remove(i);
             }
```
sarakhmen, line 67: could you compose these two IFs into single one? The body is the same. You can move conditions to the separate methods if needed in order to make the code more clear

sarakhmen (changes requested):
let's slightly improve your solution :)

MAMentorFX (approved):

PR #1403, opened 2024-06-18
Created ArrayList class which implements the List interface and overr…

ovodolazov:
…ide their methods

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,100 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+        values[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
sarakhmen, line 34: let's use for each loop instead

sarakhmen (approved):
Excellent job!

PR #1413, opened 2024-07-10
Created custom `ArrayList` class which implements the custom `List` i…

Pancher07:
…nterface. The required methods were overridden.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,125 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_INDEX = 1.5;
+    private static final int START_INDEX_VALUE = 0;
+    private T[] dataArray;
+    private int size;
+
+    public ArrayList() {
+        dataArray = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int capacity) {
+        dataArray = (T[]) new Object[capacity];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == dataArray.length) {
+            dataArray = grow();
+        }
+        dataArray[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("index is invalid");
+        }
+        if (size == dataArray.length) {
+            dataArray = grow();
+        }
+        if (index == size) {
+            dataArray[size++] = value;
+        } else {
+            System.arraycopy(dataArray, index, dataArray, index + 1, size++);
+            dataArray[index] = value;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int oldSize = size;
+        int neededSize = size + list.size();
+        int indexOfValueFromList = START_INDEX_VALUE;
+        if (size == dataArray.length || dataArray.length < neededSize) {
+            dataArray = growToNeededSize(neededSize);
+        }
+        for (int i = oldSize; i < neededSize; i++) {
+            dataArray[i] = list.get(indexOfValueFromList++);
+            size++;
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return (T) dataArray[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        checkIndex(index);
+        dataArray[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        return remove(dataArray[index]);
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        T[] temp = (T[]) new Object[Math.max(DEFAULT_CAPACITY, dataArray.length - 1)];
+        for (int i = 0; i < dataArray.length; i++) {
+            if (element == null && dataArray[i] == null
+                    || element != null && element.equals(dataArray[i])) {
+                for (int index = 0; index < i; index++) {
+                    temp[index] = dataArray[index];
+                }
+                for (int j = i; j < dataArray.length - 1; j++) {
+                    temp[j] = dataArray[j + 1];
+                }
+                System.arraycopy(temp, 0, dataArray, 0, temp.length);
+                size--;
+                return element;
+            }
+        }
+        throw new NoSuchElementException("This element does not belong to this list");
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T[] grow() {
+        double newCapacity = dataArray.length * GROW_INDEX;
+        T[] newArray = (T[]) new Object[(int) newCapacity];
+        System.arraycopy(dataArray, 0, newArray, 0, dataArray.length);
+        return newArray;
+    }
+
+    private T[] growToNeededSize(int neededSize) {
+        while (neededSize > dataArray.length) {
+            dataArray = grow();
+        }
+        return dataArray;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("index is invalid");
+        }
     }
 }


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+    private T[] dataArray;
+    private int size;
+
+    public ArrayList() {
+        dataArray = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int capacity) {
+        dataArray = (T[]) new Object[capacity];
+    }
```
MentorMateAcademy, line 18:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
-
+        int neededSize = size + list.size();
+        int indexOfValueFromList = START_INDEX_VALUE;
+        if (dataArray.length < neededSize) {
+            ensureCapacity(neededSize);
+        }
+        for (int i = size; i < neededSize; i++) {
+            dataArray[i] = list.get(indexOfValueFromList++);
+            size++;
+        }
```
MentorMateAcademy, line 55: just reuse method add

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,120 @@
...
+                    || element != null && element.equals(dataArray[i])) {
+                if (i < dataArray.length - 1) {
+                    System.arraycopy(dataArray, i + 1, dataArray, i, size - i);
+                }
+                dataArray[size - 1] = null;
+                size--;
+                return element;
+            }
+        }
+        throw new NoSuchElementException("This element does not belong to this list");
```
MentorMateAcademy, line 95: try to reuse method remove by index

MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Pancher07: In 'addAll' method I provided ensuring of capacity to be performed only once via checking:
```
if (dataArray.length < neededSize) {
            ensureCapacity(neededSize);
        }
```
This avoids repeated checks and expansions of the array.

If I reuse the add method, the program will redundantly check whether 'enshureCapacity' is needed every time the add method is executed:
```
public void add(T value) {
            ensureCapacity(size);
        }
        dataArray[size++] = value;
    }
```
This is why I add elements directly to the dataArray instead of reusing method add.

Or am I wrong and I still need to reuse the add method.

okuzan: You are right, but the performance impact of `if` is totally negligible, i'd argue having a clear and concise code is more important in most cases

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        final T returnedElement = dataArray[index];
+        if (index < size - 1) {
+            System.arraycopy(dataArray, index + 1, dataArray, index, size - index);
+        }
```
okuzan, line 65: this if is redundant, system.arraycopy does it for us

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < dataArray.length; i++) {
+            if (element == null && dataArray[i] == null
+                    || element != null && element.equals(dataArray[i])) {
+                return remove(i);
+            }
+        }
+        throw new NoSuchElementException("This element does not belong to this list");
```
okuzan, line 79: please add relevant details to the exception message (element)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("index is invalid");
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("index is invalid");
```
okuzan, line 107: please add index and size to messages

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        }
+    }
+
```
okuzan, line 110:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_INDEX = 1.5;
+    private T[] dataArray;
```
```suggestion
    private T[] data;
```
not a mistake, but we usually don't include data type in variable's name - Java is a strongly typed language so we know it anyway

okuzan (changes requested):
Well done, please add more info to exception messages to facilitate debugging

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    private int size;
+
+    public ArrayList() {
+        data = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == data.length) {
```
Elena-Bruyako, line 17:
```suggestion
        resize();
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            ensureCapacity(size);
+        }
+        data[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        checkIndexForAdd(index);
+        if (size == data.length) {
```
Elena-Bruyako, line 26:
```suggestion
        resize();
        System.arraycopy(data, index, data, index + 1, size - index);
        data[index] = value;
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            System.arraycopy(data, index, data, index + 1, size - index);
+        }
+        data[index] = value;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        int neededSize = size + list.size();
```
Elena-Bruyako, line 38:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            resize();
            data[size] = list.get(i);
            size++;
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        System.arraycopy(data, index + 1, data, index, size - index - 1);
+        data[size - 1] = null;
+        size--;
+        return returnedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < data.length; i++) {
```
Elena-Bruyako, line 71:
```suggestion
        int index = indexOf(element);
        if (index == -1) {
            throw new NoSuchElementException("There is no such element in the list, " + element);
        }
        return remove(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureCapacity(int expandableSize) {
```
Elena-Bruyako, line 91:
```suggestion
    private void resize() {
        if (data.length == size) {
            int newCapacity = (int) (data.length * GROW_INDEX);
            Object[] newArray = new Object[newCapacity];
            System.arraycopy(data, 0, newArray, 0, size);
            data = newArray;
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+            throw new ArrayListIndexOutOfBoundsException("index " + index + " is invalid, "
+                    + "size of this list is " + size);
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("index " + index + " is invalid, "
+                    + "size of this list is " + size);
+        }
```
Elena-Bruyako, line 109: Create method `private int index(T element)`

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,124 @@
...
-        return null;
+        checkIndex(index);
+        final T returnedElement = data[index];
+        System.arraycopy(data, index + 1, data, index, size - index - 1);
+        data[size - 1] = null;
+        size--;
+        return returnedElement;
     }
 
+    /* @Override
```
Elena-Bruyako, line 61: remove commented code

Elena-Bruyako (approved):

PR #1432, opened 2024-08-02
Added task solution

Lessorg:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int DEFAULT_SIZE = 0;
+    private static final Object[] DEFAULT_EMPTY_DATA_ARRAY = {};
+    private static final String INVALID_INDEX_ERROR_MESSAGE = 
+            "Index should be bigger or equal then 0 and less then ArrayList size";
+
+    private Object[] dataArray;
```
```suggestion
    private T[] dataArray;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+
+    private Object[] copyOf(Object[] dataArray, int newLength) {
+        Object[] newArray = new Object[newLength];
+
+        for (int i = 0; i < dataArray.length; i++) {
+            newArray[i] = dataArray[i];
+        }
+        dataArray = newArray;
+        return dataArray;
+    }
```
okuzan, line 113: use system.arraycopy instead of copying manually

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+
+    private void arrayCopy(Object[] sourceArray, int srcPos, Object[] dataArray,
+                           int destPos, int length) {
+        Object[] cloneArray = sourceArray.clone();
+        while (length != 0) {
+            dataArray[destPos++] = cloneArray[srcPos++];
+            --length;
+        }
```
okuzan, line 121: same

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int DEFAULT_SIZE = 0;
+    private static final Object[] DEFAULT_EMPTY_DATA_ARRAY = {};
+    private static final String INVALID_INDEX_ERROR_MESSAGE = 
+            "Index should be bigger or equal then 0 and less then ArrayList size";
+
+    private Object[] dataArray;
+    private int size;
+
+    ArrayList() {
```
okuzan, line 15: add public

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        dataArray = DEFAULT_EMPTY_DATA_ARRAY;
```
okuzan, line 16: you should create an array of default size 10 (should be a constant)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+        return size == DEFAULT_SIZE;
+    }
+
+    private void growIfArrayFull() {
+        if (size == dataArray.length) {
+            dataArray = grow();
+        }
+    }
+
+    private Object[] grow() {
```
okuzan, line 97: always use class type parameter given here (T) instead of object, that's what it was introduced for

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    public ArrayList() {
+        dataArray = (T[]) new Object[DEFAULT_SIZE];
+    }
+
     @Override
     public void add(T value) {
-
+        growIfArrayFull();
+        dataArray[size] = value;
+        ++size;
```
```suggestion
        dataArray[size++] = value;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,112 @@
...
+    }
+
+    private void growIfArrayFull() {
+        if (size == dataArray.length) {
+            dataArray = grow();
+        }
+    }
+
+    private T[] grow() {
+        int newLength = (size << 1) - (size >> 1);
```
okuzan, line 97: it's cool but not readable, i would say minority of coders would understand what happens here :)) Better just use GROW_FACTOR, performance impact neglible

okuzan (approved):

PR #1450, opened 2024-08-24
Implemented "ArrayList" class in accordance with original class.

f3ops3:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index, int upperBound) {
```
okuzan, line 81: Imo it's easier to remove second argument and just use a different checker in add, but this implementation is also possible

okuzan (changes requested):

Elena-Bruyako (approved):

PR #1451, opened 2024-08-26
This ArrayList class is a custom implementation of a resizable array …

bilberryyy:
…list that mimics the behavior of java.util.ArrayList. It dynamically adjusts the size of the internal array when needed, supports adding elements by index, retrieving elements, and removing elements by both index and value. Custom exceptions are thrown for invalid operations, ensuring safe and predictable behavior.

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,48 +1,110 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
+    }
+
+    private void growIfNeeded() {
+        if (size == elementData.length) {
+            resizeArray();
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size;
+
+    public ArrayList() {
+        elementData = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Elena-Bruyako, line 12:
```suggestion
```

int type = 0 by default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    }
+
+    private void growIfNeeded() {
```
Elena-Bruyako, line 15: private methods should be placed after all public methods

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,110 @@
...
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
+    }
+
+    private void resizeArray() {
+        int newCapacity = (int) (elementData.length * 3) / 2 + 1;
```
Elena-Bruyako, line 34:
```suggestion
        int newCapacity = (int) (elements.length * CAPACITY_INDEX);
```

create constant `private static final double CAPACITY_INDEX = 1.5;`

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1453, opened 2024-08-27
first solution

Rzarcik97:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final Object[] EMPTY_ELEMENT_DATA = {};
```
MateuszRostek, line 6: Remove it.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elementData;
```
MateuszRostek, line 8: You should use generics here,
```suggestion
    private T[] elementData;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+    private int size = 0;
+
+    public ArrayList() {
+        this.elementData = EMPTY_ELEMENT_DATA;
```
MateuszRostek, line 12:
```suggestion
        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
```

MateuszRostek (changes requested):
You should remove every occurence of `Object[]` and use generic type. There is a lot of refactoring needed, so implement these changes and then re-request and I will go through your solution once more then.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+            System.arraycopy(elements, index + 1, elements, index, (size - 1) - index);
+        }
+        elements[size -= 1] = null;
+    }
+
+    private void checkIndex(int index) {
+        if ((index > size - 1 || index < 0) || index == size) {
+            throw new ArrayListIndexOutOfBoundsException("index is out of bound of array");
+        }
+    }
```
MateuszRostek, line 26:
```suggestion
    private void validIndexCheck(int index) {
        if (index < 0 || index >= size) {
            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,132 @@
...
+        int oldCapacity = elementData.length;
+        if (oldCapacity > 0) {
+            int newCapacity = oldCapacity + (oldCapacity >> 1);
+            Object[] copyOfElementData = elementData;
+            elementData = new Object[Math.max(newCapacity,minCapacity)];
+            System.arraycopy(copyOfElementData,0, elementData, 0, oldCapacity);
+            return elementData;
+        }
+        return elementData = new Object[Math.max(DEFAULT_CAPACITY, minCapacity)];
+    }
```
MateuszRostek, line 38: You should do it immediately on this array, without returning anything, for example like this:
```suggestion
    private void growIfArrayFull() {
        if (size >= arrayUnderneath.length) {
            int newArraySize = (int) Math.round(arrayUnderneath.length * GROWTH_MULTIPLIER);
            T[] newArray = (T[]) new Object[newArraySize];
            System.arraycopy(arrayUnderneath, 0, newArray, 0, arrayUnderneath.length);
            arrayUnderneath = newArray;
        }
    }
```

MateuszRostek (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            growIfArrayFull(size);
+        }
+    }
+
     @Override
     public void add(T value) {
-
+        sizeCheck();
+        elementData[size] = value;
+        this.size = size + 1;
```
MateuszRostek, line 47:
```suggestion
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
-
+        T[] inputListToArray = (T[]) new Object[list.size()];
+        for (int i = 0; i < list.size(); i++) {
+            inputListToArray[i] = list.get(i);
+        }
+        if (inputListToArray.length > elementData.length - size) {
+            growIfArrayFull(inputListToArray.length + size);
+        }
+        System.arraycopy(inputListToArray, 0, elementData, size, inputListToArray.length);
+        size = size + inputListToArray.length;
```
MateuszRostek, line 73:
```suggestion
        size += inputListToArray.length;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final double GROWTH_MULTIPLIER = 1.5;
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elementData;
+    private int size = 0;
```
MateuszRostek, line 9: It's 0 by default, so you don't need to specify it.
```suggestion
    private int size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+                    if (element.equals(elementData[index])) {
+                        break found;
+                    }
+                }
+            }
+            throw new NoSuchElementException(element + " not found");
+        }
+        removeElementAtIndex(elementData, index);
+        return element;
     }
```
MateuszRostek, line 118: Please consider refactoring this method to something like this:

```java
    @Override
    public T remove(T element) {
        return null;
            if (element == arrayUnderneath[i] || (element != null
                    && element.equals(arrayUnderneath[i]))) {
                T elementToRemove = arrayUnderneath[i];
                removeElement(i);
                return elementToRemove;
            }
        }
        throw new NoSuchElementException("No such element: " + element);
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
     @Override
     public int size() {
-        return 0;
+        return this.size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return this.size == 0;
```
MateuszRostek, line 127:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,129 @@
...
+            throw new NoSuchElementException(element + " not found");
+        }
+        removeElementAtIndex(elementData, index);
+        return element;
     }
 
     @Override
     public int size() {
-        return 0;
+        return this.size;
```
MateuszRostek, line 122:
```suggestion
        return size;
```

MateuszRostek (approved):
Good job! See some advice and implement it!

PR #1454, opened 2024-08-28
I don`t understand why this code not working. A few times I checked c…

ynazarets:
…ode from others and it don`t work too...


MentorMateAcademy (changes requested):
image
Fix your checks

PR #1462, opened 2024-09-12
feat: implemented custom ArrayList with dynamic resizing, element add…

PavloMelnyk10:
…ition, removal by index and value. Added exception handling for invalid indexes and elements which not exist

### DIFF
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -97,7 +97,8 @@ public class ArrayList<T> implements List<T> {
         }
     }
 
-    private boolean equals(T a, T b) {
-        return (a == b) || (a != null && a.equals(b));
+    private boolean equals(T firstElement, T secondElement) {
+        return (firstElement == secondElement)
+                || (firstElement != null && firstElement.equals(secondElement));
     }
 }
--- a/src/main/java/core/basesyntax/ArrayList.java
+++ b/src/main/java/core/basesyntax/ArrayList.java
@@ -1,7 +1,6 @@
 package core.basesyntax;
 
 import java.util.NoSuchElementException;
-import java.util.Objects;
 
 public class ArrayList<T> implements List<T> {
     private static final int DEFAULT_CAPACITY = 10;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] elements;
```
okuzan, line 7: don't use Object type, we have generics in Java :)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```
0 is default

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,102 @@
...
+    }
+
+    private void ensureCapacity() {
+        if (size == elements.length) {
+            resize();
+        }
+    }
+
+    private void resize() {
+        int newCapacity = elements.length + elements.length / 2;
```
okuzan, line 86: Better create clearly defined 1.5 GROW_FACTOR constant

okuzan (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,103 @@
...
+    }
+
+    private void validateIndex(int index, boolean isAddOperation) {
+        if (index < 0 || index > size || (!isAddOperation && index == size)) {
+            throw new ArrayListIndexOutOfBoundsException(
+                    "Invalid index: " + index + ", Size: " + size);
+        }
+    }
+
+    private boolean equals(T a, T b) {
```
Elena-Bruyako, line 100: rename a and b

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1463, opened 2024-09-15
Solution

vsupruniuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+    }
+
     @Override
     public void add(T value) {
+        if (this.size == this.currentCapacity) {
+            this.resizeStorage();
+        }
 
+        this.storage[this.size] = value;
+        this.size += 1;
```
liliia-ponomarenko, line 27:
```suggestion
       size ++;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+        if (this.size == this.currentCapacity) {
+            this.resizeStorage();
+        }
+
+        int idx = this.currentCapacity - 1;
+
+        while (idx > index) {
+            this.storage[idx] = this.storage[idx - 1];
+            idx -= 1;
+        }
```
liliia-ponomarenko, line 43: Use [System.arraycopy()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int-) to move your array elements.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+
+        this.storage[index] = value;
+        this.size += 1;
     }
 
     @Override
     public void addAll(List<T> list) {
+        while (this.size + list.size() > this.currentCapacity) {
+            this.resizeStorage();
+        }
```
liliia-ponomarenko, line 53: redundant loop you call this method inside the method add

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+    @SuppressWarnings("unchecked")
+    public ArrayList() {
+        this.storage = (T[]) new Object[DEFAULT_CAPACITY];
+        this.currentCapacity = DEFAULT_CAPACITY;
+    }
+
     @Override
     public void add(T value) {
+        if (this.size == this.currentCapacity) {
+            this.resizeStorage();
```
liliia-ponomarenko, line 23: please do not use the keyword `this` when calling methods of this class, in fact even with fields it is optional. Its purpose is to identify the variable of the current class from any external class variable with the same name (there is nothing similar in your class). It is also customary to use it in the constructor when initializing class fields

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+                removedValue = this.storage[i];
+            }
+
+            if (i > index) {
+                this.storage[i - 1] = this.storage[i];
+            }
+
+            if (i == this.size - 1) {
+                this.storage[i] = null;
+            }
```
liliia-ponomarenko, line 91: use System.arrayCopy

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+        }
+
+        this.size -= 1;
```
liliia-ponomarenko, line 94:
```suggestion
        size --;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+
+        return removedValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < this.size; i += 1) {
```
liliia-ponomarenko, line 101:
```suggestion
        for (int i = 0; i < this.size; i++) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+            if (this.storage[i] == element
+                    || (element != null
+                    && element.equals(this.storage[i]))
```
liliia-ponomarenko, line 104:
```suggestion
                    || element != null
                    && element.equals(this.storage[i])
```
These brackets are redundant because && has higher priority then ||

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,144 @@
...
+
+    @SuppressWarnings("unchecked")
+    private void resizeStorage() {
+        int newCapacity = (int) (this.currentCapacity * CAPACITY_MULTIPLIER);
+
+        T[] newStorage = (T[]) new Object[newCapacity];
+
+        for (int i = 0; i < currentCapacity; i += 1) {
+            newStorage[i] = this.storage[i];
+        }
```
liliia-ponomarenko, line 131: use System.arrayCopy

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

liliia-ponomarenko (approved):
Awesome 😎

PR #1467, opened 2024-09-19
Implemented methods in class 'ArrayList'

Petliuk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        }
+        size--;
+        return oldValue;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (element == null && elements[i] == null
```
```suggestion
            if (element == elements[i]
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+        checkIndex(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T oldValue = elements[index];
+        for (int i = index; i < size - 1; i++) {
```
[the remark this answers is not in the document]
okuzan: Use System.arraycopy() to move your array elements. It is more efficient than using a loop for copying elements when removing an element at a specific index.

src/main/java/core/basesyntax/ArrayList.java
Petliuk: Can't compare objects via '==' in 'remove' method, because '==' checks if the references point to the same object, not the same data. For this, you should use 'equals', which compares the contents of objects.
In my code, the data is stored in a 'heap' including the objects of the 'elements' array.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    @SuppressWarnings("unchecked")
```
ahoienko, line 82:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,95 @@
...
+    private void growArray() {
+        if (size == elements.length) {
+            int newCapacity = (int) (elements.length * INCREASING_ARRAY);
+            elements = Arrays.copyOf(elements, newCapacity);
```
ahoienko, line 86: use  System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,111 @@
...
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        ensureCapacity();
+        elements[size] = value;
+        size++;
```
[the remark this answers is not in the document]
ahoienko: Not resolved

ahoienko (changes requested):
Nice, but some previous comments are not resolved

Elena-Bruyako (approved):

PR #1495, opened 2024-10-20
created ArrayList.java which implements List interface.

opdhk:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+
+    private void increaseArraySizeIfFilled() {
+        if (size() == elements.length) {
+            Object[] increasedArray = new Object[(int) (size * ARRAY_SCALE_FACTOR + 1)];
+            System.arraycopy(elements, 0, increasedArray, 0, size);
+            elements = increasedArray;
+        }
+    }
+
+    private void ifIndexNotOutOfBonds(int index, int size) {
```
JJJazl, line 93: This method can throw `ArrayListIndexOutOfBoundsException`, so `check` in name will make it clearer
```suggestion
    private void checkIfIndexOutOfBounds(int index, int size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
-
+        increaseArraySizeIfFilled();
+        elements[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        ifIndexNotOutOfBonds(index, size());
```
JJJazl, line 28:
```suggestion
        ifIndexNotOutOfBonds(index, size);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
     public T get(int index) {
-        return null;
+        ifIndexNotOutOfBonds(index, size() - 1);
+        return (T) elements[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        ifIndexNotOutOfBonds(index, size() - 1);
```
JJJazl, line 50:
```suggestion
        ifIndexNotOutOfBonds(index, size - 1);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        ifIndexNotOutOfBonds(index, size() - 1);
```
JJJazl, line 56:
```suggestion
        ifIndexNotOutOfBonds(index, size - 1);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        T removedElement = (T) elements[index];
+        if (index < size() - 1) {
+            System.arraycopy(elements, index + 1, elements, index, size() - index);
+        }
+        size--;
```
JJJazl, line 61: Does it remove last element?

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,98 @@
...
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if (elements[i] != null && elements[i].equals(element) || elements[i] == element) {
+                return remove(i);
```
JJJazl, line 69: I think it would be better to be more specific in this case
```suggestion
        for (int index = 0; index < size; index++) {
            if (elements[index] != null && elements[index].equals(element) || elements[index] == element) {
                return remove(index);
```

JJJazl (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,101 @@
...
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double ARRAY_SCALE_FACTOR = 1.5;
+    private int size;
+    private Object[] elements;
+
+    public ArrayList(int capacity) {
+        this.elements = new Object[capacity];
+    }
```
liliia-ponomarenko, line 13: why do you need this constructor?  is it a good idea to provide such a constructor to the user?

liliia-ponomarenko (approved):
Good job! One question ;)

PR #1516, opened 2024-11-24
fixed

LizaBalshv:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     }
 
     @Override
     public void set(T value, int index) {
-
+        if (index >= 0 && index < size) {
+            elementData[index] = value;
+        } else {
+            throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", is invalid ");
+        }
```
liliia-ponomarenko, line 62: avoid using if-else construction. 
```suggestion
        checkIndex(index); 
        elementData[index] = value;
```
And create private method:
```
           throw new ArrayListIndexOutOfBoundsException("Index: " + index + ", is invalid ");
        }
    }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
```
liliia-ponomarenko, line 67: the same. just call checkIndex method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,105 @@
...
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < elementData.length; i++) {
+            if (elementData[i] == null && element == null) {
+                return remove(i);
+            }
+            if (elementData[i] != null && elementData[i].equals(element)) {
+                return remove(i);
+            }
```
liliia-ponomarenko, line 84: can you combine these conditionals?

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Elena-Bruyako: fix comment

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        }
+        throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < elementData.length; i++) {
+            if ((elementData[i] == null && element == null)
+                    || (elementData[i] != null && elementData[i].equals(element))) {
```
Elena-Bruyako, line 71:
```suggestion
            if (elementData[i] == element
                    || elementData[i] != null && elementData[i].equals(element)) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
Sanderfuck: Use here method that you have already

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
+            T deleteValue = elementData[index];
+            System.arraycopy(elementData, index + 1, elementData, index, size - index - 1);
+            elementData[--size] = null;
+            return deleteValue;
+        }
+        throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
```
Sanderfuck, line 64:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+        elementData[size] = value;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Invalid index: " + index);
+        }
```
Sanderfuck, line 28: you can extract this check to private method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,100 @@
...
+    public ArrayList() {
+        this.elementData = (T[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size == elementData.length) {
+            grow();
+        }
```
Sanderfuck, line 19:
```suggestion
            grow();
```

Sanderfuck: You can add this inside method grow

Sanderfuck (changes requested):

PR #1530, opened 2024-12-11
Array

VovaPryt:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private T[] elements;
+    private int size;
+
+    public ArrayList() {
+        elements = (T[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Elena-Bruyako, line 12: don't need to initialize int, it = 0 by default
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        checkIndex(index);
+        elements[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        T removedElement = elements[index];
+        shiftLeft(index);
```
Elena-Bruyako, line 58: just use System.arraycopy()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,131 @@
...
+        size--;
+        return removedElement;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        for (int i = 0; i < size; i++) {
+            if ((elements[i] == null && element == null) || (elements[i]
+                    != null && elements[i].equals(element))) {
```
Elena-Bruyako, line 67:
```suggestion
            if (elements[i] == element == null || elements[i]
                    != null && elements[i].equals(element)) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
VovaPryt: I think you wrote something wrong here

VovaPryt: elements[i] == element == null

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+            while (newCapacity < minCapacity) {
+                newCapacity = (int) (newCapacity * DEFAULT_MAGNIFICATION);
+            }
+            T[] newElements = (T[]) new Object[newCapacity];
+            System.arraycopy(elements, 0, newElements, 0, size);
+            elements = newElements;
+        }
+    }
+
+    private void shiftRight(int index) {
```
Elena-Bruyako, line 106: this method is redundant, remove it and use `System.arraycopy(elements, index, elements, index + 1, size - index);` directly in method

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,127 @@
...
+        System.arraycopy(elements, index, elements, index + 1, size - index);
+    }
+
+    private void shiftLeft(int index) {
```
Elena-Bruyako, line 110: same here

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1532, opened 2024-12-13
initial commit

Tetiana-Hrytsenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    public static final int DEFAULT_CAPACITY = 10;
+    public static final double ARRAY_GROWTH_VALUE = 1.5;
+    private Object[] elementData;
```
okuzan, line 8: use your class type parameter T (you'll have to cast in constructor)

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
     public void add(T value, int index) {
+        checkIndexForAdd(index);
+        growIfArrayIsFull();
+        System.arraycopy(elementData, index, elementData, index + 1,
+                size - index);
+        elementData[index] = value;
+        size++;
+    }
+
+    private void checkIndexForAdd(int index) {
```
okuzan, line 32: put private methods after public ones

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    return remove(i);
+                }
+            }
+        }
+        throw new NoSuchElementException("Element not found");
```
okuzan, line 95: include element in the message

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        if (element == null) {
+            for (int i = 0; i < size; i++) {
+                if (elementData[i] == null) {
+                    return remove(i);
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (element.equals(elementData[i])) {
+                    return remove(i);
```
okuzan, line 91: lines 85, 91 - actual action is the same in both clauses - it means you just have to merge two conditions into one
use the same code as in Java's objects.equals (which you can't use directly)

okuzan (changes requested):
Good job!

Elena-Bruyako (approved):

PR #1533, opened 2024-12-17
Custom ArrayList implementation

euulu:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+            data = grow();
+        }
+        data[size] = value;
+        size += 1;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size) {
```
Elena-Bruyako, line 26:
```suggestion
        if (index < 0 || index > size) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+    private void isIndexInBounds(int index) {
+        if (index >= size) {
+            throw new ArrayListIndexOutOfBoundsException("Index "
+                    + index
+                    + " is out of bounds for the size "
+                    + (size));
+        }
+    }
+
+    private void isIndexPositiveNumber(int index) {
```
Elena-Bruyako, line 115: remove

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     public void addAll(List<T> list) {
-
+        int newSize = this.size + list.size();
+        if (data.length < newSize) {
+            data = grow(newSize);
+        }
+        for (int i = size, j = 0; i < newSize; i++, j++) {
+            data[i] = list.get(j);
+        }
+        size += list.size();
```
Elena-Bruyako, line 50:
```suggestion
        for (int i = 0; i < list.size(); i++) {
            add(list.get(i));
        }
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get(int index) {
-        return null;
+        isIndexInBounds(index);
+        isIndexPositiveNumber(index);
```
Elena-Bruyako, line 57: create method checkIndex 
where `if (index < 0 || index >= size)`

```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        return (T) data[index];
     }
 
     @Override
     public void set(T value, int index) {
-
+        isIndexInBounds(index);
+        isIndexPositiveNumber(index);
```
Elena-Bruyako, line 64:
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        data[index] = value;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T remove(int index) {
-        return null;
+        isIndexInBounds(index);
+        isIndexPositiveNumber(index);
```
Elena-Bruyako, line 72:
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
     @Override
     public int size() {
-        return 0;
+        return this.size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= 0;
```
Elena-Bruyako, line 103:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        if (!isElementFound) {
+            throw new NoSuchElementException("No element found by value " + element);
+        }
+        return this.remove(indexToRemove);
     }
 
     @Override
     public int size() {
-        return 0;
+        return this.size;
```
Elena-Bruyako, line 98:
```suggestion
        return size;
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+        T elementToRemove = (T) data[index];
+        System.arraycopy(data, index + 1, data, index, size - index - 1);
+        size--;
+        return elementToRemove;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int indexToRemove = 0;
```
Elena-Bruyako, line 81: let's find index of element - create method `private int index(T element)`
```
int index = index(element);
if (index == -1) {
            throw new NoSuchElementException(........
}
return remove(index);
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
euulu: Can you please suggest the better way to apply your suggestions(when i'm studying and when developing a commercial priject):
- commit suggestion through github interface
- change code localy push and re-request review
Thank you.

src/main/java/core/basesyntax/ArrayList.java
euulu: Can you please suggest how to better implement checkIndex(). Can i have two separate if blocksin the method: first with i < 0, second with i >= size, because i need to throw different exceptions for these cases?

euulu: @Elena-Bruyako Also why is it better to perform two different checks in one method?

euulu: @Elena-Bruyako Can you please suggest how to implement this check. As for now i applied all the changes you requested(did not push because some tests are failing).
I've added checkIndex method, but if i use `(index < 0 || index >= size)` i get two tests fail with:
`core.basesyntax.ArrayListIndexOutOfBoundsException: Index 0 is out of bounds for the size 0
core.basesyntax.ArrayListIndexOutOfBoundsException: Index 5 is out of bounds for the size 5`
but if i use `(index < 0 || index > size)` condition i get three tests fail with:
`java.lang.AssertionError: Expected exception: core.basesyntax.ArrayListIndexOutOfBoundsException
java.lang.Exception: Unexpected exception, expected<core.basesyntax.ArrayListIndexOutOfBoundsException> but was<java.lang.ArrayIndexOutOfBoundsException>`
Can't get what am i doing wrong.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,137 @@
...
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= data.length) {
+            data = grow();
+        }
+        data[size] = value;
+        size += 1;
```
liliia-ponomarenko, line 21:
```suggestion
        size++;
```

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: Always prefer a local resolving of suggestions (merge conflicts)

liliia-ponomarenko, line 26: not fixed

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: not fixed

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: not fixed

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: not fixed

src/main/java/core/basesyntax/ArrayList.java
liliia-ponomarenko: not fixed

liliia-ponomarenko (changes requested):
Let’s fix all previous comments! If you have some issues let's discuss them in the chat on the platform

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final double GROW_CAPACITY_COEFFICIENT = 1.5;
+    private Object[] data;
```
```suggestion
    private T[] data;
```
the whole point of class type parameters is to avoid Object :)

src/main/java/core/basesyntax/ArrayList.java
okuzan: not fixed

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,117 @@
...
+        this.data = new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void add(T value) {
-
+        if (size >= data.length) {
+            data = grow();
+        }
+        data[size] = value;
```
```suggestion
        data[size++] = value;
```
you can do it like this

okuzan (changes requested):

liliia-ponomarenko (approved):
Great 👍

PR #1539, opened 2024-12-26
Task has been completed

MishaHMK:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayListIndexOutOfBoundsException("Non existent "
+                    + "position: " + index);
+        }
+        if (size >= array.length) {
+            grow();
+        }
+        Object[] result = new Object[++size];
```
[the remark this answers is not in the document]
MishaHMK: I'm forbidden from using utility methods, so i will make my own implementation in next commit

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,133 @@
...
+        if (size >= array.length) {
+            grow();
+        }
+        array[size++] = value;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
```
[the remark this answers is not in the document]
MishaHMK: My unit tests are cheking cases, where 'index == size' and they are succesful, so I should maintain only 'bigger than size cases' it in order to pass unit tests

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Non existent "
+                    + "position: " + index);
```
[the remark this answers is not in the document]
MishaHMK: It is in separate class

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
 package core.basesyntax;
 
+import java.util.NoSuchElementException;
+
+@SuppressWarnings("unchecked")
 public class ArrayList<T> implements List<T> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] array;
```
Sanderfuck, line 8:
```suggestion
    private T[] array;
```

Sanderfuck: you have generic

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+        array = new Object[DEFAULT_CAPACITY];
+    }
+
+    public ArrayList(int capacity) {
+        if (capacity > 0) {
+            array = new Object[capacity];
+        } else {
+            throw new IllegalArgumentException("Capacity must be greater than zero.");
+        }
+    }
```
Sanderfuck, line 21:
```suggestion
    public ArrayList() {
       // init here default array
    }
```

Sanderfuck: make it more simple

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
+
     @Override
     public void add(T value) {
-
+        if (size >= array.length) {
+            grow();
+        }
```
Sanderfuck, line 27: put `if` inside method grow()

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (isNotValidIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Not found "
+                    + "element at position: ");
+        } else {
+            return (T) array[index];
```
Sanderfuck, line 60: Better cast in constructor

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        if (isNotValidIndex(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Not found "
+                    + "element at position: " + index);
+        }
+        Object toDelete = array[index];
+        removeAndTrim(index);
+        return (T) toDelete;
```
Sanderfuck, line 81: Cast one time in constructor

src/main/java/core/basesyntax/ArrayList.java
Sanderfuck, line 35: create method `checkIndex`

src/main/java/core/basesyntax/ArrayList.java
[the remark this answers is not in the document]
Sanderfuck: you have arrayCopy method for this

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,139 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void grow() {
+        Object[] newArray = new Object[array.length + array.length / 2];
```
Sanderfuck, line 114: Strange number should be constant

Sanderfuck (changes requested):

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
 
     @Override
     public T get(int index) {
-        return null;
+        if (isNotValidToFind(index)) {
+            throw new ArrayListIndexOutOfBoundsException("Not found "
+                    + "element at position: ");
+        } else {
+            return array[index];
+        }
```
okuzan, line 48: - you don't need else after returning from function (this means throwing too)
- put throwing inside of private method, since it's a repeated part too

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+    }
+
+    boolean isNotValidToFind(int index) {
+        return index < 0 || index >= size;
+    }
+
+    boolean isNotValidToAdd(int index) {
+        return index < 0 || index > size;
+    }
```
okuzan, line 114: make your utility methods private

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,128 @@
...
+        }
+        T toDelete = array[index];
+        removeAndTrim(index);
+        return toDelete;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int index = -1;
```
okuzan, line 73: let's streamline this, we don't need additional local variables
return from the for loop, throw exception outside of it, don't use break, it's less readable

okuzan (changes requested):
Good job so far!

sarakhmen (approved):

PR #1577, opened 2025-02-25
arraylist

weronikarzekiecka:


src/main/java/core/basesyntax/ArrayListIndexOutOfBoundsException.java
```diff
@@ -1,7 +1,7 @@
 package core.basesyntax;
 
-public class ArrayListIndexOutOfBoundsException extends RuntimeException {
+public class ArrayListIndexOutOfBoundsException extends IndexOutOfBoundsException {
     public ArrayListIndexOutOfBoundsException(String message) {
         super(message);
     }
-}
+}
\ No newline at end of file
```
slade13, line 7: There is no newline at the end of the file. It's a good practice to end files with a newline character to avoid issues with some tools and to follow POSIX standards.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+import java.util.NoSuchElementException;
+
+public class Main {
+    public static void main(String[] args) {
+        List<String> list = new ArrayList<>();
```
slade13, line 7: The ArrayList class is not imported. You need to import your custom ArrayList class to use it.

src/main/java/core/basesyntax/List.java
```diff
@@ -1,21 +1,9 @@
 package core.basesyntax;
 
 public interface List<T> {
-    void add(T value);
-
-    void add(T value, int index);
-
-    void addAll(List<T> list);
-
+    void add(T element);
```
slade13, line 4: The method signature for `add(T element)` is correct, but the previous method `add(T value, int index)` has been removed. This method is essential for inserting elements at specific positions in the list, which is a common feature of the List interface.

src/main/java/core/basesyntax/List.java
```diff
@@ -1,21 +1,9 @@
...
     T get(int index);
-
-    void set(T value, int index);
-
-    T remove(int index);
-
-    T remove(T element);
-
+    void remove(T element);
```
slade13, line 6: The method `remove(T element)` is present, but the method `remove(int index)` has been removed. Both methods are important for different use cases: removing by element and removing by index.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,74 @@
...
+            throw new ArrayListIndexOutOfBoundsException("Index out of bounds: " + index);
+        }
     }
 
-    @Override
-    public boolean isEmpty() {
-        return false;
+    private int indexOf(T element) {
+        for (int i = 0; i < size; i++) {
+            if (elements[i].equals(element)) {
```
slade13, line 63: The `indexOf` method should handle `null` elements correctly. Currently, it may throw a `NullPointerException` if `element` is `null`. Consider adding a null check.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,74 @@
...
 
+    public ArrayList() {
+        this.elements = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
     }
 
     @Override
-    public void add(T value, int index) {
-
+    public void add(T element) {
```
slade13, line 16: The `add` method should also handle the case where an element is added at a specific index, as per the `List` interface contract. Consider implementing an overloaded `add` method that takes an index.

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,74 @@
...
-
+    public T get(int index) {
+        checkIndex(index);
+        return (T) elements[index];
     }
 
     @Override
-    public T get(int index) {
-        return null;
+    public void remove(T element) {
```
slade13, line 30: The `remove` method should return the element that was removed, as per the `List` interface contract. Currently, it is a void method.

src/main/java/core/basesyntax/List.java
slade13, line 1: You should not make any changes in the List interface, please revoke this changes.

slade13 (changes requested):
Please do not change the `List` interface - also please implement all methods from it. In overall it's a good code but requires some changes.

PR #1594, opened 2025-03-19
implemented all required methods

DZotsenko:


src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+        return elementToRemove;
     }
 
     @Override
     public T remove(T element) {
-        return null;
+        int indexToRemove = NOT_FOUND;
+        for (int i = 0; i < size; i++) {
+            if (element == null && elementData[i] == null
+                    || element != null && element.equals(elementData[i])) {
```
Elena-Bruyako, line 76:
```suggestion
            if (element == elementData[i] 
                    || element != null && element.equals(elementData[i])) {
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+                indexToRemove = i;
+                break;
```
Elena-Bruyako, line 78:
```suggestion
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
+
+    private void grow(int minCapacity) {
+        int newCapacity = elementData.length + (elementData.length >> 1);
+        if (newCapacity < minCapacity) {
+            newCapacity = minCapacity;
+        }
+        if (newCapacity > MAX_ARRAY_SIZE) {
+            throw new OutOfMemoryError("Array size exceeds maximum allowed: " + MAX_ARRAY_SIZE);
+        }
+        Object[] biggerElementData = new Object[newCapacity];
```
Elena-Bruyako, line 105:
```suggestion
        int newCapacity = (int) (elementData.length * CAPACITY_INDEX);
        Object[] newArray = new Object[newCapacity];
```

src/main/java/core/basesyntax/ArrayList.java
```diff
@@ -1,48 +1,115 @@
...
-
+        checkIfIndexValid(index);
+        elementData[index] = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIfIndexValid(index);
+        T elementToRemove = get(index);
```
Elena-Bruyako, line 65:
```suggestion
        T elementToRemove = (T) elementData[index];
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ArrayList.java
DZotsenko: if (newCapacity < minCapacity) {
    newCapacity = minCapacity;
}
I still had to leave this check because addAll(List<T> list) method doesn't work without it. Because the number of elements added by one list can be greater than the 1.5x growth of the array.
I would be grateful for an explanation if I'm wrong.

DZotsenko:
if (newCapacity < minCapacity) {
newCapacity = minCapacity;
}
I still had to leave this check in grow(int minCapacity) method because addAll(List list) method doesn't work properly without it. Because the number of elements added by one list can be greater than the 1.5x growth of the array.
I would be grateful for an explanation if I'm wrong.

Elena-Bruyako (approved):

PR #1629, opened 2025-05-28
created my own ArrayList

b1gsal:


b1gsal:
Can you explain what you mean when say ArrayListIndexOutOfBoundsException is missing from your codebase.

PR #1634, opened 2025-06-10
array list task

nikolasdymchenko7:


nikolasdymchenko7:
It looks that you are wrong. Can you show me what is not good? My ArrayList class explicitly implement the List interface here: public class ArrayList<T> implements List<T>  (line 6). 
You can find require custom exception here: ArrayListIndexOutOfBoundsException (line: 184 and line 111). 

@mateacademy-ai-mentor

PR #1662, opened 2025-08-31
Dev/solution

nowakjd:
Implementing required methods in ArrayList class


nowakjd:
@mateacademy-ai-mentor Task require implementation of  interface core.basesyntax.List and there is no method named clean in this interface.

PR #1686, opened 2025-09-19
add implementation of ArrayList

slade13:


slade13:
@mateacademy-ai-mentor
My code is okay:

- Task doesn't require clear() method
- it's not magic number, I'm using binary operator and I'm adding half of current size to it. There is no magic number.
- I'm using maven checkstyle plugin and there are no errors about any empty lines.

slade13:
@mateacademy-ai-mentor

Tests are passed, all of the requirements are met, all interface methods are implemented and build passes.
`self-add edge case` - that's right, but it is out of scope for this task.

